PR #2, opened 2026-08-17
Rollup engine stage 1: dependency graph resolution

jonstump:
First implementation in the repo. Stage 1 of [SPEC-0001](docs/openspec/specs/rollup-engine/spec.md) only — resolve a target into a DAG, propagate and aggregate quantities. Stages 2 (producer rollup) and 3 (power) are untouched.

Stage 1 is what design.md's build order puts first, and it is the only stage needing no Tier 2 constants — so it is buildable now, before the extraction spike settles whether the economy constants exist in the game files at all.

## What's here

| File | |
|---|---|
| [`internal/domain/tier1.go`](internal/domain/tier1.go) | Tier 1 artifact schema, loader, structural validation |
| [`internal/domain/resolve.go`](internal/domain/resolve.go) | Stage 1: graph walk, method resolution, quantity propagation, cycle detection, provenance |
| [`internal/domain/errors.go`](internal/domain/errors.go) | Sentinel errors |
| [`internal/domain/testdata/stasis-device.tier1.json`](internal/domain/testdata/stasis-device.tier1.json) | The 34-node Stasis Device fixture |

20 tests, one per stage-1 spec scenario plus validation cases. 89.6% statement coverage. `go vet` and `gofmt` clean. No module dependencies.

It resolves the real tree — 34 nodes, every total matching ADR-0001's Confirmation section (500 each Sulphurine/Nitrogen/Radon, 300 Condensed Carbon aggregated across six recipes, 300 Frost Crystal, 150 Ionised Cobalt):

```
 Stasis Device             1  craft
   Cryogenic Chamber         1  craft
     Cryo-Pump                 1  craft
      ~Hot Ice                   1  craft
        ~Enriched Carbon           2  craft
          ~Condensed Carbon        300  raw
          ~Sulphurine              500  raw
```

`~` is the propagated unverified flag.

## Decisions worth a second look

**Quantities propagate as `math/big.Rat`, not float.** Stage 1's arithmetic is entirely integral, so this buys nothing yet. It is here because stages 2 and 3 introduce class multipliers of ×0.5/1.5 and fractional fill durations, and retrofitting exactness after the fact is precisely how the 499.9999999 bug design.md warns about gets in.

**Provenance taints downward**, the same direction quantities propagate: a total derived *through* an unverified node is itself unverified. Glass and Hot Ice carry the fixture's flags, so Frost Crystal, Sulphurine, Nitrogen and Condensed Carbon all inherit it. Frost Crystal shows unverified even under Heat Capacitor, because it is one aggregated node and Glass reaches it. Over-flagging is the honest direction per the spec.

**The ADR-0003 `no-syscall/js` boundary is enforced by construction**, not convention: `syscall/js` exists only on `js/wasm`, so the package building for darwin/arm64 *is* the check.

**Node order is DFS post-order with children sorted by ID** — terminals first, target last. Deterministic, and also the tab order the tree-canvas handoff asks for.

## The fixture is a placeholder, and says so

It carries `extracted: false` and `game_version: "community-2026-08"`, transcribed from the design handoff and prototype rather than decoded from MBIN files. Per SPEC-0001's version-pin requirement, `TestFixtureGameVersionIsPinned` fails loudly with "the Tier 1 data changed" rather than silently drifting. **It gets replaced by real extraction output when the spike runs on Linux.**

Two transformations applied to the prototype data, both worth checking:

1. **Edge labels converted to per-unit quantities.** The prototype's labels are aggregate totals at quantity 1 (`E('sul','ec','×500')`), but the spec requires per-unit quantities multiplied by parent totals — so 250 per unit of Enriched Carbon, of which 2 are needed. The reconciliation checks out in every branch, which is the strongest evidence the fixture is faithful.

2. **Carbon and a Condensed Carbon refine recipe added, deliberately outside the 34-node tree.** The spec requires that a raw-resolved node not expand "even when the Tier 1 artifact contains a recipe that could produce it" — and nothing in the original fixture was both mineable and craftable, so that scenario had nothing to bite on. Condensed Carbon is exactly that item in the real game (mineable, and refined 2:1 from Carbon). It defaults to `raw`, so the graph stays at 34 nodes.

## What this is not

There is no canvas on screen yet — no React app, no WASM bridge, no `cmd/`. Stage 1 is the data layer the canvas renders. The next increment is the Vite + React Flow + elkjs scaffold from ADR-0004 plus the WASM adapter marshalling `PlanInput` in and `ResolvedGraph` out.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/jonstump/nms-base-planner
+
+go 1.26.6
--- a/internal/domain/errors.go
+++ b/internal/domain/errors.go
@@ -1,0 +1,39 @@
+// Package domain holds the planner's computational core: dependency graph
+// resolution, producer rollup, and power budgeting.
+//
+// Governing: ADR-0003 (Go/WASM domain core). This package MUST NOT import
+// syscall/js — that boundary is what lets the ingestion CLI share it and what
+// makes it testable under plain `go test`. If it erodes, the decision has
+// failed even if the app works.
+package domain
+
+import "errors"
+
+// Sentinel errors for the domain-specific failure modes callers need to
+// distinguish programmatically.
+//
+// Governing: SPEC-0001 REQ "Error Handling Standards" — "Sentinel errors MUST
+// be defined for domain-specific failure modes that callers need to
+// distinguish programmatically — at minimum: unknown item, illegal method,
+// cycle detected, and missing constant".
+var (
+	// ErrUnknownItem reports an item ID absent from the Tier 1 artifact.
+	ErrUnknownItem = errors.New("unknown item")
+
+	// ErrIllegalMethod reports a method requested for a node that has no
+	// recipe of that kind, or a method outside the vocabulary entirely.
+	ErrIllegalMethod = errors.New("illegal method")
+
+	// ErrCycleDetected reports that method selection made a node its own
+	// ancestor.
+	ErrCycleDetected = errors.New("cycle detected")
+
+	// ErrMissingConstant reports an absent Tier 2 economy constant. Stage 1
+	// never returns it; it is defined here so the sentinel set is complete
+	// and stable across stages.
+	ErrMissingConstant = errors.New("missing constant")
+
+	// ErrInvalidArtifact reports a Tier 1 artifact that failed structural
+	// validation at load time.
+	ErrInvalidArtifact = errors.New("invalid tier 1 artifact")
+)
--- a/internal/domain/resolve.go
+++ b/internal/domain/resolve.go
@@ -1,0 +1,302 @@
+package domain
+
+import (
+	"fmt"
+	"math/big"
+	"sort"
+	"strings"
+)
+
+// PlanInput is the immutable input to stage 1.
+//
+// Governing: SPEC-0001 design.md "Pure function pipeline over an immutable
+// plan input" — no internal mutable state persists between calls, so
+// determinism falls out for free and the WASM boundary marshals one value in
+// and one value out.
+type PlanInput struct {
+	// Target is the item ID being planned.
+	Target string
+
+	// Quantity is how many of Target are wanted. Must be positive.
+	Quantity int64
+
+	// Methods overrides the per-item default method. Items absent from the
+	// map use their Tier 1 default.
+	Methods map[string]Method
+}
+
+// Edge is one resolved dependency, carrying the per-unit-of-parent quantity.
+type Edge struct {
+	// From is the consuming (parent) item ID.
+	From string
+	// To is the consumed (child) item ID.
+	To string
+	// PerUnit is how many To are needed per single unit of From.
+	PerUnit int64
+}
+
+// Node is one item in the resolved graph, with its aggregated total.
+type Node struct {
+	ItemID string
+	Name   string
+
+	// Method is the single resolved method for this node.
+	Method Method
+
+	// LegalMethods lists every method available to this item, sorted, so the
+	// view can render unavailable options as inert rather than hiding them.
+	LegalMethods []Method
+
+	// Terminal reports that expansion stopped here (method raw).
+	Terminal bool
+
+	// Children are the edges to this node's inputs, sorted by child item ID.
+	// Empty for terminal nodes.
+	Children []Edge
+
+	// Verified is false when this node's total is derived from anything
+	// marked unverified in the Tier 1 artifact — the node itself, its
+	// resolved recipe, or any ancestor.
+	//
+	// Governing: SPEC-0001 REQ "Provenance Propagation".
+	Verified bool
+
+	total *big.Rat
+}
+
+// Total returns the aggregated required quantity as an exact rational.
+//
+// Governing: SPEC-0001 REQ "Exact Arithmetic and Rounding Discipline" — "The
+// engine MUST NOT accumulate binary floating-point error across the graph."
+func (n *Node) Total() *big.Rat { return new(big.Rat).Set(n.total) }
+
+// TotalInt returns the total as an integer, reporting whether it was exact.
+func (n *Node) TotalInt() (int64, bool) {
+	if !n.total.IsInt() {
+		return 0, false
+	}
+	return n.total.Num().Int64(), true
+}
+
+// ResolvedGraph is stage 1's output.
+type ResolvedGraph struct {
+	// Target is the planned item ID.
+	Target string
+	// Quantity is the planned target quantity.
+	Quantity int64
+	// GameVersion is copied from the Tier 1 artifact, so a fixture failure is
+	// attributable to changed game data rather than to an engine regression.
+	GameVersion string
+
+	// Nodes are in topological order: terminals first, target last. This is
+	// the tab order the tree canvas handoff specifies.
+	Nodes []*Node
+
+	byID map[string]*Node
+}
+
+// Node looks up a resolved node by item ID.
+func (g *ResolvedGraph) Node(id string) (*Node, bool) {
+	n, ok := g.byID[id]
+	return n, ok
+}
+
+// Leaves returns the terminal nodes, in graph order.
+func (g *ResolvedGraph) Leaves() []*Node {
+	var out []*Node
+	for _, n := range g.Nodes {
+		if n.Terminal {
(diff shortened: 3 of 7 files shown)


jonstump (commented):
## Review: SPEC-0001 stage 1

> Submitted as a **comment** rather than `REQUEST_CHANGES` only because GitHub refuses change-requests on your own PR. Both findings below are intended as blocking.

Verified against [SPEC-0001](docs/openspec/specs/rollup-engine/spec.md), ADR-0001, and ADR-0003. Conflict-marker gate: clean. CI: green.

Every claim in the description checks out independently — 20 tests, 89.6% statement coverage, `go vet` and `gofmt` clean, `-race` clean, `go.mod` with no `require` block. The ADR-0003 boundary holds: `go list -deps` shows no `syscall/js`, and the package's only imports are `encoding/json`, `fmt`, `io`, `math/big`, `sort`, `strings`. Governing comments are present and correctly formatted throughout.

The design is right. Reverse DFS post-order gives a valid topological order on a DAG, so `propagate` visits every parent before its children — which is what makes both the aggregation and the provenance taint correct in a single pass. The error-wrapping de-duplication at the target frame is a nice detail.

Two things before merge.

---

### 1. `TotalInt()` reports out-of-range values as exact

`resolve.go:74`

```go
func (n *Node) TotalInt() (int64, bool) {
	if !n.total.IsInt() {
		return 0, false
	}
	return n.total.Num().Int64(), true
}
```

`big.Rat.IsInt()` reports that the denominator is 1 — it says nothing about int64 range. `big.Int.Int64()` is documented as undefined when the value does not fit. So an out-of-range total returns a wrapped value flagged as exact.

Reproduced against this branch:

```
total=9223372036854775808  TotalInt()=(-9223372036854775808, exact=true)
```

A **negative** count, reported as exact. This is the silent-wrong-answer shape that SPEC-0001 REQ "Exact Arithmetic and Rounding Discipline" exists to prevent, and it contradicts REQ "Error Handling Standards" on silent swallowing — the `bool` is precisely the channel that should have caught it.

Reachability is low: `Quantity` is an unbounded positive `int64`, and the deepest multiplier in this fixture is 500, so it needs a quantity around 1.8e16. But plan state is URL-hash-serialized, so the quantity is attacker- or typo-reachable rather than purely internal, and the failure is silent rather than loud.

One-line fix:

```go
if !n.total.IsInt() || !n.total.Num().IsInt64() {
	return 0, false
}
```

Worth a case in `TestRationalArithmeticIsExact` asserting `exact == false` above the boundary. This also lifts `TotalInt`'s 66.7% function coverage, which is the lowest in the package and is pointing at exactly this branch.

---

### 2. The `cook` method has no test and no fixture data

SPEC-0001 REQ "Method Resolution" carries this scenario:

> **Scenario: Cook method expands a nutrient processor recipe**
> **WHEN** a node resolves to method `cook` and the Tier 1 artifact holds a nutrient processor recipe for it
> **THEN** the node gains child edges for that recipe's inputs, and its quantities propagate exactly as for `craft` and `refine`

That scenario is stage-1 scope — method resolution and expansion, no Tier 2 constants involved. It currently has no coverage:

```
'cook' in resolve_test.go:  0
'cook' in the fixture:      0
MethodCook outside tier1.go: (none)
```

`cook` is in the vocabulary and `validMethods`, so it resolves and expands through the same path as `craft` — but nothing exercises it, and a regression in that path would pass the suite silently. The description's "one test per stage-1 spec scenario" is accurate for every scenario except this one.

The Stasis Device tree is industrial, so this fixture genuinely cannot exercise it. The smallest honest fix is a second small fixture with one nutrient-processor recipe — which also derisks the CAKE target that the base-planner handoff already specifies.

---

### Not blocking

- `Node.Children` and `Node.LegalMethods` are exported slices callers can mutate. `LegalMethods()` on `Tier1` already returns a defensive copy; the `Node` fields do not. Idiomatic Go often lives with this — noting it only because the rest of the package is careful about exactly this.
- No missing ADR references found. qmd surfaced ADR-0003 (93%) and ADR-0001 (39%); both are already cited in governing comments. ADR-0002 (50%) is save import, which this PR correctly does not touch.

jonstump:
Both findings addressed in 7ca956e.

**1. `TotalInt()` out-of-range** — added the `IsInt64()` check, so the `bool` now means what callers assume. The repro from the review (`2^63` → `-9223372036854775808, exact=true`) is now a table case in `TestTotalIntRejectsOutOfRange`, covering in-range, exactly `MaxInt64`, one past it, and fractional. It also asserts the returned value is `0` whenever `exact` is false, so a caller ignoring the bool gets an obviously-wrong zero rather than a plausible-looking negative. Function coverage 66.7% → 100%.

**2. `cook` untested** — added `testdata/hexaberry-cake.tier1.json`, built on the CAKE target the base-planner handoff already specifies. It is a multi-level chain (cake → batter/butter → flour → wheat) rather than a single recipe, so propagation through cook edges is covered alongside expansion, and `TestCookProvenanceTaint` confirms taint flows through cook the same as craft and refine.

The fixture carries the same caveats as the Stasis one — `extracted: false`, pinned game version, and a note stating its quantities are illustrative and that what is asserted against it is *behaviour*, never real recipe values. It is not a claim about the real cake recipe.

20 → 27 tests, coverage 89.6% → 90.3%. `go vet`, `gofmt`, and `-race` clean.

The non-blocking note about `Node.Children` and `Node.LegalMethods` being mutable exported slices is left as-is — worth revisiting when the WASM adapter lands, since that is the point where an external caller could actually reach them.

jonstump (commented):
## Review round 2 (7ca956e)

Conflict-marker gate: clean. CI: green. Mergeable: clean.

**Both round-1 findings verified fixed**, re-run against this head rather than taken on trust:

| Finding | Verification |
|---|---|
| `TotalInt()` out-of-range | `2^63` now returns `(0, exact=false)` — was `(-9223372036854775808, exact=true)`. Returns 0 rather than a wrapped value, so a caller ignoring the bool gets an obviously-wrong zero. Function coverage 66.7% → 100%. |
| `cook` unexercised | `TestCookMethodExpandsNutrientProcessorRecipe` and `TestCookProvenanceTaint` both pass against the new cake fixture. |

Suite: 27 tests, 90.3% coverage, `go vet` / `gofmt` / `-race` clean.

The cake fixture is the right shape — a multi-level chain rather than a single recipe, so it covers propagation through cook edges and not just expansion, and it carries the same `extracted: false` and pinned-version honesty as the Stasis one.

---

### New finding: the two fixtures share one version constant

`resolve_test.go:18`

```go
const fixtureGameVersion = "community-2026-08"
```

It is asserted against **both** fixtures — Stasis at lines 60 and 81, cake at line 509. Both currently read `community-2026-08`, so it passes today.

It breaks at the next milestone. When the extraction spike replaces `stasis-device.tier1.json` with a real extracted artifact, its `game_version` becomes a real game version and the constant has to change to match — at which point **every cake test fails**, despite the cake fixture being untouched and still a perfectly valid placeholder at `community-2026-08`.

That inverts the purpose of the pin. SPEC-0001 REQ "Dependency Graph Resolution" wants a failure to read as *"the Tier 1 data changed"* for the fixture that actually changed. A shared constant makes one fixture's re-extraction masquerade as a data change in the other, which is exactly the misattribution the requirement exists to prevent.

Fix is one line each:

```go
const stasisFixtureGameVersion = "community-2026-08"
const cakeFixtureGameVersion   = "community-2026-08"
```

Same values now, independently movable later.

**Related asymmetry**: `TestFixtureGameVersionIsPinned` also asserts `!a1.Extracted`, guarding that the Stasis placeholder stays honestly labelled. The cake fixture has no equivalent — `loadCakeFixture` checks the version but not the `extracted` flag. Worth mirroring so both placeholders are held to the same standard.

---

### On the standing of this review

The round-1 findings were fixed by the same session that raised them, and this round was performed by that session too — so nothing independent has checked that work. Treat the two verifications above as *"the stated defect no longer reproduces"*, which I did confirm by execution, rather than as an independent audit of the fix.

This is also why the review is a comment: GitHub refuses both `APPROVE` and `REQUEST_CHANGES` on your own PR, so `/sdd:review` cannot produce the approval its own merge rule requires. **Not merged.** The merge decision is yours.

PR #12, opened 2026-08-18
CI: static analysis, race detection, and formatting

jonstump:
Closes #8
Part of #7

The repository had no CI at all. PRs #2 and #5 each reported "vet / gofmt / -race clean" from a local run — accurate at the time, but nothing enforced it, and nothing would have caught a regression.

## Three independent jobs

Split so a formatting failure and a test failure surface separately rather than one masking the other.

| Job | Check |
|---|---|
| Formatting | `gofmt -l .` — fails with the offending file list and the fix command |
| Static analysis | `go vet ./...` then `staticcheck ./...` |
| Tests (race) | `go test -race ./...` |

Runs on every pull request and on pushes to `main`. `permissions: contents: read` — the workflow needs nothing else.

## The staticcheck version pin is deliberate

`go.mod` requires go 1.26.6, and staticcheck refuses to analyze a module built for a newer Go than the analyzer itself:

| Version | Built with | Result |
|---|---|---|
| `2025.1.1` (v0.6.1) | go1.24.13 | ❌ `module requires at least go1.26.6` |
| `v0.7.0` | go1.25.13 | ❌ `package requires newer Go version go1.26` |
| `v0.8.0-rc.1` | go1.26 | ✅ exits 0, no findings |

So the pin is to a release candidate. That is a real trade — I'd rather pin an RC that works than ship a job that fails on every run, and the alternative was dropping static analysis beyond `go vet` entirely. The workflow carries a comment saying to move to `v0.8.0` final once it ships.

## Verified locally

All three checks were run in the worktree before opening this PR:

- `gofmt -l .` — no output
- `go vet ./...` — clean
- `staticcheck ./...` (v0.8.0-rc.1) — exit 0, no findings
- `go test -race ./...` — `internal/domain` ok, `internal/psarc` ok

## Note on the game-files constraint

Per [ADR-0001](docs/adrs/ADR-0001-two-tier-nms-data-ingestion.md), ingestion is developer-local and cannot run in CI. The test job deliberately does not set `NMS_PCBANKS`, so the opt-in full-archive tests that [SPEC-0003](docs/openspec/specs/hgpak-reader/spec.md) REQ "Real-Archive Verification" specifies must **skip** rather than fail. That requirement is implemented in #9; this workflow is the environment it has to skip cleanly in.

## Also here

`.gitignore` gains `.claude/worktrees/` — the scratch directory `/sdd:work` creates for agent worktrees, which otherwise shows up as untracked in the main tree.

[...]

Issue #11, opened 2026-08-18
Safe extraction and CLI pipeline acceptance


jonstump:
Part of #7. Governed by [SPEC-0003](docs/openspec/specs/hgpak-reader/spec.md).

**Blocked by #10** — needs path resolution and selective extraction in place.

Writes entries to disk safely, rewires the CLI, and runs the acceptance test for the whole ingestion stage. Extends `cmd/nmsextract/main.go` (currently `list`/`extract` with a substring filter over `internal/psarc`) to use the new reader.

Per the SPEC-0003 migration plan, **this story deletes `internal/psarc`** — in the same PR that proves the replacement works, not before. It has no other consumer and no value as reference.

## Requirements

- [ ] **Safe Extraction to Disk** (SPEC-0003) — Archive paths are untrusted input. When extracting to a directory, the reader MUST reject any entry whose resolved destination falls outside that directory, including absolute paths, paths containing `..`, and paths traversing a symlink out of the tree. A rejected path MUST fail the extraction with an error naming it; the reader MUST NOT silently skip it. The reader MUST create parent directories as needed, and MUST tolerate a leading separator on an archive path by treating it as relative.
  - WHEN an entry's path escapes the output directory THEN extraction fails naming that path, and nothing is written outside the directory
  - WHEN an entry at `metadata/reality/tables/costtable.mbin` is extracted THEN the intermediate directories are created and the file lands at the mirrored path

- [ ] **Pipeline Fitness** (SPEC-0003) — Extraction MUST produce byte-exact entry contents; the reader performs no transformation beyond decompressing the containing blocks. The CLI MUST support filtering by path substring on both listing and extraction, since a single archive holds tens of thousands of entries. Extraction of `metadata/reality/tables/` from `NMSARC.Precache.pak` MUST yield files beginning with MBIN magic `cccccccc` that MBINCompiler decompiles without error. This is the acceptance test for the whole stage.
  - WHEN `metadata/reality/tables/` is extracted from `NMSARC.Precache.pak` THEN all 54 tables are written, each begins with `cccccccc`, and MBINCompiler decompiles `nms_reality_gcproducttable.mbin` to EXML without error
  - WHEN a 47,000-entry archive is listed with a path filter THEN only matching paths are printed, and the count of matches is reported

## Acceptance Criteria

- Path traversal is refused with a named error, verified by a test using a crafted archive
[...]

Issue #10, opened 2026-08-18
Manifest, path resolution, and selective extraction


jonstump:
Part of #7. Governed by [SPEC-0003](docs/openspec/specs/hgpak-reader/spec.md).

**Blocked by #9** — builds on the `Archive` type, block reader, and sentinels from that story.

Turns a parsed container into something addressable by path. Entry 0 is a plaintext manifest, so names come from the archive alone with no external hash mapping. The fixed 65,536-byte decompressed block size makes stream position to block index pure arithmetic, which is what keeps a 26 KB read from costing a 565 MB decompression.

These two requirements share the manifest bytes and the block-index arithmetic, so they belong in one story.

## Requirements

- [ ] **Manifest and Path Resolution** (SPEC-0003) — The reader MUST treat entry 0 as the path manifest: CRLF-separated lowercase paths, where manifest entry *n* names archive entry *n*, so it MUST hold exactly `entryCount - 1` non-empty paths. The reader MUST expose every entry by its path, MUST NOT require the caller to supply a hash, and MUST NOT depend on any external hash-to-name mapping. It MUST verify that an entry's 16-byte hash equals the MD5 of its lowercase path and report a mismatch as malformed rather than silently preferring one of the two. A manifest whose path count disagrees with `entryCount - 1` MUST be reported as malformed.
  - WHEN an archive is opened with no external data available THEN every entry is listed by its full path
  - WHEN the manifest holds a number of paths other than `entryCount - 1` THEN the read fails naming both counts
  - WHEN each entry's path is hashed with MD5 THEN the digest equals the entry's stored hash, and any mismatch fails the read

- [ ] **Selective Extraction** (SPEC-0003) — Because block size is fixed at 65,536 decompressed bytes, an entry at stream position `P` begins in block `P / 65536`. The reader MUST use this to decompress only the blocks covering a requested entry, and MUST NOT decompress the whole archive to read one file. The reader MUST expose listing without extraction, so a caller can enumerate paths having decompressed only the blocks covering the manifest. Entry contents MUST be read on demand rather than at open — `NMSARC.MetadataEtc.pak` decompresses to roughly 565 MB, which MUST NOT be required resident to read one table.
  - WHEN the paths of a 47,000-entry archive are listed THEN only the blocks covering the manifest were decompressed
  - WHEN a single 26 KB entry is extracted from a 47 MB archive THEN only the blocks spanning that entry were decompressed

## Acceptance Criteria

[...]

Issue #7, opened 2026-08-18
Implement HGPAK Archive Reader


jonstump:
Stage 1 of the [ADR-0001](docs/adrs/ADR-0001-two-tier-nms-data-ingestion.md) ingestion pipeline: unpack the game's `.pak` archives into `.MBIN` files that MBINCompiler can decompile.

## Spec

- [spec.md](docs/openspec/specs/hgpak-reader/spec.md) — SPEC-0003, 8 requirements
- [design.md](docs/openspec/specs/hgpak-reader/design.md)

## Governing ADRs

- **ADR-0001** — two-tier NMS data ingestion; this epic implements the unpack step of Tier 1

## Why this exists

`internal/psarc` (merged in #5) implements PlayStation Archive and can open no file in the game install. All 97 archives under `GAMEDATA/PCBANKS` are `HGPAK` containers. That reader was written against an unverified label on a diagram edge, and its tests validated its own writer against its own reader — so a fully green suite proved only self-consistency. SPEC-0003 replaces it, and leads with a real-archive verification requirement precisely so that failure mode cannot recur.

## Stories

| # | Title | Branch | Depends on |
|---|-------|--------|-----------|
| #8 | CI: static analysis, race detection, formatting | `feat/8-ci-static-analysis-race-formatting` | — |
| #9 | Fixtures and container parsing | `feat/9-fixtures-and-container-parsing` | — |
| #10 | Manifest, path resolution, and selective extraction | `feat/10-manifest-path-resolution-selective-extract` | #9 |
| #11 | Safe extraction and CLI pipeline acceptance | `feat/11-safe-extraction-and-cli-acceptance` | #10 |

Foundation stories #8 and #9 merge before the rest begins. #10 and #11 are strictly sequential — each builds directly on the previous story's types.

## Requirement coverage

All 8 SPEC-0003 requirements are covered, each in exactly one story:

| Requirement | Story |
|---|---|
| Real-Archive Verification | #9 |
| Container Identification | #9 |
| Structural Layout | #9 |
| Error Handling Standards | #9 |
| Manifest and Path Resolution | #10 |
| Selective Extraction | #10 |
| Safe Extraction to Disk | #11 |
| Pipeline Fitness | #11 |

## Acceptance

[...]

PR #13, opened 2026-08-18
HGPAK container parsing with a real-archive fixture

### LINKED ISSUE #9: Fixtures and container parsing

The foundation story: the committed real-archive fixture, the `HGPAK` header/table/block parser, and the error sentinels every later story returns. Establishes `internal/hgpak` and adds the zstd module dependency (the project is stdlib-only today; there is no `compress/zstd`).

New package — no existing code to extend. `internal/psarc` is left untouched and unreferenced; it is deleted in the story that proves the replacement works.

**The fixture is the point of this story.** Per the design doc it must span at least two blocks, because the two bugs most likely to recur — forgetting 16-byte block alignment, and treating entry offsets as stream-relative — both survive any synthetic fixture built from the same misunderstanding.

## Requirements

- [ ] **Real-Archive Verification** (SPEC-0003) — The reader MUST NOT be considered implemented until it has been run against an unmodified archive from a real game install. The test suite MUST include at least one fixture derived from a real archive; a suite exercising the reader only against archives built by the project's own helpers does not satisfy this, regardless of coverage. The committed fixture MUST be a structurally faithful excerpt carrying real header layout, block framing, alignment, and manifest formatting. An opt-in path (an environment variable naming a PCBANKS directory) MUST exercise a full archive and skip when unset.
  - WHEN the test suite runs with no game install available THEN it opens the committed real-archive excerpt, lists its manifest, and extracts its entries
  - WHEN the environment variable naming a PCBANKS directory is set THEN the suite opens every archive in it and verifies each one's manifest and entry count, AND WHEN it is unset those tests report as skipped rather than passing
  - WHEN every archive a test opens was produced by the project's own helper THEN the requirement is not met, even at full statement coverage

- [ ] **Container Identification** (SPEC-0003) — The reader MUST verify the 8-byte magic `HGPAK\0\0\0` before interpreting any other field, and MUST reject a mismatch with an error naming the magic found. It MUST reject an unimplemented version rather than attempting a best-effort parse. All multi-byte integers MUST be decoded as little-endian unsigned 64-bit, except the opaque 16-byte entry hash.
  - WHEN a PSARC archive is opened THEN the read fails naming the magic found, and no entry table is parsed
  - WHEN the version field holds a value this reader does not implement THEN the read fails naming the version, rather than parsing with version 2 rules

- [ ] **Structural Layout** (SPEC-0003) — The reader MUST parse a 0x30-byte header; an entry table of `entryCount` 32-byte records at 0x30 (16-byte hash, u64 offset, u64 size); a block table of `blockCount` u64 compressed lengths; then the blocks. Each block MUST decompress with zstd to exactly 65,536 bytes except a possibly shorter final block, and any other length MUST be treated as malformed. Blocks MUST be located by accumulating compressed lengths from the data-start offset, advancing to the next 16-byte boundary after each. Entry offsets MUST be interpreted as positions in a virtual image of the file, so stream position is `entry.offset - dataStart`.
  - WHEN an archive with more than one block is read THEN every block decompresses successfully, and the total decompressed length equals the block count times 65,536 less any short final block
  - WHEN the first entry of an archive is read THEN it resolves to stream position zero, not to position `dataStart`
  - WHEN an entry's extent runs past the end of the decompressed stream THEN the read fails naming the entry, rather than returning a short buffer

- [ ] **Error Handling Standards** (SPEC-0003) — Errors MUST be wrapped with contextual information at each layer boundary. Sentinel errors MUST be defined for: not an HGPAK archive, unsupported version, malformed archive, entry not found, and unsafe extraction path. Silent error swallowing MUST NOT occur. Structured logging MUST be used for error reporting (key-value pairs, not string interpolation).
  - WHEN a caller opens a PSARC file, versus an HGPAK whose blocks will not decompress THEN the two failures carry different sentinels
  - WHEN extraction fails on one entry of many THEN the error names that entry's path and the structural expectation that was violated

## Acceptance Criteria

- A real-archive excerpt spanning at least two blocks is committed and opened by the default test run
- The full-archive test skips cleanly when its environment variable is unset
- Omitting 16-byte block alignment makes the suite fail (this is the regression the fixture exists to catch)
- Treating entry offsets as stream-relative rather than virtual makes the suite fail
- All five sentinels are defined and distinguishable by callers

### Branch

`feat/9-fixtures-and-container-parsing`

### PR Convention

```
Closes #9
Part of #7

Implements SPEC-0003 REQ "Real-Archive Verification", REQ "Container
Identification", REQ "Structural Layout", REQ "Error Handling Standards".
```

jonstump:
Closes #9
Part of #7

Implements SPEC-0003 REQ "Real-Archive Verification", REQ "Container Identification", REQ "Structural Layout", REQ "Error Handling Standards".

Adds `internal/hgpak`. `internal/psarc` is untouched and unreferenced — it is deleted in #11, in the PR that proves the replacement works.

## The fixture is the point of this story

`testdata/excerpt.pak` is a **real-archive excerpt**, not a synthetic one: 27 verbatim entries — real bytes, real paths — from `NMSARC.globals.pak` of a real install, assembled by `testdata/gen.go` to mirror that archive's layout exactly. 21.7 KB on disk, 72,503 bytes decompressed.

It spans **two blocks** deliberately. The two mistakes most likely to recur are invisible in a single-block fixture:

- omitting the 16-byte block alignment (a reader that just accumulates compressed lengths lands mid-frame on block 1)
- treating entry offsets as stream-relative rather than as offsets into a virtual image of the file

Both now fail the suite loudly. `TestBlockAlignmentIsHonoured` and `TestEntryOffsetsAreVirtualNotStreamRelative` assert them by name rather than relying on a smoke test.

## The opt-in test found two variants the spec does not describe

This is the part worth reviewing. `TestFullArchivesFromRealInstall` is gated on `NMS_PCBANKS` and skips in CI. Running it against all 97 archives failed immediately, twice, on assumptions I had written into both the spec and the ADR:

**1. Header `0x20` is a storage flag, not an unknown field.**

SPEC-0003 and ADR-0001 both say "an unknown field (1 in every archive observed)". That was true of the two archives I had parsed in full, and false in general. It is `1` for a zstd block stream (95 archives) and `0` for **stored** archives — `NMSARC.audio.pak` and `NMSARC.audioBNK.pak` — where there is no block table between the entry table and the data, `dataStart` is exactly `headerSize + entryCount*32`, entry offsets are direct file offsets, and entry bytes including the manifest sit uncompressed. Audio archives carry already-compressed WEM/BNK payloads, so a second pass would buy nothing.

**2. A block whose compressed length is exactly 65536 is stored raw.**

Eight archives use this — `NMSARC.UI.pak` and the `TexBiomes*` family — and their block 0 is plaintext manifest text where a zstd frame should be. The reader treats `blockLen == BlockSize` as stored.

That rule is a **length comparison, not a magic sniff**, which matters: the design doc explicitly rejects PSARC's inflate-and-fall-back, and this does not reintroduce it. A block that is not exactly `BlockSize` long and then fails to decompress is still a hard `ErrMalformed`.

**Result: 97 of 97 archives parse**, with manifest counts and the MD5-of-lowercase-path hash convention verified on each.

SPEC-0003's own Open Questions asked "do any archives use a compression method other than zstd, or a block size other than 65536? ... only two have been parsed in full — the opt-in test across all of them is what answers this." This is that answer, and it arrived the first time the test ran.

`docs/openspec/specs/hgpak-reader/` and `docs/adrs/` are protected paths, so these are recorded here rather than edited on this branch:

- `docs/openspec/specs/hgpak-reader/spec.md`: REQ "Structural Layout" must describe the storage flag at `0x20` and both layouts, and must state the raw-block rule (`compressed length == 65536` means stored). Its Open Questions entry about compression uniformity is now answered and should be removed.
- `docs/openspec/specs/hgpak-reader/design.md`: the "unknown header field at 0x20" risk is resolved — it is a storage selector. The contrast-with-PSARC decision should note that raw blocks exist but are identified by length, not by sniffing.
- `docs/adrs/ADR-0001-two-tier-nms-data-ingestion.md`: the HGPAK note says "an unknown field (1 in every archive observed)" — now known to be a storage flag with two observed values.

## Scope boundary

This story parses structure. `Manifest()` returns entry 0's raw bytes; resolving entries **by path** — the count check, the MD5 verification as a first-class guarantee — is #10. The hash convention is asserted here anyway (`TestEntryHashesAreMD5OfLowercasePath`) because it is a property of the container this story parses, and the fixture would be the wrong place to discover it broken later.

`ReadEntry` already decompresses only the blocks an entry spans, since that is the natural implementation of the block-index arithmetic. #10 adds the guarantees and the counter-based tests that prove it, plus listing without extraction.

## Error handling

Five sentinels per REQ "Error Handling Standards": `ErrNotHGPAK`, `ErrUnsupportedVersion`, `ErrMalformed`, `ErrEntryNotFound`, `ErrUnsafePath`. The last is defined here but returned by #11's extraction layer, so the whole vocabulary lives in one place.

`StructureError` exposes `Op`/`Index`/`Want`/`Got` as fields with a `LogAttrs()` helper, so a structured logger emits key-value pairs rather than scraping a formatted message. The package itself does not log — it returns wrapped errors, which is the "returned to the caller" branch of the requirement. The `slog` call sites land with the CLI in #11.

## Verification

| Check | Result |
|---|---|
| `go test ./internal/hgpak/` | 17 pass, 1 skip (the `NMS_PCBANKS` gate) |
| `NMS_PCBANKS=... go test` | 97/97 archive subtests pass |
| `go vet ./...` | clean |
| `staticcheck` (v0.8.0-rc.1) | exit 0, no findings |
| `gofmt -l .` | no output |

Depends on #12 for those checks to run in CI; it is independent of this branch and can merge in either order.

## Honest limits

- **Uncertainty moved, it did not disappear.** Two variants turned up the first time the reader met all 97 archives. Everything here still comes from one game build, and the next update can move it again — which is why unrecognized storage flags and wrong-length blocks fail loudly instead of being guessed at.
- **`blockCount` is unparsed for stored archives.** It is non-zero (1017 for `audioBNK`) but no block table exists in the region between the entry table and the data. I did not chase where it points; nothing in the ADR-0001 pipeline needs it. Noted in the code.
- **A stored final block is theoretically ambiguous.** For the last block, `blockLen` equals the decompressed length whether it is stored or compressed, so the `== BlockSize` rule cannot distinguish the two. No archive in the install exhibits it, so it is unhandled rather than speculatively coded.


jonstump (commented):
## Review: Changes requested

Spec-aware review against issue #9 acceptance criteria and SPEC-0003. (Submitted as a comment review — GitHub blocks `REQUEST_CHANGES` on a self-authored PR. Treat this as changes-requested.)

The substance of this PR is right, and the part the body flags as most worth reviewing is the part that holds up best. Running the opt-in test against all 97 archives and finding two undocumented variants — the `0x20` storage flag and the `blockLen == 65536` raw-block rule — is exactly the failure mode SPEC-0003 was written to prevent recurring, caught the first time the reader met real data. The `== BlockSize` length rule genuinely is not a magic sniff, and the distinction is worth the paragraph it got.

Verified locally on `a0372cb`: `go test -race ./...` passes, `go vet` clean, `gofmt -l .` empty.

**Three findings below. Two are confirmed panics — reproduced, stack traces included.**

---

### 1. Blocking — entry `Size` overflows the extent check, panicking instead of returning `ErrMalformed`

`hgpak.go:376` computes `end := pos + e.Size` in `uint64`. Both operands come straight from the entry table with no upper bound on `e.Size`, so the addition wraps and the guard on the next line passes for an entry that is nowhere near valid.

```go
pos := a.StreamPos(e)
end := pos + e.Size     // wraps when e.Size is near 2^64
if end > a.streamLen {  // defeated by the wrap
```

Execution then reaches `out := make([]byte, 0, e.Size)` at `hgpak.go:390` (or `make([]byte, e.Size)` at :382 on the stored path).

Reproduced against the committed fixture — entry 1 has `pos=848`, so setting its `Size` to `^uint64(0)-848+17` wraps `end` to 16:

```
Open succeeded; calling ReadEntry(1) with Size= 18446744073709550784
panic: runtime error: makeslice: cap out of range
  hgpak.(*Archive).ReadEntry  internal/hgpak/hgpak.go:390
```

This defeats REQ "Structural Layout" scenario *"A truncated archive is refused"* — the spec requires the read to fail naming the entry; it panics instead. It also violates REQ "Error Handling Standards", which requires every error to be **returned to the caller**. A panic is not a returned error, and it takes the sentinel vocabulary with it: no caller can `errors.Is` its way out of a dead process.

Suggested fix — compare against the remaining stream rather than computing a sum that can wrap:

```go
if e.Size > a.streamLen-pos {
    return nil, malformed("entry extent", i,
        fmt.Sprintf("<= %d bytes remaining from stream position %d", a.streamLen-pos, pos), e.Size)
}
```

`pos <= a.streamLen` is already guaranteed by the `e.Offset < a.dataStart` check plus a stream-length bound, so the subtraction is safe in that order.

### 2. Blocking — block length `>= 2^63` sign-flips the extent check, panicking inside `Open`

Same class, earlier in the read path, and worse because it fires during `Open` rather than on demand. `hgpak.go:286`:

```go
if pos+int64(n) > a.size {   // n is uint64; int64(n) is negative when n >= 2^63
```

A block length of `2^63` converts to a negative `int64`, the comparison passes, and `readTables` stores it. `readTables` then probes the final block to settle `streamLen`, reaching `raw := make([]byte, a.blockLen[i])` at `hgpak.go:321`:

```
panic: runtime error: makeslice: len out of range
  hgpak.(*Archive).block      internal/hgpak/hgpak.go:321
  hgpak.(*Archive).readTables internal/hgpak/hgpak.go:303
  hgpak.Open                  internal/hgpak/hgpak.go:232
```

There is a second, quieter symptom on the same line. Putting the huge length on a **non-final** block instead pushes `pos` negative, and the failure surfaces as:

```
reading block 1 at -9223372036854774848: bytes.Reader.ReadAt: negative offset
```

That error wraps the raw `ReadAt` failure and **not** `ErrMalformed`, so `errors.Is(err, ErrMalformed)` is `false` for an archive whose block table is plainly self-contradictory. REQ "Error Handling Standards" requires malformed structure to be distinguishable by sentinel; here a caller sees an I/O error and would reasonably conclude the disk is at fault rather than the file.

Suggested fix — bound `n` in unsigned space before any `int64` conversion:

```go
if n > uint64(a.size)-uint64(pos) {
    return malformed("block extent", i,
        fmt.Sprintf("<= %d bytes remaining", uint64(a.size)-uint64(pos)), n)
}
```

Worth noting these two are the same oversight the package already guards against one layer up — `Open` explicitly bounds `entryCount` and `blockCount` before allocating, with the comment *"a corrupt count must not turn into a multi-gigabyte make()"*. That instinct is correct and just needs extending to the per-entry `Size` and per-block length fields, which are equally file-controlled and currently unbounded.

On threat model: these are the user's own game files, not hostile input, so this is not a security finding. It matters because a truncated download or a post-update format change should surface as `ErrMalformed` naming the entry — which is what the spec promises and what the CLI in #11 will need to print — rather than as a crash with no path in it.

### 3. Non-blocking — `printableMagic` names nothing when the magic starts non-printable

`printableMagic` truncates at the first byte outside `0x20..0x7e`, so a file whose first byte is already non-printable yields an empty string:

```
psarc        -> magic is "PSAR", want "HGPAK": not an HGPAK archive
gzip         -> magic is "",     want "HGPAK": not an HGPAK archive
binary-lead  -> magic is "",     want "HGPAK": not an HGPAK archive
```

REQ "Container Identification" requires rejecting a mismatch *"with an error naming the magic found"*. The PSARC case in the spec's own scenario works correctly, which is why the suite doesn't catch this — but `magic is ""` names nothing at all. A hex fallback when the printable prefix is empty would satisfy the requirement across the board.

---

### Acceptance criteria

| Criterion (issue #9) | Status |
|---|---|
| Real-archive excerpt spanning ≥2 blocks, opened by the default test run | Met — `excerpt.pak`, 27 verbatim entries from `NMSARC.globals.pak`, two blocks, `TestRealArchiveExcerpt` |
| Full-archive test skips cleanly when the env var is unset | Met — `TestFullArchivesFromRealInstall` gated on `NMS_PCBANKS`, skips in CI |
| Omitting 16-byte block alignment makes the suite fail | Met — `TestBlockAlignmentIsHonoured` asserts it by name |
| Treating entry offsets as stream-relative makes the suite fail | Met — `TestEntryOffsetsAreVirtualNotStreamRelative` |
| All five sentinels defined and distinguishable | Met as written — `TestAllSentinelsExistAndAreDistinct`, but see finding 2: one real malformed-structure path returns an I/O error instead of `ErrMalformed` |

Naming the two regression tests after the mistakes they catch rather than after the functions they exercise is the right call, and `testdata/README.md` documenting which real archive each hand-assembled variant was checked against satisfies the spec's "MUST state in a comment which real archive its structure was checked against" for the synthesized cases.

### Architecture references

ADR-0001 and SPEC-0003 governing comments are present in `hgpak.go`, `errors.go`, and `hgpak_test.go`. A qmd sweep of `nms-base-planner-adrs` and `nms-base-planner-issues` surfaced no ADR or open issue this PR should reference but doesn't.

### Deferred design-doc updates

The PR body correctly records the three artifact edits it cannot make from a protected path — the `0x20` storage flag in SPEC-0003 REQ "Structural Layout" and ADR-0001, the resolved risk in `design.md`, and the now-answered Open Question on compression uniformity. Those should land before the epic closes so the next reader of the spec isn't misled by "an unknown field (1 in every archive observed)". Worth a tracking issue rather than leaving it in a PR description.

### CI

Only GitGuardian reports on this branch — the Go jobs live in #12, which isn't on `main` yet. Merging #12 first gives this branch real CI coverage on its next push.

jonstump:
## Review feedback addressed — findings 1 and 2 fixed in 80ee4c1

Both blocking findings were the same defect at two layers: an extent check written as a sum that can wrap, letting a file-controlled `u64` reach `make()`.

### Finding 1 — entry `Size` overflow (`hgpak.go:376`) — fixed

`ReadEntry` now compares `Size` against the bytes remaining instead of computing `pos + e.Size`:

```go
if pos > a.streamLen {
    return nil, malformed("entry offset", i,
        fmt.Sprintf("<= stream length %d", a.streamLen), pos)
}
if remaining := a.streamLen - pos; e.Size > remaining {
    return nil, malformed("entry extent", i,
        fmt.Sprintf("<= %d bytes remaining from stream position %d", remaining, pos), e.Size)
}
```

The `pos > a.streamLen` guard comes first so the subtraction underneath cannot underflow.

### Finding 2 — block length sign-flip (`hgpak.go:286`) — fixed

Bounded in unsigned space, against the bytes remaining:

```go
if pos > a.size {
    return malformed("block start", i,
        fmt.Sprintf("<= file size %d", a.size), pos)
}
...
if remaining := uint64(a.size) - uint64(pos); n > remaining {
    return malformed("block extent", i,
        fmt.Sprintf("<= %d bytes remaining from offset %d", remaining, pos), n)
}
```

The `pos > a.size` check at the top of each iteration is load-bearing and not redundant with the extent bound: alignment padding after the previous block can carry `pos` past the end of the file, which would underflow the subtraction on the next pass.

This also resolves the quieter half of the finding — the earlier-block case that surfaced as `bytes.Reader.ReadAt: negative offset` without wrapping `ErrMalformed`. Both positions now return `ErrMalformed` naming `block extent`, so `errors.Is` works for a self-contradicting block table wherever the bad length sits.

### Regression tests

`TestOversizedEntryIsRefusedNotPanicked` and `TestHugeBlockLengthIsRefusedWithSentinel` (the latter subtested over both the first and final block, because they failed differently before the fix — panic on the final, wrong sentinel on the first).

Both were confirmed to **panic with the fix reverted** and pass with it, so they are real regression tests rather than assertions that happen to hold. They assert the sentinel and the named structural expectation, not merely that the call didn't crash — a panic is not a returned error, which is the actual REQ "Error Handling Standards" violation.

### Verification

| Check | Result |
|---|---|
| `go vet ./...` | clean |
| `staticcheck` v0.8.0-rc.1 | exit 0, no findings |
| `gofmt -l .` | no output |
| `NMS_PCBANKS=... go test` | **97/97 archives pass** |

That last row is the one that mattered: a bound that is too strict would have started rejecting real archives, and it rejects none. CI is also green on this branch for the first time now that #12 is on `main` — all four checks pass.

### Finding 3 left open

`printableMagic` still returns `""` for a magic starting with a non-printable byte, so the REQ "Container Identification" requirement to name the magic found is unmet for non-ASCII-leading files. Out of scope for this pass; happy to fold the hex fallback in here or leave it for a follow-up.

The deferred design-doc updates (the `0x20` storage flag in SPEC-0003 and ADR-0001, and the now-answered Open Question on compression uniformity) are also still outstanding and still want a tracking issue before the epic closes.

PR #14, opened 2026-08-18
docs: record the HGPAK storage flag and raw-block rule

jonstump:
Part of #7

Closes the design-doc drift deferred by #13. The reader on `main` parses two container layouts and a verbatim-block rule that ADR-0001 and SPEC-0003 never described — until this PR that knowledge lived only in a PR description, while the governing artifacts asserted something the merged code disproves.

That gap matters more than usual here. `internal/psarc` exists because a reader was written against an unverified label on a diagram edge, and SPEC-0003 leads with a real-archive verification requirement so it can't recur. Stories #10 and #11 will be built by agents reading these documents.

## What changed

| Artifact | Was | Now |
|---|---|---|
| `ADR-0001` | header `0x20` is "an unknown field (1)" | a storage flag, with both observed values and the archives exhibiting each |
| `spec.md` REQ "Structural Layout" | block-stream layout only | both layouts, third-value rejection, the raw-block rule, and an overflow bound |
| `design.md` Risks | "the unknown header field at 0x20 … parsed and ignored" | resolved, with the reason the mitigation worked |
| `design.md` Open Questions | two questions about `0x20` and compression uniformity | marked answered by the 97-archive run; two genuinely-open ones added |

## Notes for review

**The raw-block rule is stated as a length comparison, deliberately.** `design.md` already argues against PSARC's inflate-and-fall-back, and a careless reading of "some blocks are stored" would reintroduce exactly that. The requirement says the rule is decided *before* decompression is attempted, and that a wrong-length block failing to decompress remains a hard error.

**One addition goes beyond correcting drift.** REQ "Structural Layout" now requires file-controlled sizes to be bounded against the bytes remaining, with no overflowing sums or signed conversions, and requires a malformed archive to error rather than panic. That reflects shipped behaviour (80ee4c1) rather than new intent, and #10 adds more file-controlled values, so the constraint is worth stating before it's needed. Easy to drop if you'd rather keep this PR purely corrective.

**Every scenario added is backed by a test already on `main`** — `TestStoredArchiveLayout`, `TestUnknownStorageFlagIsRefused`, `TestRawBlockIsStoredNotCompressed`, `TestOversizedEntryIsRefusedNotPanicked`, `TestHugeBlockLengthIsRefusedWithSentinel`. Nothing here is aspirational.

**`design.md` keeps the `0x20` risk instead of deleting it.** The mitigation it asked for — assert on the field rather than silently discard it — is what surfaced the answer, and that's worth leaving legible.

## Still open

- SPEC-0003 is still `status: draft` with half its stories shipped. Not changed here since I don't know your convention for an in-progress state.
- Finding 3 from the #13 review is unfixed: `printableMagic` returns `""` for a magic starting with a non-printable byte, so REQ "Container Identification"'s "name the magic found" is unmet for non-ASCII-leading files. Code fix, not a docs fix.

Docs-only — no code touched.

### DIFF
--- a/docs/adrs/ADR-0001-two-tier-nms-data-ingestion.md
+++ b/docs/adrs/ADR-0001-two-tier-nms-data-ingestion.md
@@ -135,9 +135,10 @@ graph TD
 
 Confirmed by parsing real archives end to end:
 
-* Header is little-endian: 8-byte `HGPAK\0\0\0` magic, then u64 version (2), entry count, block count, an unknown field (1), and a data-start offset.
+* Header is little-endian: 8-byte `HGPAK\0\0\0` magic, then u64 version (2), entry count, block count, a **storage flag**, and a data-start offset.
 * An entry table of 32-byte records — 16-byte MD5 + u64 offset + u64 size — followed by a block table of u64 compressed lengths.
-* Blocks are **zstd**, each decompressing to exactly 65,536 bytes, each starting 16-byte aligned. `NMSARC.globals.pak`: 87 blocks decompressing to 5,701,632 bytes = 87 x 64 KiB exactly.
+* The storage flag selects the layout. `1` is a zstd block stream (95 archives); `0` is **stored** — no block table, entry offsets are direct file offsets, and entry bytes including the manifest sit uncompressed (`NMSARC.audio.pak`, `NMSARC.audioBNK.pak`, whose WEM/BNK payloads are already compressed). An earlier revision of this ADR called this an unknown field, "1 in every archive observed" — true of the two archives parsed at the time, and false in general. Parsing all 97 is what corrected it.
+* Blocks are **zstd**, each decompressing to exactly 65,536 bytes, each starting 16-byte aligned. `NMSARC.globals.pak`: 87 blocks decompressing to 5,701,632 bytes = 87 x 64 KiB exactly. A block whose *compressed* length is exactly 65,536 is stored verbatim instead (`NMSARC.UI.pak` and the `TexBiomes*` family) — a length rule, not a magic sniff.
 * Entry offsets are into a **virtual image** of the file, so stream position is `entry.offset - dataStart`.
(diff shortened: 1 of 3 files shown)


jonstump (commented):
**Verdict: approve** (posted as a comment — GitHub does not permit self-approval on this account).

Reviewed against SPEC-0003 and the code merged in #13. Every factual claim here was re-verified against the real install rather than taken from the PR description:

| Claim | Verified |
|---|---|
| Flag `1` = block stream, 95 archives | 95 of 97 measured |
| Flag `0` = stored, `NMSARC.audio.pak` + `NMSARC.audioBNK.pak` | both confirmed; `dataStart == 0x30 + entryCount*32` holds for each |
| Raw block when compressed length == 65,536 | 8 archives — `NMSARC.UI.pak` plus 7 `TexBiomes*` |
| Stored manifest is plaintext | confirmed — `audioBNK` entry 0 begins `audio/windows/...` with no zstd frame |

Three things this gets right that are easy to get wrong:

1. **The raw-block rule is stated as a length rule decided _before_ decompression**, and explicitly not a fallback. That distinction is the whole reason it does not contradict the design doc's rejection of PSARC's inflate-and-fall-back, and the spec says so in as many words.
2. **The overflow bound is generalized into a requirement** rather than left as an implementation detail of `80eeu4c1`. "MUST bound each against the bytes actually remaining **before** allocating for it, and MUST NOT perform that check by computing a sum or a signed conversion that can overflow" is the rule that would have prevented the bug, not just a description of the patch.
3. **The ADR records why the original claim was wrong** — "true of the two archives parsed at the time, and false in general" — rather than quietly correcting it. That is the same failure mode as the original PSARC edge label, and naming it is what stops it recurring.

Scenario coverage matches the new normative text: both layouts, verbatim blocks, and oversized extents each get a WHEN/THEN.

No conflict markers. CI green (Formatting, Static analysis, Tests (race), GitGuardian).

PR #15, opened 2026-08-18
HGPAK manifest, path resolution, and selective extraction

### LINKED ISSUE #10: Manifest, path resolution, and selective extraction

jonstump:
Closes #10
Part of #7

Implements SPEC-0003 REQ "Manifest and Path Resolution", REQ "Selective Extraction".

Adds `Paths`, `Lookup`, `Path`, and `ReadPath` to `internal/hgpak`. Entry 0 is a plaintext CRLF manifest, so names come from the archive alone — no external hash-to-name mapping, which is what makes this reader usable against an install nobody has catalogued.

## Selectivity is measured, not asserted

The acceptance criteria ask for a block-decompression counter, so `BlocksRead` is part of the API rather than test-only scaffolding. Measured on `NMSARC.Precache.pak` — 5.8 MB on disk, decompressing to 107.7 MB across 1,644 blocks, 13,824 entries:

| Operation | Blocks touched |
|---|---|
| `Open` | 1 (the `StreamLen` probe) |
| List all 13,823 paths | 17 |
| Read one 2 MB table | 33 |

That is the requirement's whole point: reading one table costs 33 blocks, not 1,644. The listing figure is why `Paths` resolves lazily rather than at `Open` — a caller who only wants `Entry(3)` should not pay for a 13,823-path manifest.

Tests assert **deltas** around a call rather than absolute counts, because `Open` legitimately reads the final block to settle `StreamLen`. An absolute count would either bake in that off-by-one or quietly tolerate any regression that changed it.

## Verification across a real install

`TestFullArchivePathResolutionFromRealInstall` resolved and hash-verified **184,823 paths across all 97 archives**. Because `Paths` verifies every entry's hash as it resolves, a clean pass is full MD5-of-lowercase-path verification — not the 25-entry spot check #9 used, which the acceptance criteria explicitly asked to replace.

## Decisions worth a look

**Resolution is lazy and cached, errors included.** A malformed manifest does not become well-formed on retry, so caching the failure costs nothing and keeps repeat calls cheap. If you'd rather a caller be able to retry after, say, a transient `ReaderAt` failure, that's the line to change.

**Duplicate manifest paths are rejected**, which the spec does not explicitly require. A path silently naming one of two entries is the same class of quiet wrongness as a hash mismatch, and the map would otherwise pick a winner by iteration order. Easy to relax if a real archive turns out to contain duplicates — none of the 97 do.

**`Lookup` lowercases the query and tolerates a leading `/`.** Manifest paths are lowercase by format, so a caller should not need to know that. The CLI in #11 will want it for user-supplied filters.

**Concurrency.** `Paths` mutates `Archive` state on first call while other readers may be reading it, so resolution goes through `sync.Once` and the counter is atomic. `TestConcurrentUseIsSafe` exercises 16 goroutines racing on the lazy resolve, and CI runs `-race`.

## Note on the acceptance criterion's wording

The criterion reads "listing `NMSARC.Precache.pak` resolves all 13,824 entries by path". It resolves **13,823** paths for **13,824** entries — entry 0 is the manifest and has no path of its own, which is the format working as specified rather than an off-by-one. `Paths()[n]` names entry `n+1`.

## Verification

| Check | Result |
|---|---|
| `NMS_PCBANKS=... go test -race` | 97/97 archives, 184,823 paths verified |
| `go vet ./...` | clean |
| `gofmt -l .` | no output |

## Scope

Extraction to disk, path traversal safety, and the CLI are #11 — as is deleting `internal/psarc`, in the PR that proves the replacement works.

Reading `metadata/reality/tables/nms_reality_gcproducttable.mbin` by path already returns 2,056,778 bytes beginning `cccccccc`, so the MBIN magic the epic's acceptance test depends on is confirmed reachable through this API.

Depends on nothing outstanding; #14 (docs) is independent and can merge in either order.

### DIFF
--- a/internal/hgpak/hgpak.go
+++ b/internal/hgpak/hgpak.go
@@ -52,6 +52,8 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+	"sync"
+	"sync/atomic"
 
 	"github.com/klauspost/compress/zstd"
 )
@@ -118,6 +120,21 @@ type Archive struct {
 	streamLen uint64
 
 	dec *zstd.Decoder
+
+	// blocksRead counts blocks materialized, so tests can prove that
+	// listing and single-entry reads touch only the blocks they must
+	// (SPEC-0003 REQ "Selective Extraction"). Atomic because reads are
+	// safe to make concurrently.
+	blocksRead atomic.Uint64
+
+	// The manifest is resolved at most once, on first use rather than at
+	// Open: REQ "Selective Extraction" requires opening an archive not to
+	// materialize its stream, and a 47,000-entry manifest is real work a
+	// caller who only wants Entry(3) should not pay for.
+	pathOnce sync.Once
+	pathErr  error
+	paths    []string
+	byPath   map[string]int
 }
 
 // StreamPos returns the entry's position within the concatenated
@@ -149,6 +166,17 @@ func (a *Archive) BlockCount() int { return len(a.blockLen) }
 
 // StreamLen returns the total decompressed length of the block stream.
 func (a *Archive) StreamLen() uint64 { return a.streamLen }
+
+// BlocksRead returns how many blocks have been materialized over this
+// Archive's lifetime — read from the file and either decompressed or taken
+// verbatim. It backs the SPEC-0003 REQ "Selective Extraction" guarantee that
+// listing costs only the manifest's blocks and reading one entry costs only
+// that entry's.
+//
+// Blocks are not cached, so re-reading an entry counts its blocks again.
+// Open itself reads the final block to settle StreamLen, so measure a
+// delta against the value just after Open rather than an absolute count.
+func (a *Archive) BlocksRead() uint64 { return a.blocksRead.Load() }
 
 // Entries returns a copy of the entry table.
 func (a *Archive) Entries() []Entry {
@@ -334,6 +362,7 @@ func (a *Archive) block(i int) ([]byte, error) {
 	if _, err := a.r.ReadAt(raw, a.blockPos[i]); err != nil {
 		return nil, fmt.Errorf("reading block %d at %d: %w", i, a.blockPos[i], err)
 	}
+	a.blocksRead.Add(1)
 	dec := a.dec
 	if dec == nil {
 		// readTables probes the final block before the shared decoder
--- a/internal/hgpak/manifest.go
+++ b/internal/hgpak/manifest.go
@@ -1,0 +1,151 @@
+package hgpak
+
+// Governing: ADR-0001 (two-tier NMS data ingestion), SPEC-0003 REQ "Manifest
+// and Path Resolution", REQ "Selective Extraction".
+//
+// Entry 0 is a plaintext manifest of CRLF-separated lowercase paths, and
+// manifest path n names archive entry n. Names therefore come from the
+// archive alone — no external hash-to-name mapping, which is what makes this
+// reader usable against an install nobody has catalogued.
+//
+// Each entry's 16-byte hash is the MD5 of its lowercase path. That is
+// redundant with the manifest, which is exactly why it is worth checking:
+// the two structures agreeing is a free consistency check, and a cheap early
+// warning that a game update changed the convention. The hash is never used
+// as the lookup key — doing so would reintroduce the dependency the manifest
+// removes.
+
+import (
+	"bytes"
+	"crypto/md5"
+	"fmt"
+	"strings"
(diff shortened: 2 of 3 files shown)


jonstump (commented):
**Verdict: approve** (posted as a comment — GitHub does not permit self-approval on this account).

Reviewed against issue #10's acceptance criteria and SPEC-0003 REQ "Manifest and Path Resolution" / REQ "Selective Extraction". I ran the criteria against the real `NMSARC.Precache.pak` rather than relying on the suite's own assertions:

| Acceptance criterion | Measured |
|---|---|
| Listing Precache resolves every entry by path | 13,823 paths for 13,824 entries — manifest + files, exactly `entryCount - 1` |
| Listing touches only manifest blocks | 18 blocks of 1,644 (`Open` itself costs 1, for the final-block length probe) |
| Single-entry read touches only that entry's blocks | product table: 2,056,778 bytes for 33 blocks of 1,644 |
| Open does not materialize the stream | 1 block read at `Open` |
| MD5-of-lowercase-path holds across every archive | full 97-archive opt-in run passes |

Independently: all 54 `metadata/reality/tables/` paths resolve, and `ReadPath("metadata/reality/tables/nms_reality_gcproducttable.mbin")` returns bytes beginning `cccccccc` — MBIN magic, which is the thing the whole ADR-0001 pipeline is after.

`BlocksRead()` is the right call. The requirement "MUST NOT decompress the whole archive to read one file" is otherwise unfalsifiable from outside — a correct implementation and a wasteful one return identical bytes. Exposing the counter turns a performance claim into an assertable one, and `TestListingCostsOnlyManifestBlocks` / `TestSingleEntryReadCostsOnlyItsBlocks` are the tests that could not exist without it.

`gofmt` clean, `go vet` clean, `-race` clean, no conflict markers, CI green.

Depends on #9 (merged). Nothing here blocks #11.

PR #17, opened 2026-08-19
Safe extraction, CLI rewiring, and removal of internal/psarc

### LINKED ISSUE #11: Safe extraction and CLI pipeline acceptance

jonstump:
Closes #11
Part of #7

Implements SPEC-0003 REQ "Safe Extraction to Disk", REQ "Pipeline Fitness".
Supersedes `internal/psarc` per the SPEC-0003 migration plan.

Adds `ExtractTo`, rewires `cmd/nmsextract` onto `internal/hgpak`, and deletes `internal/psarc` — in the PR that proves the replacement works, as the migration plan requires.

## The acceptance test passes end to end

Run against a real install (NMS 5.97) with MBINCompiler 6.45.0.1:

```
$ nmsextract list    "$PCBANKS/NMSARC.Precache.pak" reality/tables
54 entries, 14498796 bytes uncompressed
(filtered by "reality/tables")

$ nmsextract extract "$PCBANKS/NMSARC.Precache.pak" ./out metadata/reality/tables/
extracted 54 entries, 14498796 bytes, to ./out
```

All 54 files begin `cccccccc`. MBINCompiler decompiled `nms_reality_gcproducttable.mbin` — 2,056,778 bytes in, 9,799,762 bytes of `.MXML` out, exit 0, no errors. The result parses as XML and contains `ULTRAPROD2`, the Stasis Device root that ADR-0001's confirmation section depends on.

That closes the loop the epic opened: game archive → MBIN on disk → MBINCompiler → XML the normalizer can read.

Note the extension is `.MXML`, per the correction in #16. Issue #11's body still says EXML; the spec is right and the issue text is stale.

## Two symlink escapes found and fixed

The extraction layer landed with lexical traversal checks plus a symlink check on the destination's parent chain. Two escapes got through both, and each is now a regression test that fails without its fix:

**A symlink at the destination file was followed.** `confinedToRoot` checks `filepath.Dir(dest)`, so a pre-existing `out/x.mbin -> /outside/victim` resolved its parent chain cleanly and `os.WriteFile` followed the final link. `ExtractTo` returned `Files: 1, err: nil` while overwriting a file outside the tree — a silent success, which is the worst shape for this failure. `Lstat` before writing sees the link rather than its target.

**`MkdirAll` ran before the containment check.** A path through a symlinked parent (`link/sub/deep/x.mbin`) built its directory chain *inside the symlink target*, and only then did the check refuse the write. The file never landed outside, but the directories did. REQ "Safe Extraction to Disk" says nothing is written outside the directory, and directories count. Containment is now checked against the deepest existing ancestor before anything is created.

The post-`MkdirAll` check is kept as well — it costs nothing and still covers a link planted between the two.

Both are same-class bugs: the check was correct about the path it examined and examined the wrong path.

**Scope note.** These close the archive-controlled escapes the spec names. A concurrent attacker who plants a symlink between the `Lstat` and the `WriteFile` is still theoretically unhandled — that is a TOCTOU race requiring local write access to the output directory mid-extraction, which is outside the threat model for a developer-local tool whose untrusted input is the archive. Worth naming rather than leaving implied.

## A #10 bug fixed here too

The manifest index was keyed by the **raw** path while `Lookup` normalized its query, so any manifest path not already lowercase and slash-free was unreachable by its own name — contradicting REQ "Manifest and Path Resolution" ("MUST expose every entry by its path").

No shipping archive triggers it: all 184,823 paths across the install are already normalized, which is exactly why the asymmetry survived #10's review and its full-install run. `TestEveryManifestPathIsReachableByItsOwnName` uses a crafted archive to catch it.

## CLI

`list` and `extract` both take an optional case-insensitive substring filter, and `list` reports the match count — required because a single archive holds tens of thousands of entries.

Errors now go through `slog` with the sentinel and any `StructureError` fields attached as key-value pairs, per REQ "Error Handling Standards". That requirement asked for structured logging and #9 only had the error vocabulary; this is the call site that uses it.

## Verification

| Check | Result |
|---|---|
| `NMS_PCBANKS=... go test` | 97/97 archives, full extraction suite |
| Real CLI acceptance | 54 tables, all `cccccccc`, MBINCompiler exit 0 |
| `go vet ./...` | clean |
| `gofmt -l .` | no output |

## `internal/psarc` is gone

Package and both test files deleted; nothing references it. The only remaining `psarc` mentions in Go are in `hgpak_test.go`, where a crafted PSARC header verifies that such a file is rejected with `ErrNotHGPAK` naming the magic found — the superseded format's one remaining job.

## Still open

`printableMagic` returns `""` when the magic starts with a non-printable byte, so REQ "Container Identification"'s "name the magic found" is unmet for non-ASCII-leading files. Carried over from the #13 review and deliberately not folded in here — this PR is already load-bearing, and that fix belongs with a test of its own.


jonstump (commented):
**Verdict: one finding, otherwise approve** (comment rather than approval — GitHub does not permit self-approval on this account).

## Your two symlink fixes are correct, and both were real

I verified rather than took the commit message's word for it: reverting `extract.go` to `ff44810` makes `TestSymlinkAtDestinationFileIsRefused` and `TestNoDirectoriesAreCreatedOutsideTheOutputTree` both fail; with `6199bdf` they pass.

Both were bugs in code I wrote, and the second is the more interesting one. My `TestSymlinkEscapeIsRefused` **passed** the whole time — it asserted the file wasn't created through the link, and stopped there. It never checked whether directories had been created outside, and it never considered a symlink at the destination itself. A green test on the exact hazard it was written for, with two holes on either side of it. Same shape as the PSARC suite that validated its own writer against its own reader.

`deepestExisting` is the right primitive: `EvalSymlinks` needs an extant path, and checking the deepest existing ancestor *before* `MkdirAll` is what makes "nothing is written outside the directory" true for directories and not just files. Keeping the post-`MkdirAll` check as well is worth its zero cost.

## Finding: a hardlink at the destination still escapes

`Lstat` sees a symlink; a hardlink reports as a regular file, and `os.WriteFile` writes straight through it to the shared inode. Same precondition as the symlink-at-destination case this commit just closed — the attacker pre-creates the link inside the output directory — so leaving it open is inconsistent with the fix.

Reproduced against this PR's head:

```
victim OUTSIDE the tree was modified through a hardlink (err=<nil>)
victim now = "PAYLOAD"
```

Note the `err=<nil>`: `ExtractTo` reports **success** while clobbering a file outside the output tree. That is precisely the failure mode the commit message objects to in the symlink case.

Suggested fix — unlink before writing, which handles symlink, hardlink, and stale-file cases in one line and makes re-extraction idempotent:

```go
// Unlink first: os.WriteFile follows a symlink and writes through a
// hardlink to the shared inode. Removing the link itself leaves whatever
// it pointed at alone.
if err := os.Remove(dest); err != nil && !errors.Is(err, fs.ErrNotExist) {
    return res, fmt.Errorf("clearing destination for %q: %w", p, err)
}
```

That subsumes the `Lstat` check, though keeping both is defensible.

**Severity is low in context** — this is a developer-local tool writing into a directory the operator chooses, per ADR-0001, and there is no untrusted-input path. I would not block on it if the alternative were shipping nothing. But it is an unclosed escape in the one function whose entire job is preventing escapes, and it is a one-line fix, so it seems worth closing before merge rather than after.

## Everything else verified

| Check | Result |
|---|---|
| Acceptance test (real install) | 54 tables extracted, all `cccccccc`; MBINCompiler 6.45.0.1 converts all 54 |
| Full gated suite with `NMS_PCBANKS` | pass, `-race` |
| `gofmt` / `go vet` / staticcheck | clean |
| CI on this PR | 4/4 pass |
| Conflict markers | none |
| `internal/psarc` deleted | 3 files, 847 lines, nothing references it |

The manifest-index normalization fix carries its own regression test and is spec-justified under REQ "Manifest and Path Resolution" ("MUST expose every entry by its path"); blast radius on real data is 0 of 184,823 paths.

jonstump (commented):
Hardlink escape closed in 75efbb4, verified failing without the fix. Kept refusal semantics rather than the unlink-before-write I first suggested — that regressed `TestSymlinkAtDestinationFileIsRefused`, and your test catching it is why the loud behaviour survived. Link counting is build-tagged since Nlink has no portable form. All four checks green.

PR #18, opened 2026-08-19
docs: confirm the ADR-0001 Tier 2 finding — four of five constants are extractable

jonstump:
Documents only. Closes the last open question in [ADR-0001](docs/adrs/ADR-0001-two-tier-nms-data-ingestion.md), and marks it `accepted`.

The ADR gated its own acceptance on confirming its Tier 2 finding against decompiled MBIN files rather than libMBIN struct names. That confirmation is now done — **and the finding did not survive it.**

## Four of the five constants are extractable

| Constant | Where it actually lives |
|---|---|
| Generator / extractor / consumer rates | `GcBaseLinkGridData.Rate` and `.Storage` — 101 base parts carry a nonzero value |
| C/B/A/S class scaling | `METADATA/SIMULATION/SCANNING/REGIONHOTSPOTSTABLE.MBIN` — `ClassStrengths`, `ClassWeightings` |
| Crop yield | reward table — `GcRewardSpecificSubstance` with `AmountMin` / `AmountMax` |
| Crop growth time | `Storage` on the plant's `PlantGrowth` connection, in seconds |
| Biodome crop-slot count | **Not found** — the only remaining Tier 2 constant |

**Class scaling is ranges attached to hotspots, not multipliers attached to devices.** A part declares a base `Rate` plus a `DependsOnHotspots` category; the hotspot carries the class. Power hotspots: `ClassStrengths` 150 / 220 / 250 / 300 for C/B/A/S. Mineral and Gas: 1 / 1.5 / 2 / 2.5. Plus `ClassWeightings` biasing which class spawns (Power 10/6/2/1, Gas 20/4/2/1). Searching the parts table for `_C`/`_B`/`_A` variants finds nothing *because the class is never a property of the device* — which is exactly how the first pass talked itself into "classes aren't in the data."

The link-grid model computes a base's power budget directly:

```
U_EXTRACTOR_S   Rate=100  Storage=360000  DependsOnHotspots=Mineral
                └─ DependentConnections: Power, DependentRate=-50, DependentEffect=EnablesRate
U_SOLAR_S       Rate=50                   U_BATTERY_S  Storage=45000
BIOROOM         Rate=-50                  PLANTER Rate=-5   PLANTERMEGA Rate=-20
```

Plants declare `DependentRate = 0` against Power, so a biodome's −50 is the entire power cost of its contents — there is no per-plant draw to accumulate.

Crop yields are flat per crop: Cactus Flesh 100; Gamma Root, Solanium, Frost Crystal, Fungal Mould 50; Star Bulb, Mordite, Faecium 25. Growth times come from the same `Storage` field — 3,600 s for frostwart, 14,400 s for most, 57,600 s for the slowest.

## What this does to the tiers

**Tier 1 gains** per-part rates and storage, class strengths and weightings, crop yields, crop growth times, battery capacity, and refiner throughput (`RefinerProductsMadeInTime` / `RefinerSubsMadeInTime`, difficulty-dependent).

**Tier 2 shrinks to one entry** — biodome crop-slot count.

The decision is unaffected. Option C was chosen for licensing and refresh-cycle reasons independent of how the tiers split, and this ADR anticipated the outcome in its own words: *"if the constants turn out to be extractable, Tier 2 shrinks or disappears and this decision should be revised rather than superseded."* What changed is the size of Tier 2, not whether the two-tier model is right — so `status: accepted` stands.

## How the first pass got it wrong

This PR originally asserted the opposite, with a coverage list that read as thorough. The commit history preserves both, and the ADR records the failure, because it is the third instance of the same pattern this document already tracks.

The first pass searched `METADATA/REALITY/TABLES/` and the globals, read `DependentRate` inside connection dependencies, and **never read the sibling `Rate` and `Storage` on the same structure** — despite `Rate` appearing 2,167 times in its own grep output. It never opened `METADATA/SIMULATION/SCANNING/`, despite `gcterrainglobals` naming `RegionHotspotsTable` outright. And the reward table holding every crop yield had already been decompiled and grepped.

No missing capability — a bounded search reported as a general result. Same move that produced the `PSARC` edge label and "an unknown field (1 in every archive observed)". The lesson the ADR now states: **state the boundary of a search whenever you state its result.**

## Still open

Biodome crop-slot count is not in any table searched. Reported as 16 per the community wiki; likely geometric (snap points in the scene) rather than table-driven. It stays Tier 2, curated with a `source` and a `verified` date — which is precisely what Tier 2 is for.

### DIFF
--- a/docs/adrs/ADR-0001-two-tier-nms-data-ingestion.md
+++ b/docs/adrs/ADR-0001-two-tier-nms-data-ingestion.md
@@ -33,9 +33,9 @@ Where does that data come from, in what form does it reach the app, and under wh
 
 Chosen option: **C, direct extraction with a two-tier data model**, because it is the only option that gets the recipe graph under a permissive footing while keeping the refresh cycle under our control — and because the constants that extraction cannot supply turn out to be a small curated set rather than a second pipeline.
 
-**Tier 1 — extracted.** A Go CLI locates the game's `.pak` archives, unpacks them (they are HGPAK containers — see the note below and SPEC-0003), shells out to MBINCompiler to convert `.MBIN` → `.MXML`, then parses and normalizes the XML in Go into a version-stamped artifact: the recipe graph (products, substances, refinery, nutrient processor), item metadata, the biome→gas mapping from `GcGeneratorUnitComponentData.BiomeGasRewards`, the extractor taxonomy, the base parts catalog, and refiner throughput (`RefinerProductsMadeInTime` / `RefinerSubsMadeInTime` and their Survival variants, from `gcgameplayglobals`). Regenerated per game version; never hand-edited.
+**Tier 1 — extracted.** A Go CLI locates the game's `.pak` archives, unpacks them (they are HGPAK containers — see the note below and SPEC-0003), shells out to MBINCompiler to convert `.MBIN` → `.MXML`, then parses and normalizes the XML in Go into a version-stamped artifact: the recipe graph (products, substances, refinery, nutrient processor), item metadata, the biome→gas mapping from `GcGeneratorUnitComponentData.BiomeGasRewards`, the extractor taxonomy, the base parts catalog, refiner throughput (`RefinerProductsMadeInTime` / `RefinerSubsMadeInTime` and their Survival variants, from `gcgameplayglobals`), the per-part production and consumption rates and storage buffers from `GcBaseLinkGridData`, the C/B/A/S hotspot class strengths and weightings from `REGIONHOTSPOTSTABLE`, and crop yields and growth times. Regenerated per game version; never hand-edited.
 
-**Tier 2 — curated.** A hand-maintained YAML file of the base-economy constants that are *not* in the game files (generator outputs, power draws, extractor rates per class, biodome capacity, crop yields — confirmed absent, see the finding below), each entry carrying a `source` and a `verified` date, feeding the same `unverified` badge convention the tree canvas uses.
+**Tier 2 — curated.** A hand-maintained YAML file of the base-economy constants that are genuinely *not* in the game files. As of the 2026-08-18 confirmation that is one entry — biodome crop-slot count — since generator and consumer rates, class scaling, crop yields, and growth times all turned out to be extractable (see the finding below). Each entry carries a `source` and a `verified` date, feeding the same `unverified` badge convention the tree canvas uses.
 
 MBINCompiler is invoked as a **subprocess**, never linked. It is LGPL-3.0, and a license governs a program rather than its output, so the extracted artifact carries no copyleft obligation. AssistantNMS is retained as a **cross-check** on Tier 1 correctness — read, compared against, never vendored.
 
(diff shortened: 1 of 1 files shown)


jonstump:
## Finding correction — `c914b8e` → `5f24e6b`

Tracking this here as well as in the ADR, since the PR asserted the opposite for a while and the history should be readable from either side.

**What this PR said first:** all five base-economy constants are absent from the game files; Tier 2 shrinks but survives; the search covered "the right places with real values."

**What is actually true:** four of the five are extractable. Tier 2 shrinks to a single entry.

| Constant | First pass | Actual |
|---|---|---|
| Generator / extractor rates | absent | `GcBaseLinkGridData.Rate` / `.Storage`, 101 parts |
| C/B/A/S class scaling | absent — "classes aren't separate parts" | `REGIONHOTSPOTSTABLE.MBIN` — `ClassStrengths` 150/220/250/300 (Power), 1/1.5/2/2.5 (Mineral, Gas) |
| Crop yield | absent | reward table — `AmountMin`/`AmountMax`; Frost Crystal 50 |
| Crop growth time | not looked for | `Storage` on the `PlantGrowth` connection, seconds |
| Biodome crop-slot count | absent | still absent — the one real Tier 2 entry |

### Three misses, all in the same shape

1. **Read `DependentRate`, never read `Rate`.** Both sit on `GcBaseLinkGridData`. The first pass found the connection-dependency field, reported "only 24 of 1,997 entries carry a rate," and stopped — while `Rate` appeared **2,167 times in its own grep output**, was noted, and was dropped.
2. **Never opened `METADATA/SIMULATION/SCANNING/`.** The search covered `METADATA/REALITY/TABLES/` and the globals and called that comprehensive. `gcterrainglobals` names `RegionHotspotsTable` by path; that pointer was read past.
3. **Had the crop yields already.** `rewardtable.mbin` was among the 54 decompiled tables, was grepped, and returned `SpecialHarvestID`/`SpecialHarvestMul` — which looked irrelevant, so `GcRewardSpecificSubstance` two levels down was never reached.

### On the reframing that unblocked it

The correction came from @jonstump pointing out that classes are *ranges each thing falls into*, not multipliers. That is a different search — range-typed fields on hotspots rather than scalar multipliers on devices — and it is what led to `REGIONHOTSPOTSTABLE`. The "yield might be on the item, not the plant" hypothesis turned out **not** to hold (the `PLANT_SNOW` substance record carries no yield field, verified by full nested dump), but chasing it is what surfaced the reward-table path.

Worth recording that the wrong hypothesis was still the useful one.

### The rule this earns

The failure was not capability. Every one of these was reachable with tools already in hand and, in two cases, with output already on screen. It was **reporting a bounded search as a general result** — the same move as the `PSARC` diagram edge and "an unknown field (1 in every archive observed)."

The ADR now states it: *state the boundary of a search whenever you state its result.*

### Also confirmed by the data

Plants declare `DependentRate = 0` against Power. A biodome's −50 is the entire power cost of its contents; there is no per-plant draw. That matches @jonstump's read and simplifies the POWER block — dome count drives the number, not plant count.

Issue #24, opened 2026-08-19
Base economy data


jonstump:
Part of #21. Governed by [SPEC-0004](docs/openspec/specs/tier1-normalizer/spec.md).

**Blocked by #22** — needs the extended schema to emit into.

Emits the base-economy values confirmed extractable on 2026-08-18, so they are version-stamped and regenerated rather than hand-curated. This is the work that shrank Tier 2 from five constants to one.

Sources span three locations, which is the main reason this is its own story: rates come from the reality tables, class scaling from `METADATA/SIMULATION/SCANNING/`, and refiner throughput from the globals archive.

## Requirements

- [ ] **Base Economy Data** (SPEC-0004) — The artifact MUST carry per-part production and consumption rates and storage buffers from `GcBaseLinkGridData` with the network each applies to and any dependent-connection rate; hotspot class strengths and weightings for C/B/A/S from `REGIONHOTSPOTSTABLE`; crop yields from the reward table's `GcRewardSpecificSubstance` minimum and maximum amounts; crop growth times from the plant's `PlantGrowth` connection storage value; and refiner throughput from `gcgameplayglobals` including difficulty variants. Class scaling MUST be modelled as a property of the hotspot, not of the device — the normalizer MUST NOT emit per-class device variants, which do not exist in the source. Values the source expresses as a minimum and maximum MUST be preserved as a range.
  - WHEN `U_EXTRACTOR_S` is emitted THEN its rate, its storage, and its dependent power draw are present, each identified by the network it applies to
  - WHEN class strengths are emitted THEN they are keyed by hotspot category and class, and no per-class device variants appear
  - WHEN a crop yield is expressed as a minimum and maximum in the source THEN both bounds are emitted, rather than one derived value

- [ ] **Search Boundaries Are Recorded** (SPEC-0004) — Where the normalizer derives a value by searching for it rather than reading a known field, the artifact or its generation log MUST record which sources were searched.
  - WHEN a value is produced by searching several tables rather than reading one known field THEN the sources searched are recorded alongside it

## Acceptance Criteria

- `U_EXTRACTOR_S` emits rate 100, storage 360000, and a dependent Power draw of −50 tagged `EnablesRate`
- Hotspot class strengths emit as Power 150/220/250/300 and Mineral/Gas 1/1.5/2/2.5, keyed by category and class
- Crop yields emit as ranges — Frost Crystal 50/50, Cactus Flesh 100/100, Star Bulb 25/25
[...]

Issue #23, opened 2026-08-19
Recipe graph and identity


jonstump:
Part of #21. Governed by [SPEC-0004](docs/openspec/specs/tier1-normalizer/spec.md).

**Blocked by #29** — needs the recipe-list schema and yield field. (#22, which provided the artifact envelope and sentinels, is merged.)

**Scope note (ADR-0005):** SPEC-0004 REQ "Recipe Graph Construction" was amended after this story was written. The normalizer must now emit *every* recipe per output and method rather than one, read each recipe's yield from the source, distinguish cook from refine by the per-recipe `Cooking` flag, and exclude self-referential recipes while recording the count in provenance.

Builds the `items` and `recipes` half of the artifact from the reality tables, resolves identities, and joins the localisation tables for display names. This is the half that `testdata/stasis-device.tier1.json` can be diffed against.

The localisation join is the part most likely to be underestimated: the reality tables carry only keys (`UI_ULTRAPROD_2_NAME_L`), and those resolve in `language/nms_loc*_english.mbin` inside a **different archive** than the product table. Searching `GcProductTable` for "Stasis Device" returns nothing.

## Requirements

- [ ] **Recipe Graph Construction** (SPEC-0004) — The normalizer MUST build `items` and `recipes` from the reality tables — products, substances, and the refinery and cooking recipe tables. Every recipe's `method` MUST be one of `craft`, `refine`, `raw`, `cook`; the normalizer MUST NOT emit a `buy` method. Every `Input.item` and `Recipe.output` MUST reference an `Item.id` present in the same artifact. Items with no recipe MUST be emitted with `raw_obtainable` true and `default_method` `raw`.
  - WHEN an artifact is generated THEN every recipe input and output resolves to an item in the same artifact
  - WHEN a recipe references an item ID absent from the tables THEN generation fails naming the recipe and the missing ID, and no artifact is written
  - WHEN a source table implies a method outside craft, refine, raw, cook THEN generation fails naming the method found, rather than emitting it

- [ ] **Identifier Policy** (SPEC-0004) — Item identifiers MUST be the game's own IDs (`ULTRAPROD2`, `PLANT_SNOW`), not invented short codes. The normalizer MUST NOT normalize, case-fold, or otherwise rewrite game IDs. The existing fixtures' short codes (`fc`, `sb`) are illustrative of shape, not of identifier policy.
[...]

Issue #21, opened 2026-08-19
Implement Tier 1 Normalizer


jonstump:
The missing middle of the [ADR-0001](docs/adrs/ADR-0001-two-tier-nms-data-ingestion.md) pipeline: the stage that turns decompiled game tables into the artifact the app loads.

## Spec

- [spec.md](docs/openspec/specs/tier1-normalizer/spec.md) — SPEC-0004, 12 requirements
- [design.md](docs/openspec/specs/tier1-normalizer/design.md)

## Governing ADRs

- **ADR-0001** — two-tier NMS data ingestion. This epic builds the Tier 1 producer.

## Why now

Two of three stages exist. SPEC-0003 unpacks HGPAK archives; MBINCompiler 6.45.0.1 decompiles to `.MXML`; SPEC-0001's rollup engine loads a `Tier1` artifact and resolves the 34-node Stasis Device tree **today**, from hand-authored fixtures carrying `extracted: false`. Nothing produces a real artifact.

Scope grew on 2026-08-18: four of the five constants ADR-0001 planned to hand-curate turned out extractable, so the normalizer emits base-economy data as well as the recipe graph, and `Tier1` grows to hold it.

## Stories

| # | Title | Branch | Depends on |
|---|-------|--------|-----------|
| #22 | Artifact schema, provenance, fail-closed parsing `foundation` | `feat/22-artifact-schema-provenance-fail-closed` | — (merged) |
| #29 | Recipe lists, yields, and deterministic selection `foundation` | `feat/29-recipe-lists-yields-and-selection` | #22 |
| #23 | Recipe graph and identity | `feat/23-recipe-graph-and-identity` | #29 |
| #24 | Base economy data | `feat/24-base-economy-data` | #22 |
| #25 | Acceptance against generated output | `feat/25-acceptance-against-generated-output` | #23 |

#29 was split out of #23 after [ADR-0005](docs/adrs/ADR-0005-multiple-recipes-per-output.md): the game defines many recipes per output, recipes carry a yield the schema had no field for, and `Tier1.Validate` currently rejects exactly the data the normalizer needs to emit. That is a change to merged, tested code plus new traversal work for the selection rule — a different blast radius from parsing tables, so it is its own story.

#24 is unblocked and independent of the whole recipe question — different source tables, different artifact section.

[...]

PR #26, opened 2026-08-19
Artifact schema, provenance envelope, and fail-closed errors

### LINKED ISSUE #22: Artifact schema, provenance, and fail-closed parsing

The foundation story: extend the artifact schema to hold what the normalizer will emit, establish the provenance envelope and sentinel vocabulary, and set the fail-closed parsing posture everything else writes through.

**Extends `internal/domain/tier1.go`** (currently defines `Tier1`, `Item`, `Recipe`, `Method` and `LoadTier1`) rather than starting fresh. `LoadTier1` decodes with `DisallowUnknownFields`, so the schema, the loader, and the committed fixtures MUST move together — a normalizer emitting sections the struct does not declare produces artifacts that fail to load.

## Requirements

- [ ] **Schema Extension and Load Compatibility** (SPEC-0004) — The `Tier1` schema MUST be extended to hold the base-economy data, and `SchemaVersion` MUST be incremented in the same change. The producer, the schema, and the loader MUST change together; the normalizer MUST NOT emit a section that `internal/domain` cannot decode. Existing fixtures MUST continue to load, migrated in the same change where the extension requires it.
  - WHEN an artifact carrying base-economy sections is passed to `LoadTier1` THEN it decodes without an unknown-field error and validates
  - WHEN the committed fixtures are loaded after the schema extension THEN they decode and validate, having been migrated if the extension required it

- [ ] **Source Provenance and Version Stamping** (SPEC-0004) — The artifact MUST set `extracted` to `true`, MUST record the game build in `game_version`, and MUST record the source archives and MBINCompiler version in `source`. The normalizer MUST NOT emit a `game_version` it did not read from the install; where it cannot be determined the read MUST fail rather than emit a guess or placeholder.
  - WHEN an artifact is generated from a real install THEN `extracted` is `true`, `game_version` names the build read, and `source` names the archives and the MBINCompiler version used
  - WHEN the game version cannot be determined from the install THEN generation fails naming what could not be read, and no artifact is written

- [ ] **Deterministic Output** (SPEC-0004) — Running the normalizer twice against the same install MUST produce byte-identical output. Collections MUST be emitted in a defined, stable order — not map-iteration order, and not source-table order.
  - WHEN the normalizer runs twice against an unchanged install THEN the two artifacts are byte-identical and `git diff` reports no change
  - WHEN the game updates and one recipe's quantity changes THEN the artifact diff shows that change and no unrelated reordering

- [ ] **Structural Surprise Fails Loudly** (SPEC-0004) — Where a source table's structure does not match expectation — a missing field, an unrecognized enum value, a table absent from the archive — generation MUST fail naming the table and the expectation violated. The normalizer MUST NOT emit a partial artifact, silently skip an unparseable row, or substitute a default for a value it could not read.
  - WHEN an expected field is absent from a source table THEN generation fails naming the table and the field, and no artifact is written
  - WHEN generation fails partway through THEN no artifact file is left behind

- [ ] **Error Handling Standards** (SPEC-0004) — Errors MUST be wrapped with contextual information at each layer boundary. Sentinel errors MUST be defined for, at minimum: source table missing, structure unrecognized, reference unresolved, and localisation key unresolved. Silent error swallowing MUST NOT occur. Structured logging MUST be used for error reporting (key-value pairs, not string interpolation).
  - WHEN normalization fails on one row of one table THEN the error names the table, the row's identifier, and the expectation violated
  - WHEN a caller encounters a missing source table versus an unresolved localisation key THEN the two failures carry different sentinels

## Acceptance Criteria

- `LoadTier1` accepts an artifact carrying the new base-economy sections
- Both committed fixtures still load and validate after the schema change
- `SchemaVersion` is incremented
- Two runs over an unchanged install produce byte-identical output
- All four sentinels are defined and distinguishable via `errors.Is`
- A deliberately corrupted source table produces a named error and leaves no artifact behind

### Branch

`feat/22-artifact-schema-provenance-fail-closed`

### PR Convention

```
Closes #22
Part of #21

Implements SPEC-0004 REQ "Schema Extension and Load Compatibility", REQ
"Source Provenance and Version Stamping", REQ "Deterministic Output", REQ
"Structural Surprise Fails Loudly", REQ "Error Handling Standards".
```

jonstump:
Closes #22
Part of #21

Implements SPEC-0004 REQ "Schema Extension and Load Compatibility", REQ "Source Provenance and Version Stamping", REQ "Deterministic Output", REQ "Structural Surprise Fails Loudly", REQ "Error Handling Standards".

The foundation for the Tier 1 normalizer: the schema it emits into, the provenance envelope, the sentinel vocabulary, and a deterministic atomic writer. #23 and #24 fill it; this PR is the shape they fill.

## The schema and its fixtures move together

`LoadTier1` decodes with `DisallowUnknownFields`, so a producer emitting sections the struct does not declare yields artifacts that fail to load. `SchemaVersion` goes **1 → 2**, `Validate` now refuses an artifact from the other side of that change rather than half-reading it, and both committed fixtures plus two inline test artifacts are migrated in the same commit.

Verified: `hexaberry-cake.tier1.json` and `stasis-device.tier1.json` both load and validate at schema 2, with 7/4 and 35/25 items/recipes respectively.

## Economy section mirrors the source, not a tidier abstraction

`internal/domain/economy.go` carries what the 2026-08-18 finding correction made extractable:

- **Flows** keyed by network (`power`, `resources`, `plant_growth`, `byte_beat`) with rate and storage — because a part participates in several networks and a single "rate" field would lose which one
- **`ClassValues`** hanging off **hotspot categories**, not devices — class is a property of the hotspot, which is why searching parts for `_C`/`_B`/`_A` finds nothing
- **`Range`** for crop yields, keeping both bounds even when equal, so a later widening in the game data is a visible change rather than a silent one
- **`Refining`** carrying both standard and Survival throughput — settling the design's open question by letting the planner choose rather than assuming Normal

Cross-references are validated: a crop yielding an unknown item, or a part naming an absent hotspot category, is refused.

## Determinism belongs to the emitter, not the callers

`sortArtifact` imposes a total order at assembly, so a caller that ranged over a map still produces stable output. The test builds the same artifact from **opposite insertion orders** and asserts the encodings are byte-identical — which is what makes it a test of the emitter rather than of the caller.

The artifact deliberately carries **no generation timestamp**. One would defeat byte-identical regeneration for no benefit `game_version` does not already provide, and a test asserts its absence so nobody adds one later thinking it is free.

## Fail-closed, in two places

"No artifact file is left behind" has two failure points, so there are two tests:

- **Generation fails** → never reaches the writer. A recipe with a dangling input fails assembly; no file appears.
- **Write fails** → previous artifact intact. `WriteFile` is temp-file-and-rename, so a mid-stream failure leaves the prior artifact rather than a truncated one, and a reader never observes a half-written file. Exercised by making the directory read-only.

Both tests **pass** rather than skip on this machine — worth noting since the second one skips if run as root.

## Sentinels mirror hgpak deliberately

Four sentinels — source missing, structure unrecognized, reference unresolved, localisation key unresolved — plus a `SourceError` carrying `Table`/`Row`/`Field`/`Want`/`Got` and a `LogAttrs()` helper.

The shape intentionally mirrors `hgpak.StructureError`, which I found via the qmd code pre-search rather than by memory. The two packages sit in one pipeline; a caller who has learned to read one error vocabulary should not have to learn a second.

Localisation gets its own sentinel rather than folding into "reference unresolved" because the remedy differs: it usually means the wrong language tables were read, not that the data is inconsistent.

## Verification

| Check | Result |
|---|---|
| `go test ./...` | all packages ok |
| `gofmt -l .` | no output |
| `go vet ./...` | clean |
| `staticcheck` (v0.8.0-rc.1) | exit 0 |
| Fixtures load at schema 2 | both, verified explicitly |

## Note on scope

This story owns the envelope, not the parsing. There is no table reader here — `Builder` accumulates items, recipes and economy from whatever fills it, and #23/#24 are what fill it. The sentinels and `SourceError` exist now so those stories inherit a vocabulary rather than inventing two.

### DIFF
--- a/internal/domain/economy.go
+++ b/internal/domain/economy.go
@@ -1,0 +1,186 @@
+package domain
+
+import "fmt"
+
+// Base-economy data carried alongside the recipe graph.
+//
+// Governing: ADR-0001 (two-tier NMS data ingestion), SPEC-0004 REQ "Schema
+// Extension and Load Compatibility"
+//
+// ADR-0001 originally planned these as hand-curated Tier 2 constants. The
+// 2026-08-18 confirmation found four of the five extractable, so they belong
+// to the machine-generated tier and live here. Only biodome crop-slot count
+// remains curated.
+//
+// The shapes below mirror the source data rather than a convenient
+// abstraction of it: rates are per-network because a part participates in
+// several, class scaling hangs off hotspots because that is where the game
+// puts it, and yields keep both bounds because collapsing a range to a point
+// discards the best/worst case the planner exists to show.
+
+// Network is a link-grid network a rate applies to.
+type Network string
+
+const (
+	NetworkPower       Network = "power"
+	NetworkResources   Network = "resources"
+	NetworkPlantGrowth Network = "plant_growth"
+	NetworkByteBeat    Network = "byte_beat"
+)
+
+var validNetworks = map[Network]bool{
+	NetworkPower: true, NetworkResources: true,
+	NetworkPlantGrowth: true, NetworkByteBeat: true,
+}
+
+// Valid reports whether n is a known network.
+func (n Network) Valid() bool { return validNetworks[n] }
+
+// Flow is a rate on one network, with the buffer that accumulates it.
+// A positive Rate produces, a negative Rate consumes.
+type Flow struct {
+	Network Network `json:"network"`
+	Rate    int64   `json:"rate"`
+	Storage int64   `json:"storage,omitempty"`
+}
+
+// Dependency is a part's requirement on another network — typically a power
+// draw that gates the part's primary rate.
+type Dependency struct {
+	Network Network `json:"network"`
+	Rate    int64   `json:"rate"`
+	// Effect is the source's DependentEffect, e.g. "EnablesRate". Carried
+	// verbatim rather than interpreted, because the planner needs to know
+	// whether an unpowered part produces nothing or merely produces less.
+	Effect string `json:"effect,omitempty"`
+}
+
+// Part is one buildable's economic behaviour.
+type Part struct {
+	ID           string       `json:"id"`
+	Name         string       `json:"name,omitempty"`
+	Primary      Flow         `json:"primary"`
+	Dependencies []Dependency `json:"dependencies,omitempty"`
+	// Hotspot names the hotspot category this part's rate scales with, if
+	// any. The class lives on the hotspot, never on the part.
+	Hotspot string `json:"hotspot,omitempty"`
+}
+
+// ClassValues holds one value per C/B/A/S class.
+type ClassValues struct {
+	C float64 `json:"c"`
+	B float64 `json:"b"`
+	A float64 `json:"a"`
+	S float64 `json:"s"`
+}
+
+// Hotspot carries the class scaling for one hotspot category.
+//
+// Classes are ranges a hotspot falls into, not multipliers attached to a
+// device. Searching the parts table for per-class variants finds nothing
+// precisely because the game models it this way.
+type Hotspot struct {
+	Category   string      `json:"category"`
+	Strengths  ClassValues `json:"strengths"`
+	Weightings ClassValues `json:"weightings"`
+}
+
+// Range is an inclusive minimum and maximum. Both bounds are kept even when
+// equal, so a later widening in the game data is a visible change rather
+// than a silent one.
+type Range struct {
+	Min int64 `json:"min"`
+	Max int64 `json:"max"`
+}
+
+// Crop is one farmable plant's yield and growth time.
+type Crop struct {
+	// ID is the buildable plant's ID, e.g. SNOWPLANT.
+	ID string `json:"id"`
+	// Substance is the item ID the plant yields, e.g. PLANT_SNOW.
+	Substance string `json:"substance"`
+	Yield     Range  `json:"yield"`
+	// GrowthSeconds is time to maturity.
+	GrowthSeconds int64 `json:"growth_seconds"`
+}
+
+// Refining is refiner throughput, which is difficulty-dependent.
+type Refining struct {
+	ProductsPerCycle   int64 `json:"products_per_cycle"`
+	SubstancesPerCycle int64 `json:"substances_per_cycle"`
+	// Survival variants. The design's open question of whether the planner
+	// should choose is settled by carrying both and letting it.
+	ProductsPerCycleSurvival   int64 `json:"products_per_cycle_survival"`
+	SubstancesPerCycleSurvival int64 `json:"substances_per_cycle_survival"`
+}
+
+// Economy is the base-economy half of the artifact.
+type Economy struct {
+	Parts    []Part    `json:"parts,omitempty"`
+	Hotspots []Hotspot `json:"hotspots,omitempty"`
+	Crops    []Crop    `json:"crops,omitempty"`
+	Refining *Refining `json:"refining,omitempty"`
+}
+
+// validate checks the economy section's internal consistency. Item
+// references are checked by the caller, which owns the item index.
+func (e *Economy) validate(knownItem func(string) bool) error {
+	seenPart := make(map[string]bool, len(e.Parts))
+	for _, p := range e.Parts {
+		if p.ID == "" {
+			return fmt.Errorf("%w: part with empty id", ErrInvalidArtifact)
+		}
+		if seenPart[p.ID] {
+			return fmt.Errorf("%w: duplicate part id %q", ErrInvalidArtifact, p.ID)
+		}
+		seenPart[p.ID] = true
+		if !p.Primary.Network.Valid() {
+			return fmt.Errorf("%w: part %q primary network %q is not a known network", ErrInvalidArtifact, p.ID, p.Primary.Network)
+		}
+		for _, d := range p.Dependencies {
+			if !d.Network.Valid() {
+				return fmt.Errorf("%w: part %q dependency network %q is not a known network", ErrInvalidArtifact, p.ID, d.Network)
+			}
+		}
+	}
+
+	seenCat := make(map[string]bool, len(e.Hotspots))
+	for _, h := range e.Hotspots {
+		if h.Category == "" {
+			return fmt.Errorf("%w: hotspot with empty category", ErrInvalidArtifact)
+		}
+		if seenCat[h.Category] {
+			return fmt.Errorf("%w: duplicate hotspot category %q", ErrInvalidArtifact, h.Category)
+		}
+		seenCat[h.Category] = true
+	}
+
+	// A part naming a hotspot must name one the artifact carries, or the
+	// planner cannot scale its rate.
+	for _, p := range e.Parts {
+		if p.Hotspot != "" && !seenCat[p.Hotspot] {
+			return fmt.Errorf("%w: part %q references hotspot %q, which is not in this artifact", ErrInvalidArtifact, p.ID, p.Hotspot)
+		}
+	}
+
+	seenCrop := make(map[string]bool, len(e.Crops))
+	for _, c := range e.Crops {
+		if c.ID == "" {
+			return fmt.Errorf("%w: crop with empty id", ErrInvalidArtifact)
+		}
+		if seenCrop[c.ID] {
+			return fmt.Errorf("%w: duplicate crop id %q", ErrInvalidArtifact, c.ID)
+		}
+		seenCrop[c.ID] = true
+		if !knownItem(c.Substance) {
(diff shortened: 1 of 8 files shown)


jonstump (commented):
## Review: Changes requested

Spec-aware review against issue #22's acceptance criteria and SPEC-0004. (Comment review — GitHub blocks `REQUEST_CHANGES` on a self-authored PR. Treat as changes-requested.)

Conflict-marker gate: clean. CI: all four checks green. Verified locally on `737d19f`: `go test ./...` passes, and all 15 `internal/normalize` tests **run** rather than skip — including both fail-closed tests, so the PR's note about the read-only-directory test is accurate.

The envelope is well-shaped and most of the determinism design is right. One finding is spec-relevant and worth fixing before #23/#24 build on it.

---

### 1. `sortArtifact` does not impose a total order on `Part.Dependencies`

Every other collection is sorted on a key that `Validate` guarantees is unique — and that pairing is clearly deliberate:

| Collection | Sort key | Uniqueness enforced by |
|---|---|---|
| `Items` | `ID` | `duplicate item id` check |
| `Recipes` | `Output`, `Method` | `duplicate %s recipe for %q` check |
| `Economy.Parts` | `ID` | `duplicate part id` check |
| `Economy.Hotspots` | `Category` | `duplicate hotspot category` check |
| `Economy.Crops` | `ID` | `duplicate crop id` check |
| **`Part.Dependencies`** | **`Network`** | **nothing** |

`Economy.validate` checks each dependency's network is *valid*, but never that a part's networks are *distinct*. `sort.Slice` is not stable, so two dependencies on the same network keep whatever relative order the caller supplied.

Confirmed — same dependency set, opposite insertion order, different bytes:

```go
d1 := domain.Dependency{Network: domain.NetworkPower, Rate: -50, Effect: "EnablesRate"}
d2 := domain.Dependency{Network: domain.NetworkPower, Rate: -20, Effect: "ReducesRate"}
// Encode(forward) != Encode(reverse)
```
```
 "dependencies": [          "dependencies": [
   { "rate": -50, ... },      { "rate": -20, ... },
   { "rate": -20, ... }       { "rate": -50, ... }
```

This contradicts REQ "Deterministic Output" — *"Collections MUST be emitted in a defined, stable order — not map-iteration order"* — and, more directly, the property this PR sets out to provide. The doc comment on `sortArtifact` says determinism "cannot be lost by a caller that happened to range over a map." For `Dependencies` it can: a partial sort lets map-iteration order show through on ties.

It matters now rather than later because **#24 is the story that populates `Dependencies`**. If it builds them by ranging a map — the exact pattern this design defends against — regeneration produces spurious diffs, which is the failure REQ "Deterministic Output" exists to prevent.

Either fix works; the second matches the pattern the rest of the file already uses:

```go
// tiebreak
sort.Slice(p.Dependencies, func(i, j int) bool {
    a, b := p.Dependencies[i], p.Dependencies[j]
    if a.Network != b.Network { return a.Network < b.Network }
    if a.Rate != b.Rate       { return a.Rate < b.Rate }
    return a.Effect < b.Effect
})

// or reject duplicates in Economy.validate, making Network total like every other key
```

### 2. Same shape in `Recipe.Inputs` — sorted by `Item` alone, with no duplicate check

`Validate` checks each input's item exists and its quantity is positive, but not that a recipe's input items are distinct. Confirmed the same way: a recipe with `AAA×1` and `AAA×7` encodes differently depending on insertion order, and `Artifact()` accepts it.

Lower severity, because a recipe naming the same input twice is malformed source data rather than a shape the game plausibly produces. But that is an argument for **rejecting** it in `Validate` rather than leaving it to sort — a duplicate input silently merged or reordered is exactly the quiet wrongness REQ "Structural Surprise Fails Loudly" is about.

### 3. The determinism test does not exercise the Economy section

`TestOutputIsByteIdenticalRegardlessOfInsertionOrder` covers `Items`, `Recipes`, and `Recipe.Inputs` — and its inputs use distinct item IDs, so it only ever exercises keys that are already total. No `Economy` is attached to either builder.

Since the economy section is this PR's main addition and the one place the ordering is partial, that is why the gap survived. Whatever fix you pick for finding 1, the regression test wants a `Part` with two same-network dependencies added in opposite orders.

---

### Acceptance criteria

| Criterion (issue #22) | Status |
|---|---|
| `LoadTier1` accepts an artifact carrying the new base-economy sections | Met — `TestEconomySectionLoads` |
| Both committed fixtures still load and validate after the schema change | Met — both migrated to schema 2 in the same commit |
| `SchemaVersion` is incremented | Met — 1 → 2, with `Validate` refusing the other side of the change |
| Two runs over an unchanged install produce byte-identical output | **Partial** — holds for items and recipes; findings 1 and 2 above |
| All four sentinels defined and distinguishable via `errors.Is` | Met — `TestSentinelsAreDistinguishable` |
| A corrupted source table produces a named error and leaves no artifact behind | Met — both failure points tested, both run rather than skip |

### What is right and worth keeping

**Sort-key totality is enforced structurally, not by convention.** Four of six collections pair their sort key with a matching duplicate check in `Validate`. That is the correct pattern and clearly intentional — findings 1 and 2 are the two places it was not carried through, not a missing idea.

**No generation timestamp, with a test asserting its absence.** The right call, and testing for it is what stops someone re-adding it later thinking it is free.

**Both fail-closed paths are tested separately.** "Generation fails before the writer" and "write fails, previous artifact intact" are genuinely different failure points, and the temp-file-and-rename means a reader never sees a half-written artifact.

**Economy shapes mirror the source rather than tidying it.** Rates per-network because a part participates in several; class on the hotspot because that is where the game puts it; ranges keeping both bounds. The `ClassValues`-on-hotspot decision in particular explains a real dead end — searching parts for `_C`/`_B`/`_A` finds nothing — and recording that in the type comment saves the next reader the same search.

**Sentinels mirroring `hgpak.StructureError`** is the right instinct for two packages in one pipeline. Giving localisation its own sentinel because the *remedy* differs is a better reason than symmetry.

### Architecture references

ADR-0001 and SPEC-0004 governing comments are present across `economy.go`, `tier1.go`, `artifact.go`, and `errors.go`. The PR correctly ties the economy section back to the 2026-08-18 Tier 2 finding (#18) that moved these constants from curated to extracted. No missing ADR or issue reference found.

jonstump:
## Review feedback addressed — findings 1, 2, and 3 in 70fa668

### Finding 1 — `Part.Dependencies` ordering — fixed

Falls through the remaining fields rather than stopping at `Network`:

```go
sort.Slice(p.Dependencies, func(i, j int) bool {
    a, b := p.Dependencies[i], p.Dependencies[j]
    if a.Network != b.Network { return a.Network < b.Network }
    if a.Rate != b.Rate       { return a.Rate < b.Rate }
    return a.Effect < b.Effect
})
```

Two dependencies that still compare equal after all three fields are byte-identical, so their relative order cannot appear in the output. That makes the key total for the purpose the requirement cares about.

**Tiebreaking rather than rejecting duplicates in `Validate`.** Both were on the table in the review. Rejecting asserts that the game data never carries two dependencies on one network — which has not been established, and generation refusing over an unverified assumption about the source is the failure mode this pipeline has already had once. Tiebreaking is correct either way; the rejection stays available once #24 sees the real tables and can answer the question with data.

### Finding 2 — `Recipe.Inputs` ordering — fixed, sort only

Same treatment: `Item`, then `Quantity`.

Note this is the **conservative half** of the review's suggestion. The review argued a recipe naming the same input twice should be *refused* under REQ "Structural Surprise Fails Loudly", and I think that is right — but it is a behaviour change that would start failing generation on artifacts that validate today, so it is your call rather than something to fold into an ordering fix. The nondeterminism is gone either way; the question of whether duplicate inputs should be legal at all is still open.

### Finding 3 — determinism test coverage — fixed

`TestEconomyOrderingIsTotalAcrossNestedSlices` builds the same artifact from opposite insertion orders with:

- two dependencies on one network, distinguishable only by `Rate` and `Effect`
- a pair distinguishable only by `Effect` — exercising the last tiebreaker specifically
- a recipe naming the same input item twice at different quantities
- an `Economy` attached at all, which the existing test never did

**It fails without the fix and passes with it.** Worth recording the other half of that check: `TestOutputIsByteIdenticalRegardlessOfInsertionOrder` still **passes** with the fix reverted. It attaches no `Economy` and uses distinct input IDs, so it only ever exercised keys that were already total — which is precisely why this survived into review.

### Verification

| Check | Result |
|---|---|
| `go test ./...` | all packages ok |
| `go vet ./...` | clean |
| `gofmt -l .` | no output |
| Fixtures load at schema 2 | unchanged, still pass |

`sortArtifact`'s doc comment was also updated — it claimed "where one field can repeat, a second breaks the tie", which described the intent rather than the code. It now says what the code does.

PR #30, opened 2026-08-19
feat(domain): recipe lists, explicit yields, deterministic selection

### LINKED ISSUE #29: Recipe lists, yields, and deterministic selection in the engine
Part of #21. Governed by [SPEC-0001](docs/openspec/specs/rollup-engine/spec.md) as amended by [ADR-0005](docs/adrs/ADR-0005-multiple-recipes-per-output.md).

Changes `internal/domain` — the schema and the merged rollup engine — so the artifact can hold what the game actually contains. Split out of #23 because it touches shipped, tested code and because the selection rule is real traversal work rather than a map-type change.

**Blocks #23**, which cannot emit recipe lists until the schema accepts them.

## Why this is its own story

`Tier1.Validate` currently *rejects* duplicate recipes per output and method. ADR-0005 established that 261 of 403 refiner output/method pairs have more than one — so the check has to invert from rejecting duplicates to indexing them. That plus the selection rule lands in `resolve.go` and `tier1.go`, both merged in #2 and #26, with 39 existing scenarios' worth of behaviour to not break.

## Requirements

- [ ] **Recipe Selection** (SPEC-0001) — A method does not identify a recipe; each node MUST resolve to exactly one recipe. The engine MUST select a default deterministically: the candidate whose expansion resolves to the smallest total of raw inputs, ties broken by a stable recipe identifier, identical on every run and machine. The engine MUST report which recipes are legal for a node's chosen method alongside the legal methods. Changing a node's recipe MUST change its expansion and recompute all derived totals. A node using its default MUST be representable without recording a selection.
  - WHEN a node's item has more than one recipe for its chosen method THEN the engine expands the candidate with the smallest raw-input total, and repeated runs over the same artifact select the same one
  - WHEN the view asks for a node's options THEN it receives every legal recipe for the chosen method, not only the selected one
  - WHEN a node's recipe changes to another legal recipe for the same method THEN the node's child edges change to that recipe's inputs, and all ancestor quantities are recomputed
  - WHEN a recipe is requested for a node that has no such recipe in the Tier 1 artifact THEN the engine returns an error naming the node and the recipe, and the plan is left unchanged
  - WHEN every node in a plan uses its default recipe THEN the serialized plan records no recipe selections

- [ ] **Exact Arithmetic and Rounding Discipline** (SPEC-0001, amended) — All item quantities MUST be exact integers, including a recipe's **yield**. A recipe producing *y* units to satisfy a demand of *n* MUST be applied as exact arithmetic over *n* and *y*, never as floating-point division. The engine MUST NOT accumulate binary floating-point error across the graph.
  - WHEN a recipe producing 50 units satisfies a demand of 125 THEN the batch count is computed exactly and rounded up, with no floating-point division in the path
  - WHEN a plan resolves through recipes with yields other than one THEN no total differs from the exact rational result

## Scope

**In:**
- `Recipe` gains a yield field (default 1) and a stable identifier
- `recipesByOut` becomes `map[string]map[Method][]Recipe`; `Validate` inverts from rejecting duplicates to indexing them
- `PlanInput` gains a per-item recipe override, mirroring the existing `Methods` map — absent means default, so defaults cost nothing
- `ResolvedGraph` nodes report legal recipes alongside legal methods
- The default-selection rule in `resolve.go`, including the raw-input comparison

**Out:**
- Parsing recipes from source tables — that is #23
- The WASM adapter — SPEC-0002's REQ "Recipe Selection Crossing" has no implementation to amend yet; no adapter exists
- UI for choosing a recipe — view layer, later

## Acceptance Criteria

- An artifact carrying 26 refine recipes for one output loads and validates
- `Recipe` carries a yield; a recipe with yield 50 satisfying demand 125 produces exactly 3 batches, computed without floating-point division
- Two runs over the same artifact select the same default recipe for every node
- A node's reported options include every legal recipe for its method
- Requesting a recipe the artifact lacks errors, naming node and recipe, leaving the plan unchanged
- A plan where every node uses its default serializes with no recipe selections
- The existing 34-node Stasis Device fixture still resolves to the same totals — 500 Sulphurine, 500 Nitrogen, 500 Radon, 300 Condensed Carbon — since crafting is 1:1 and unaffected
- All existing `internal/domain` tests pass, migrated where the schema change requires it

### Branch

`feat/29-recipe-lists-yields-and-selection`

### PR Convention

```
Closes #29
Part of #21

Implements SPEC-0001 REQ "Recipe Selection", REQ "Exact Arithmetic and
Rounding Discipline" (as amended by ADR-0005).
```

jonstump:
## What

Implements #29 — the schema and engine half of ADR-0005.

A method does not identify a recipe. 261 of 403 refiner output/method pairs carry more than one, and 156 of 1,681 refiner recipes produce a quantity other than one, up to 250. The old schema rejected the first as a duplicate and silently assumed the second was 1, which made every refining total wrong by up to 250×.

## Changes

| Area | Change |
|---|---|
| `Recipe` | gains `ID` (stable, what a plan records) and optional `Yield` (absent means one, via `Producing()`) |
| `Tier1` | `recipesByOut` becomes `output → method → []Recipe`; `Validate` now rejects a duplicate **recipe id** rather than a duplicate method |
| `Tier1` accessors | `RecipesFor(id, m)`, `Recipe(item, m, recipeID)`, `LegalRecipes(id, m)` |
| `PlanInput` | gains `Recipes map[string]string` — item → recipe id, mirroring `Methods`; absence means default |
| `Node` | gains `Recipe`, `Yield`, `LegalRecipes`, and `Applications()` (exact, unrounded) |
| `Edge` | gains `Yield`, kept separate from `PerUnit` so the edge still reports what the recipe literally says |
| `propagate` | multiplies by `PerUnit/Yield` as a `big.Rat` — a yield of 50 stays `1/50`, never `0.02` |
| `select.go` | new: default selection by smallest raw-input total, ties broken by recipe id, memoized with a cycle guard |
| `normalize` | recipe sort key extended with `ID`, since output+method is no longer total |

## Selection rule

Per SPEC-0001 REQ "Recipe Selection": the candidate whose expansion resolves to the smallest total of raw inputs, ties broken by recipe id. `rawCost` is memoized per item and carries an in-progress set, so a self-referential refiner recipe costs as unresolvable and is never preferred over a route that terminates — it remains selectable only if it is the sole option.

`Applications()` is deliberately **unrounded**. SPEC-0001 enumerates the boundaries at which rounding up is allowed and a recipe application is not one of them, so the engine hands back the exact rational and a caller rounds once, at the point it reports batches.

## Verification

- 26 refine recipes for one output load, validate, and are all reported as legal options
- Demand 125 at yield 50 → Crystal Sulphide total `5/2` exactly, `Applications()` `5/2`, ceiling 3 batches, computed by `big.Int.QuoRem` with no float anywhere
- 25 repeated resolutions select the same recipe for every node
- An override onto a costlier route changes the expansion and the totals; an unknown recipe id errors naming both node and recipe, returning no graph
- The 34-node Stasis Device fixture still resolves to **500 Sulphurine, 500 Nitrogen, 500 Radon, 300 Condensed Carbon**
- `go vet`, `staticcheck v0.8.0-rc.1`, `gofmt -l` all clean

## Migration

Fixtures gained recipe ids (`{output}_{method}`). Two `internal/domain` tests moved from `Recipe(id, method)` to the list-shaped accessors.

Closes #29
Part of #21

Implements SPEC-0001 REQ "Recipe Selection", REQ "Exact Arithmetic and Rounding Discipline" (as amended by ADR-0005).

### DIFF
--- a/internal/domain/resolve.go
+++ b/internal/domain/resolve.go
@@ -23,6 +23,17 @@ type PlanInput struct {
 	// Methods overrides the per-item default method. Items absent from the
 	// map use their Tier 1 default.
 	Methods map[string]Method
+
+	// Recipes overrides the per-item recipe, keyed by item ID and holding a
+	// recipe ID. Items absent from the map use the engine's default choice.
+	//
+	// Absence is meaningful: SPEC-0001 REQ "Recipe Selection" requires that a
+	// node using its default be representable without recording a selection,
+	// so that plan state — and the URL hash it serializes into — carries only
+	// deliberate overrides.
+	//
+	// Governing: SPEC-0001 REQ "Recipe Selection"
+	Recipes map[string]string
 }
 
 // Edge is one resolved dependency, carrying the per-unit-of-parent quantity.
@@ -31,8 +42,20 @@ type Edge struct {
 	From string
 	// To is the consumed (child) item ID.
 	To string
-	// PerUnit is how many To are needed per single unit of From.
+	// PerUnit is how many To the recipe consumes per application.
 	PerUnit int64
+
+	// Yield is how many From one application produces. Per unit of From the
+	// requirement is therefore PerUnit/Yield, which is rational — 1 Crystal
+	// Sulphide yields 50 Sodium Nitrate, so a unit of Sodium Nitrate costs
+	// one fiftieth of a Crystal Sulphide.
+	//
+	// Kept as a separate integer rather than folded into PerUnit so the edge
+	// still reports what the recipe literally says, and so the division
+	// happens once, exactly, in propagate.
+	//
+	// Governing: SPEC-0001 REQ "Exact Arithmetic and Rounding Discipline"
+	Yield int64
 }
 
 // Node is one item in the resolved graph, with its aggregated total.
@@ -46,6 +69,23 @@ type Node struct {
 	// LegalMethods lists every method available to this item, sorted, so the
 	// view can render unavailable options as inert rather than hiding them.
 	LegalMethods []Method
+
+	// Recipe is the id of the single recipe this node resolved to. Empty for
+	// terminal nodes, which expand through no recipe.
+	//
+	// Governing: SPEC-0001 REQ "Recipe Selection"
+	Recipe string
+
+	// LegalRecipes lists every recipe id available for this node's chosen
+	// method, sorted, so the view can offer the alternatives rather than
+	// presenting one route as though it were the only one.
+	LegalRecipes []string
+
+	// Yield is how many units one application of the resolved recipe
+	// produces. Zero for terminal nodes, which apply no recipe.
+	//
+	// Governing: ADR-0005 (explicit yields)
+	Yield int64
 
 	// Terminal reports that expansion stopped here (method raw).
 	Terminal bool
@@ -84,6 +124,20 @@ func (n *Node) TotalInt() (int64, bool) {
 		return 0, false
 	}
 	return n.total.Num().Int64(), true
+}
+
+// Applications returns how many times this node's recipe must be applied to
+// meet its total, as an exact rational. Nil for a terminal node.
+//
+// Deliberately unrounded: SPEC-0001 REQ "Exact Arithmetic and Rounding
+// Discipline" enumerates the boundaries at which rounding up is allowed, and
+// a recipe application is not one of them. A caller wanting whole batches
+// rounds this value itself, once, at the point it reports them.
+func (n *Node) Applications() *big.Rat {
+	if n.Terminal || n.Yield == 0 {
+		return nil
+	}
+	return new(big.Rat).Quo(n.total, new(big.Rat).SetInt64(n.Yield))
 }
 
 // ResolvedGraph is stage 1's output.
@@ -149,7 +203,7 @@ func Resolve(t *Tier1, in PlanInput) (*ResolvedGraph, error) {
 		return nil, fmt.Errorf("resolving %s: %w: %q", in.Target, ErrUnknownItem, in.Target)
 	}
 
-	r := &resolver{t: t, in: in, state: map[string]dfsState{}, nodes: map[string]*Node{}}
+	r := &resolver{t: t, in: in, state: map[string]dfsState{}, nodes: map[string]*Node{}, costs: map[string]*big.Rat{}}
 	if err := r.walk(in.Target, nil); err != nil {
 		return nil, fmt.Errorf("resolving %s: %w", targetItem.Name, err)
 	}
@@ -172,6 +226,9 @@ func Resolve(t *Tier1, in PlanInput) (*ResolvedGraph, error) {
 }
 
 type resolver struct {
+	// costs memoizes rawCost per item: nil means known-unresolvable.
+	costs map[string]*big.Rat
+
 	t     *Tier1
 	in    PlanInput
 	state map[string]dfsState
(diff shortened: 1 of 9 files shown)


jonstump (commented):
## Review: Approved

Spec-aware review against issue #29's acceptance criteria, ADR-0005, and SPEC-0001. Verdict: **approve**. (Comment review — GitHub blocks `APPROVE` on a self-authored PR.)

Conflict-marker gate clean, CI green on all four checks. Verified locally on `a7b6a21`: `go test -race ./...` passes across all packages.

No findings. The things most likely to be wrong in a change this shaped are all handled, and I checked each rather than assuming:

### The `(output, method)` uniqueness removal is followed through

Making `recipesByOut` a `map[string]map[Method][]Recipe` invalidates the uniqueness that made `sortArtifact`'s recipe sort key total — the exact property the #26 review leaned on. The sort key correctly grew an `ID` tiebreaker in the same PR, with a comment naming ADR-0005 as the reason.

The replacement invariant holds where it needs to: `Validate` now rejects duplicate recipe **ids within an (output, method) group**, which is precisely the scope the sort key needs, and it matches how `Recipe(id, m)` looks up. Narrower than global id uniqueness, and correctly so — ids like `RECIPE_1` may legitimately repeat across different outputs.

### Yields are applied where the arithmetic actually happens

`propagate` divides at the edge — `big.NewRat(e.PerUnit, e.Yield)` — rather than pre-folding into `PerUnit`. That keeps 1/50 of a Crystal Sulphide exactly 1/50, and it means the edge still reports what the recipe literally says. Both are the right call.

`Edge.Yield` can never be zero: the single construction site sets `Yield: rc.Producing()`, `Producing()` maps absent-to-one, and `Validate` rejects a negative yield. I checked for a second construction site because `big.NewRat` panics on a zero denominator — there isn't one.

### `Applications()` returning unrounded is correct, not an omission

Issue #29's criterion reads "a recipe with yield 50 satisfying demand 125 produces exactly 3 batches", and `Applications()` returns `5/2`. That looks like a gap until you check SPEC-0001 REQ "Exact Arithmetic and Rounding Discipline", which enumerates the rounding boundaries — plants, biodomes, extractors, supply depots, fauna, nutrient processors, generators, batteries. A recipe application is deliberately not among them.

So the engine returning an exact rational and leaving the round to the caller is the spec-compliant reading, and `TestYieldAppliesExactly` covers the criterion anyway with a `ceil` helper doing integer `QuoRem` — no float involved at any point.

### Selection is genuinely deterministic

`selectRecipe`'s case analysis is complete: a resolvable candidate always beats an unresolvable one, equal-resolvability falls to cost, and cost ties fall to id. `better` handles the both-nil case by comparing ids, so even an all-unresolvable candidate set picks reproducibly rather than by artifact order.

The memoization in `rawCost` is the part I expected to find a bug in and did not. Blocking `inProgress` items is correct — an item in progress is an ancestor, so a path back to it is a genuine cycle, which means the excluded paths are exactly the ones that could not contribute to an acyclic cost. And the `len(inProgress) == 1` guard on memoizing *failures* is the right instinct: a context-dependent failure must not be cached as a general one. That guard is load-bearing and easy to delete by accident; it earns the comment it has.

### Acceptance criteria

| Criterion (issue #29) | Status |
|---|---|
| An artifact carrying 26 refine recipes for one output loads and validates | Met |
| Yield 50, demand 125 → exactly 3 batches, no float division | Met — `TestYieldAppliesExactly`, exact `ceil` via `QuoRem` |
| Two runs select the same default recipe for every node | Met — cost then id, both deterministic |
| A node's options include every legal recipe for its method | Met — `LegalRecipes`, sorted |
| Requesting an absent recipe errors naming node and recipe | Met |
| A plan where every node uses its default serializes with no selections | Met — override map absence is meaningful |
| Stasis Device still resolves to 500/500/500/300 | Met — `TestResolveStasisDeviceTree`, `TestStasisDeviceLeafTotals` both pass |
| Existing `internal/domain` tests pass, migrated where needed | Met |

### Merge ordering

This PR and #31 both modify `internal/normalize/artifact.go`. They auto-merge cleanly (verified), but this one is the foundation story and should land first; #31 then rebases onto it.

PR #31, opened 2026-08-19
feat(normalize): extract base economy data from the real tables

### LINKED ISSUE #24: Base economy data

jonstump:
## What

Implements #24. The four base-economy constants ADR-0001 planned to hand-curate are now generated from the game tables, so they are version-stamped and regenerated rather than maintained by hand. Only biodome crop capacity remains curated — it is not in the tables.

## Where the values come from

| Source | Gives |
|---|---|
| `metadata/reality/tables/basebuildingobjectstable` | per-part rate, storage, dependent draws, hotspot dependence |
| `metadata/simulation/scanning/regionhotspotstable` | C/B/A/S class strengths and weightings |
| `metadata/reality/tables/rewardtable` | crop yields |
| `models/.../interactiveflora/*/entities/plantinteraction.entity` | which crop yields what |
| `gcgameplayglobals.global` | refiner throughput, including the Survival variants |

Nothing in the buildables table says what a plant yields. The chain is: a buildable whose primary connection is `PlantGrowth` (growth time is that connection's `Storage`) → its `PlacementScene` names a flora directory → that directory's interaction entity gives a reward key → the reward table gives the substance and its min/max. That derivation is recorded in the artifact under `economy.searches`, per SPEC-0004 REQ "Search Boundaries Are Recorded".

## Two things the source does that the spec's wording did not anticipate

Both are implemented correctly and recorded on the search note rather than quietly handled:

1. **Five of the twelve farmable plants yield a *product*, not a substance** — `GcRewardSpecificProduct` with an identical `Amount` shape. SPEC-0004 REQ "Base Economy Data" names only `GcRewardSpecificSubstance`. Reading only that form would have silently dropped venom sacs, gravitino balls, nip-nip buds, albumen pearls and creature pellets. **This wording is worth amending in a follow-up.**
2. **The reward key and the substance are not always the same** — `PLANT_BARREN` yields `PLANT_DUST`. Treating the key as the substance would be wrong in exactly one of twelve cases, which is the worst kind of wrong.

## Modelling decisions

- **Hotspot families.** The table names six categories (`Power`, `Mineral1..3`, `Gas1..2`) but a part's `DependsOnHotspots` names a family (`Mineral`). The numbered variants exist to give a planet several distinct hotspots, not to scale differently, so they collapse to the family. The collapse is **guarded**: members that disagree fail loudly rather than letting one variant's numbers stand in for the rest.
- **Parts with no economic behaviour are skipped.** All 1,997 buildables carry `LinkGridData`; 108 have a non-zero rate, storage, dependency or hotspot dependence. The rest are decorations.
- **A growth-network container is not a crop.** `CARBONPLANTER` sits on `PlantGrowth` but has no flora entity and no yield. It is carried as a part and named in the search note, not dropped.
- **`Fuel` and `Portals` added to the network vocabulary** so a part using one is emitted rather than rejected.

## Verification against the real install

Run against a full 6.45.0.1 decompilation of NMS 5.97 — 108 parts, 3 hotspot families, 12 crops:

- `U_EXTRACTOR_S` → rate 100, storage 360000, dependent Power −50 `EnablesRate`, hotspot `Mineral`
- Power 150/220/250/300; Mineral and Gas 1/1.5/2/2.5 — keyed by category, never by device
- Frost Crystal 50/50, Cactus Flesh 100/100, Star Bulb 25/25, growth 3600 s for frostwart
- Refining 2 products / 250 substances per cycle; 1 / 100 on Survival
- No per-class device variants anywhere in the output

## Tests

Fixtures under `internal/normalize/testdata/economy/` are **real content sliced out of a real decompilation** — see `testdata/economy/gen.go` for exactly which rows and why. Hand-authored MXML would test only that the parser agrees with my idea of the format, which is the mistake that cost this project its first parser.

26 test cases including six fail-closed scenarios: a missing source, an unknown link network, a hotspot family whose members disagree, a crop whose reward key is unresolvable, a crop whose flora entity is absent, and a refiner field that is gone. Each must produce a `*SourceError` naming the table and return no partial economy.

## Note on ordering

`internal/normalize/mxml.go` comes over from the parked #23 work, because the economy pass needs the MXML reader and lands first. #23 should drop its copy and build the recipe tables on top of this one.

Closes #24
Part of #21

Implements SPEC-0004 REQ "Base Economy Data", REQ "Search Boundaries Are Recorded".

### DIFF
--- a/internal/domain/economy.go
+++ b/internal/domain/economy.go
@@ -193,10 +193,22 @@ func (e *Economy) validate(knownItem func(string) bool) error {
 		}
 	}
 
+	seenSearch := make(map[string]bool, len(e.Searches))
 	for _, s := range e.Searches {
 		if s.Value == "" {
 			return fmt.Errorf("%w: search record with empty value", ErrInvalidArtifact)
 		}
+		// Value is what sortArtifact orders Searches by, so it has to be
+		// unique or the emitted order falls back to however the normalizer
+		// happened to append them — the map-iteration order REQ
+		// "Deterministic Output" forbids. Rejecting rather than tiebreaking:
+		// unlike the game tables, these records are ours, so two searches
+		// deriving one value is an authoring mistake — either a duplicate to
+		// merge or two derivations wanting distinguishable names.
+		if seenSearch[s.Value] {
+			return fmt.Errorf("%w: duplicate search record for %q", ErrInvalidArtifact, s.Value)
+		}
+		seenSearch[s.Value] = true
 		if len(s.Searched) == 0 {
 			// A search record that names no sources records nothing; it
 			// would read as diligence while carrying none.
--- a/internal/domain/economy.go
+++ b/internal/domain/economy.go
@@ -26,11 +26,18 @@ const (
 	NetworkResources   Network = "resources"
 	NetworkPlantGrowth Network = "plant_growth"
 	NetworkByteBeat    Network = "byte_beat"
+	// Fuel and Portals appear once and three times respectively in the
+	// source. They carry no planner meaning today, but the vocabulary
+	// mirrors GcLinkNetworkTypes rather than a filtered subset of it, so
+	// that a part using one is emitted rather than rejected.
+	NetworkFuel    Network = "fuel"
+	NetworkPortals Network = "portals"
 )
 
 var validNetworks = map[Network]bool{
 	NetworkPower: true, NetworkResources: true,
 	NetworkPlantGrowth: true, NetworkByteBeat: true,
+	NetworkFuel: true, NetworkPortals: true,
 }
 
 // Valid reports whether n is a known network.
@@ -114,12 +121,35 @@ type Refining struct {
 	SubstancesPerCycleSurvival int64 `json:"substances_per_cycle_survival"`
 }
 
+// Search records how a value was located when it was found by searching
+// rather than read from a known field.
+//
+// Governing: SPEC-0004 REQ "Search Boundaries Are Recorded" — this exists
+// because this project has three recorded instances of a bounded search
+// being reported as a general result. Recording where we looked makes the
+// boundary of a negative result visible to the next reader.
+type Search struct {
+	// Value names what was being derived, e.g. "crop substance".
+	Value string `json:"value"`
+	// Searched lists the sources consulted, in the order consulted.
+	Searched []string `json:"searched"`
+	// Note records what the search concluded, including anything it did
+	// not cover.
+	Note string `json:"note,omitempty"`
+}
+
 // Economy is the base-economy half of the artifact.
 type Economy struct {
 	Parts    []Part    `json:"parts,omitempty"`
 	Hotspots []Hotspot `json:"hotspots,omitempty"`
 	Crops    []Crop    `json:"crops,omitempty"`
 	Refining *Refining `json:"refining,omitempty"`
+
+	// Searches records the derivation of values the normalizer had to hunt
+	// for. Sorted by Value.
+	//
+	// Governing: SPEC-0004 REQ "Search Boundaries Are Recorded"
+	Searches []Search `json:"searches,omitempty"`
 }
 
 // validate checks the economy section's internal consistency. Item
@@ -160,6 +190,17 @@ func (e *Economy) validate(knownItem func(string) bool) error {
 	for _, p := range e.Parts {
 		if p.Hotspot != "" && !seenCat[p.Hotspot] {
 			return fmt.Errorf("%w: part %q references hotspot %q, which is not in this artifact", ErrInvalidArtifact, p.ID, p.Hotspot)
+		}
+	}
+
+	for _, s := range e.Searches {
+		if s.Value == "" {
+			return fmt.Errorf("%w: search record with empty value", ErrInvalidArtifact)
+		}
+		if len(s.Searched) == 0 {
+			// A search record that names no sources records nothing; it
+			// would read as diligence while carrying none.
+			return fmt.Errorf("%w: search record %q lists no sources searched", ErrInvalidArtifact, s.Value)
 		}
 	}
 
--- a/internal/normalize/artifact.go
+++ b/internal/normalize/artifact.go
@@ -151,8 +151,10 @@ func sortArtifact(t *domain.Tier1) {
 		return t.Economy.Hotspots[i].Category < t.Economy.Hotspots[j].Category
 	})
 	sort.Slice(t.Economy.Crops, func(i, j int) bool { return t.Economy.Crops[i].ID < t.Economy.Crops[j].ID })
-	// Searches sort by value; the Searched list inside each is deliberately
-	// left in the order it was consulted, since that sequence is the record.
+	// Searches sort by value, which Validate guarantees unique — the same
+	// key-plus-uniqueness pairing every other collection here uses. The
+	// Searched list inside each is deliberately left in the order it was
+	// consulted, since that sequence is the record rather than a set.
 	sort.Slice(t.Economy.Searches, func(i, j int) bool {
 		return t.Economy.Searches[i].Value < t.Economy.Searches[j].Value
 	})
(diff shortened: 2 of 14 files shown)


jonstump (commented):
## Review: Changes requested

Spec-aware review against issue #24's acceptance criteria and SPEC-0004. (Comment review — GitHub blocks `REQUEST_CHANGES` on a self-authored PR. Treat as changes-requested.)

Conflict-marker gate clean, CI green on all four checks. Verified locally on `0fc628a`: `go test -race ./...` passes across all packages.

One finding, and it is the same shape as the one fixed on #26 two PRs ago.

---

### `Economy.Searches` sorts on a key that is not total

`sortArtifact` gained:

```go
sort.Slice(t.Economy.Searches, func(i, j int) bool {
    return t.Economy.Searches[i].Value < t.Economy.Searches[j].Value
})
```

`Economy.validate` checks each search has a non-empty `Value` and at least one entry in `Searched`, but it does **not** reject two searches sharing a `Value`. `sort.Slice` is unstable, so a duplicate `Value` leaves the caller's insertion order showing through.

Confirmed — two `Search` records with `Value: "crop substance"`, opposite insertion order, different bytes:

```go
s1 := domain.Search{Value: "crop substance", Searched: []string{"tableA"}, Note: "first"}
s2 := domain.Search{Value: "crop substance", Searched: []string{"tableB"}, Note: "second"}
// Encode(forward) != Encode(reverse)
```

This is the pattern the rest of the file already gets right: every other collection pairs its sort key with a uniqueness check in `validate` — part IDs, hotspot categories, crop IDs, and (since #26) full tiebreak chains on `Dependencies` and `Inputs`. `Searches` is the one that did not carry it through, and it is the collection this PR adds.

Lower severity than the #26 case, honestly: `Searches` records are authored by the normalizer rather than read out of the game tables, so a duplicate `Value` is a choice someone would have to make rather than something the data could hand you. But that was the argument for `Dependencies` too, and REQ "Deterministic Output" is unconditional — *"Collections MUST be emitted in a defined, stable order — not map-iteration order."*

Either fix works:

```go
// tiebreak on the remaining fields
if a.Value != b.Value { return a.Value < b.Value }
if a.Note != b.Note   { return a.Note < b.Note }
return strings.Join(a.Searched, "\x00") < strings.Join(b.Searched, "\x00")

// or reject duplicate Values in Economy.validate, making Value total like every other key
```

Here I would lean toward **rejecting** rather than tiebreaking, the opposite of the #26 call. Two search records deriving the same value are either a duplicate that should be merged or two genuinely different derivations that want distinguishing names — both are authoring mistakes worth surfacing, and unlike `Dependencies` there is no unverified question about what the game data might contain, because this data is ours.

### Leaving `Searched` unsorted is right

Worth saying explicitly since it looks like the same omission and is not: the inner `Searched` list is deliberately left in consultation order, and the comment says so. That order *is* the record — REQ "Search Boundaries Are Recorded" exists to make the boundary of a negative result legible, and sorting the sequence would destroy the thing being recorded. Correct call, correctly explained.

---

### Acceptance criteria

| Criterion (issue #24) | Status |
|---|---|
| `U_EXTRACTOR_S` emits rate 100, storage 360000, power −50 tagged `EnablesRate` | Met — asserted directly |
| Hotspot class strengths 150/220/250/300 and 1/1.5/2/2.5, keyed by category and class | Met |
| Crop yields as ranges — Frost Crystal 50/50, Cactus Flesh 100/100, Star Bulb 25/25 | Met |
| Crop growth times in seconds — frostwart 3600 | Met |
| Refiner throughput emits standard and Survival variants | Met |
| No per-class device variants anywhere in the output | Met |
| Derived values carry the list of sources searched | Met — modulo the ordering finding above |

The `U_EXTRACTOR_S` trailing-`S`-is-a-size-tier note in the test file is a good catch to have written down — it is exactly the kind of thing that would otherwise be re-derived by the next person parsing part IDs for class suffixes.

### Merge ordering

This PR and #30 both modify `internal/normalize/artifact.go`, in different hunks of `sortArtifact`. I verified they auto-merge cleanly. #30 is the foundation story (ADR-0005 recipe lists) and should land first; this one rebases onto it. Worth re-running the economy tests after the rebase, since #30 changes `Recipe` and the shared fixtures.

jonstump:
## Review feedback addressed — `Searches` ordering fixed in 16be094

Also rebased onto `main` now that #30 has merged.

### The fix: reject rather than tiebreak

`Economy.validate` now refuses two search records deriving the same value, which makes `Value` a total sort key the same way part ids, hotspot categories, crop ids, and recipe ids are for their collections:

```go
if seenSearch[s.Value] {
    return fmt.Errorf("%w: duplicate search record for %q", ErrInvalidArtifact, s.Value)
}
```

This is deliberately the **opposite** call from the one made for `Dependencies` and `Inputs` on #26, where I tiebroke instead. The difference is where the data comes from. There the open question was what the game tables might contain, and refusing to generate over an unverified assumption about the source is a failure this project has already had once. These records are written by the normalizer — they are ours — so two searches deriving one value is an authoring mistake worth surfacing: either a duplicate that wants merging, or two derivations that want distinguishable names.

`sortArtifact`'s comment now states the invariant it depends on rather than just describing the sort, so the next person changing `validate` can see what the sort is relying on.

### The inner list is unchanged

`Searched` still keeps consultation order. That sequence is the record REQ "Search Boundaries Are Recorded" exists to preserve — sorting it would destroy the thing being recorded.

### Tests

**`TestDuplicateSearchValueIsRefused`** — a crafted duplicate is refused with `ErrInvalidArtifact` naming the value. Confirmed to **fail with the fix reverted**, so it is a real regression test.

**`TestRealSearchValuesAreUnique`** — the search values the extractor actually emits from the real tables are checked unique. Without it the new guard would only ever be exercised on synthetic input, and a future normalizer change that started emitting a duplicate would fail generation with nothing pointing at why.

### Verification after rebase

| Check | Result |
|---|---|
| `go test ./...` | all packages ok |
| `go vet ./...` | clean |
| `gofmt -l .` | no output |

The economy tests were re-run against #30's merged `Recipe` changes and shared fixtures specifically — the rebase was textually clean, but #30 altered both, so a clean merge was not sufficient evidence on its own.

Issue #25, opened 2026-08-19
Acceptance against generated output


jonstump:
Part of #21. Governed by [SPEC-0004](docs/openspec/specs/tier1-normalizer/spec.md).

**Blocked by #23** — needs the recipe graph the acceptance test resolves over.

Wires ADR-0001's confirmation criteria to run through the merged rollup engine over a **generated** artifact, and regenerates and commits the real dataset. This is the story that proves the pipeline end to end.

The distinction that makes this its own story rather than a checkbox on #23: a test that resolves the tree from `testdata/stasis-device.tier1.json` verifies the rollup engine, which #2 already did. Only a test whose input the normalizer produced in that same run verifies the normalizer.

## Requirements

- [ ] **Acceptance Against the Golden Tree** (SPEC-0004) — The normalizer is correct when its artifact reproduces ADR-0001's confirmation criteria through the merged rollup engine, rooted at `ULTRAPROD2`: 34 distinct nodes across the Quantum Processor (`MEGAPROD2`), Cryogenic Chamber (`MEGAPROD3`) and Iridesite (`ALLOY8`) branches; each gas product costing 250 gas and 50 Condensed Carbon; and at quantity ×1, 500 each of Sulphurine, Nitrogen and Radon plus 300 Condensed Carbon. This acceptance MUST run against a generated artifact, not against the hand-authored fixtures.
  - WHEN the rollup engine resolves `ULTRAPROD2` at quantity 1 from a generated artifact THEN it returns 34 distinct nodes and leaf totals of 500 Sulphurine, 500 Nitrogen, 500 Radon and 300 Condensed Carbon
  - WHEN the acceptance test runs THEN its input was produced by the normalizer in that run, not read from a committed fixture

## Acceptance Criteria

- The acceptance test generates an artifact and resolves `ULTRAPROD2` through `internal/domain` in the same run
- It returns exactly 34 distinct nodes across the three named branches
- Leaf totals at ×1 are 500 Sulphurine, 500 Nitrogen, 500 Radon, 300 Condensed Carbon
- `REACTION1`/`2`/`3` each show 250 gas + 50 Condensed Carbon
- The test is gated on a real install and skips cleanly when the environment variable is unset, so CI stays green without game files
- The generated artifact is committed, and `stasis-device.tier1.json` is retained as a golden file rather than as the dataset

### Branch

`feat/25-acceptance-against-generated-output`

### PR Convention

```
Closes #25
Part of #21

[...]

PR #32, opened 2026-08-20
feat(normalize): build the recipe graph from the reality tables

### LINKED ISSUE #23: Recipe graph and identity

jonstump:
## What

Implements #23 — the `items` and `recipes` half of the Tier 1 artifact, built from the product, substance and refiner tables with names resolved through the localisation join.

Against a real install this produces **2,237 items and 2,133 recipes** that assemble and validate as a single artifact.

## Decisions the source forced

| Question | Answer, and how it was established |
|---|---|
| Recipe ids for craft routes | The product's own ID. `AltRequirements` is present on every product row and **empty on all 2,144**, so there is exactly one craft route per product. A populated one now fails rather than silently colliding two routes onto one id. |
| Craft yields | `DefaultCraftAmount`. 2,143 products craft one at a time; exactly one (`AMMO`) crafts 25 — precisely the error a default would hide. |
| Refine/cook yields | `Result/Amount`. Up to 250. |
| Cook vs refine | The per-recipe `Cooking` flag, read rather than inferred. 1,323 of 1,681 refiner rows are cooking. |
| Self-referential recipes | Excluded and counted: **27**, seven of them producing `CATALYST2`. The count lands in `provenance.self_referential_recipes_omitted`, not omitempty, because a zero is a finding rather than an absence. |
| Alternatives | All emitted, none selected. 261 of 403 output/method pairs have more than one route, up to 61. |

## Verified numbers

Everything SPEC-0004 and ADR-0005 quote, checked against the real tables: 27 self-referential, 156 yields other than one, max yield 250, 403 output/method pairs with 261 carrying alternatives, 26 `CATALYST2` refine recipes defined (19 after exclusion), 18 unnamed products, `ULTRAPROD2` → "Stasis Device", and a 34-item craft closure matching the engine's hand-authored Stasis fixture.

## Tests

Fixtures under `internal/normalize/testdata/graph/` are **whole rows sliced from a real 6.45.0.1 decompilation** — `testdata/graph/gen.go` records which rows and why. Four seeds, each for a property that has to be exercised against real data: `ULTRAPROD2` (the 34-item craft closure), `CATALYST2` (26 refine routes, seven self-referential, and the `1x Crystal Sulphide → 50x Sodium Nitrate` yield), `FOOD_P_STELLAR` (so the `Cooking` flag has both values), `AMMO` (the one product that crafts 25).

A pass over a **whole decompiled install** is opt-in via `NMS_SOURCE_DIR`, following the pattern `internal/hgpak` already uses for `NMS_PCBANKS`:

```bash
NMS_SOURCE_DIR=/path/to/decompiled go test ./internal/normalize/ -run RealInstall -v
```

A fixture chosen by someone who already knew what the parser expects cannot show that the rest of the table holds no shape this code has never seen. That is the failure this project has recorded three times, and it is what the opt-in test exists to close. It found the finding below on its first run.

## Finding: the generated artifact does not resolve under default methods

**571 of 2,237 items hit a cycle**, all of them refine loops between gatherable substances — `CAVE1 ⇄ CAVE2`, `CATALYST1 ⇄ CATALYST2`, `GAS1 → GAS3 → GAS2 → GAS1`.

The cause is a gap in SPEC-0004, not a bug here. The spec derives raw-obtainability from the *absence* of a recipe, so a substance you gather with a mining beam **and** can refine comes out defaulting to `refine`. Refining runs both ways between several of those pairs.

The engine is right to refuse — SPEC-0001 REQ "Cycle Detection" calls this a runtime condition rather than one to assume away. What is missing is a source of truth for gatherability, and the substance table has one: **`PinObjective`**, which reads `UI_REFINE_OBJ` for the six substances that are refined only, and some flavour of gather/find/process for the other 105.

I measured it rather than assuming: marking raw-obtainability from `PinObjective` resolves **all 2,237 items with zero cycles**.

That is a modelling decision SPEC-0004 does not make, so it is **recorded, not invented in the normalizer**. The opt-in test asserts the cycle today with a comment saying why; when the spec is amended, that assertion changes with it. Suggested follow-up: amend SPEC-0004 REQ "Recipe Graph Construction" to derive `raw_obtainable` from `PinObjective`, and a story to implement it before #25's acceptance run.

Closes #23
Part of #21

Implements SPEC-0004 REQ "Recipe Graph Construction", REQ "Identifier Policy", REQ "Display Name Resolution", REQ "Excluded Content".

### DIFF
--- a/internal/domain/tier1.go
+++ b/internal/domain/tier1.go
@@ -121,6 +121,19 @@ type Provenance struct {
 	Archives []string `json:"archives"`
 	// MBINCompiler is the decompiler version that produced the .MXML input.
 	MBINCompiler string `json:"mbincompiler"`
+
+	// SelfReferentialRecipesOmitted counts the recipes dropped for naming
+	// their own output among their ingredients — a doubling strategy rather
+	// than a production path, and a cycle if expanded. There are 27 in NMS
+	// 5.97.
+	//
+	// Not omitempty: a zero here is a finding, not an absence, and a silent
+	// change in this number is exactly what recording it prevents.
+	//
+	// Governing: SPEC-0004 REQ "Recipe Graph Construction" — "MUST record
+	// how many it omitted in the artifact's provenance, so a change in that
+	// number is visible rather than silent."
+	SelfReferentialRecipesOmitted int `json:"self_referential_recipes_omitted"`
 }
 
 // Tier1 is the extracted recipe graph.
--- a/internal/normalize/artifact.go
+++ b/internal/normalize/artifact.go
@@ -28,6 +28,8 @@ type Builder struct {
 	items   []domain.Item
 	recipes []domain.Recipe
 	economy *domain.Economy
+
+	selfReferentialOmitted int
 }
 
 // NewBuilder starts an artifact for the named game build.
@@ -63,6 +65,12 @@ func (b *Builder) AddRecipes(recipes ...domain.Recipe) { b.recipes = append(b.re
 // SetEconomy attaches the base-economy section.
 func (b *Builder) SetEconomy(e *domain.Economy) { b.economy = e }
 
+// SetSelfReferentialOmitted records how many recipes the graph pass dropped
+// for naming their own output among their ingredients.
+//
+// Governing: SPEC-0004 REQ "Recipe Graph Construction"
+func (b *Builder) SetSelfReferentialOmitted(n int) { b.selfReferentialOmitted = n }
+
 // Artifact assembles and validates the artifact.
 //
 // Every collection is sorted into a defined order here rather than at the
@@ -76,8 +84,9 @@ func (b *Builder) Artifact() (*domain.Tier1, error) {
 		Source:        fmt.Sprintf("%s via MBINCompiler %s", joinArchives(b.archives), b.mbinCompiler),
 		Note:          b.note,
 		Provenance: &domain.Provenance{
-			Archives:     sortedCopy(b.archives),
-			MBINCompiler: b.mbinCompiler,
+			Archives:                      sortedCopy(b.archives),
+			MBINCompiler:                  b.mbinCompiler,
+			SelfReferentialRecipesOmitted: b.selfReferentialOmitted,
 		},
 		Items:   append([]domain.Item(nil), b.items...),
 		Recipes: append([]domain.Recipe(nil), b.recipes...),
--- a/internal/normalize/localisation.go
+++ b/internal/normalize/localisation.go
@@ -1,0 +1,104 @@
+package normalize
+
+import (
+	"path/filepath"
+	"sort"
+)
+
+// Governing: ADR-0001 (two-tier NMS data ingestion), SPEC-0004 REQ "Display
+// Name Resolution"
+//
+// Reality tables carry localisation *keys*, not names. The Stasis Device's
+// NameLower is UI_ULTRAPROD_2_NAME_L; the string "Stasis Device" lives in
+// language/nms_loc*_english.mbin, inside a different archive from the
+// product table. Searching the product table for a display name therefore
+// finds nothing, which is a dead end worth failing loudly about rather than
+// papering over.
+
+// Localisation resolves localisation keys to English strings.
+type Localisation struct {
+	byKey map[string]string
+	// files records what was read, so a failure to resolve can say where we
+	// looked rather than only what was missing.
+	files []string
+}
+
+// LoadLocalisation reads every localisation table matching pattern and
+// indexes its entries by key.
+//
+// Later files do not override earlier ones: the tables partition the key
+// space rather than layering, so a duplicate key across two files means the
+// input set is wrong (both a base and a patch table, say) and is reported.
+func LoadLocalisation(paths []string) (*Localisation, error) {
+	if len(paths) == 0 {
+		return nil, Missing("language/nms_loc*_english")
+	}
+	l := &Localisation{byKey: make(map[string]string, 64_000)}
+	for _, p := range sortedStrings(paths) {
+		doc, err := readMXML(p, "cTkLocalisationTable")
+		if err != nil {
+			return nil, err
+		}
+		name := filepath.Base(p)
+		l.files = append(l.files, name)
+		rows, err := doc.rows(name, "Table", "TkLocalisationEntry")
+		if err != nil {
+			return nil, err
+		}
+		for _, r := range rows {
+			key, err := r.nonEmpty(name, r.ID, "Id")
+			if err != nil {
+				return nil, err
+			}
+			// A key present with an empty English value is a real state in
+			// these tables (placeholder entries), and is kept as such: the
+			// caller decides whether an empty name is acceptable, rather
+			// than this layer silently dropping the key and reporting it as
+			// missing later.
+			val, err := r.str(name, key, "English")
+			if err != nil {
+				return nil, err
+			}
+			if prev, dup := l.byKey[key]; dup && prev != val {
+				return nil, Unrecognized(name, key, "Id", "one definition per key", "a conflicting duplicate")
+			}
+			l.byKey[key] = val
+		}
+	}
+	return l, nil
+}
+
+// Len reports how many keys were indexed.
+func (l *Localisation) Len() int { return len(l.byKey) }
+
+// Files reports the localisation tables that were read, in sorted order.
+//
+// Governing: SPEC-0004 REQ "Search Boundaries Are Recorded" — a name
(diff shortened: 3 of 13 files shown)


jonstump (commented):
## Review: Approved

Spec-aware review against issue #23's acceptance criteria, SPEC-0004, and ADR-0005. Verdict: **approve**. (Comment review — GitHub blocks `APPROVE` on a self-authored PR.)

Conflict-marker gate clean, CI green on all four checks. Verified locally on `3d1acc0`: `go test -race ./...` passes across all packages.

No findings. I went looking specifically at the places this codebase has been bitten before, and each one is handled.

### Recipe ids satisfy the constraint #30 just introduced

#30 made `Validate` require recipe ids unique **within an (output, method) group**, which is a constraint this PR is the first real producer to have to meet — 261 of 403 pairs carry alternatives, one with 61 routes. `checkClosed` enforces exactly that key:

```go
key := r.Output + "\x00" + string(r.Method) + "\x00" + r.ID
```

matching the domain-side invariant rather than a looser or stricter one, and failing generation on collision rather than letting two routes become indistinguishable in saved plan state.

The `AltRequirements` guard is the same instinct applied forward: the craft-id scheme assumes one route per product, that assumption is checked against all 2,144 rows, and a game update that starts populating the field fails loudly instead of silently collapsing two routes onto one id. That is the correct shape for an assumption you cannot verify beyond one build.

### No map-iteration order reaches the output

Every `range` in `tables.go` is over a slice — `rows`, `items`, `recipes`, `reqs.children(...)`. The four maps (`byOutput`, `known`, `seen`, `omitted`) are lookup-only, read by key inside deterministic slice loops. Given that unstable ordering has been the finding in three of the last four reviews here, worth confirming rather than assuming.

### The finding is the best part of this PR

571 of 2,237 items hitting a cycle under default methods, traced to SPEC-0004 deriving raw-obtainability from the *absence* of a recipe, is a real result — and the handling is right in three separate ways:

- **Diagnosed rather than worked around.** The engine refusing is correct behaviour under SPEC-0001 REQ "Cycle Detection"; the gap is in the producer's model of gatherability.
- **Measured rather than argued.** "`PinObjective != UI_REFINE_OBJ` resolves all 2,237 with zero cycles" is a checkable claim, not a plausible story.
- **Recorded rather than invented.** The modelling decision goes to SPEC-0004 (#33) and a story (#34) instead of being quietly implemented in the normalizer. That is the discipline ADR-0001's PSARC episode exists to enforce, applied to the project's own spec.

The opt-in test asserting the cycle *today*, with a comment saying the assertion changes when the amendment lands, is the right way to leave it — the current behaviour is pinned, and the pin names its own expiry.

### The fixtures are real rows, and the closure test cross-validates

`testdata/graph/` slices whole rows from a real 6.45.0.1 decompilation with `gen.go` recording which rows and why, and the four seeds each buy a distinct property — `ULTRAPROD2` for the craft closure, `CATALYST2` for alternatives and the 50× yield, `FOOD_P_STELLAR` so `Cooking` has both values, `AMMO` as the single product that crafts 25.

`TestStasisDeviceCraftClosureMatchesTheEngineFixture` is the strongest test here: it checks the *generated* graph's 34-item closure against the independently hand-authored engine fixture. Two artifacts built from different sources by different means agreeing on the same 34 items is real evidence, not self-consistency — which is precisely the failure mode SPEC-0003 was written to prevent recurring.

### Acceptance criteria

| Criterion (issue #23) | Status |
|---|---|
| Items and recipes built from product, substance and refiner tables | Met — 2,237 items, 2,133 recipes |
| Names resolved through the localisation join | Met — `ULTRAPROD2` → "Stasis Device" asserted |
| Yields read, never defaulted | Met — `AMMO` yield 25 asserted; refine yields to 250 |
| Cook vs refine read from the source | Met — per-recipe `Cooking` flag, both values in fixtures |
| Self-referential recipes excluded and counted | Met — 27, recorded in provenance, not `omitempty` |
| Alternatives all emitted, none selected | Met |
| Graph is closed — no dangling endpoints | Met — `checkClosed`, plus `checkSkippedUnreferenced` |

### On the schema change

`Provenance` gains `self_referential_recipes_omitted` without a `SchemaVersion` bump. I checked this against REQ "Schema Extension and Load Compatibility" and it is correct as-is: the version-increment clause there is scoped to the base-economy extension specifically (satisfied by #26 at 1→2), while the general rule — *"the producer, the schema, and the loader MUST change together"* — is what applies here, and this PR changes all three in one commit. The committed fixtures carry no `provenance` block, so they still load.

Recording the count as non-`omitempty` because "a zero is a finding, not an absence" is the right call and worth the comment it has.

### Merge ordering

No file overlap with #33 — this touches `internal/normalize/artifact.go`, #33 touches `economy.go` and the spec. They are independent and can merge in either order.

PR #33, opened 2026-08-20
docs: amend SPEC-0004 for raw-obtainability and product crop rewards

jonstump:
## What

Two amendments to SPEC-0004, both for things the real tables do that the spec's wording did not cover. Both were found by running the normalizer against a whole install rather than against a fixture — which is what the opt-in `NMS_SOURCE_DIR` pass in #32 exists to do.

No behaviour change. One factual correction to a comment and an emitted note, explained below.

## 1. Raw-obtainability MUST be read from the source

REQ "Recipe Graph Construction" derived `raw_obtainable` from the *absence* of a recipe. That assumes gathering and refining are exclusive. They are not — you mine Cobalt with a terrain manipulator **and** you can refine it — so every gatherable substance came out non-raw the moment the game gave it a refine route, and defaulted to `refine`.

Refining runs both ways between several such pairs (`CAVE1 ⇄ CAVE2`, `CATALYST1 ⇄ CATALYST2`, `GAS1 → GAS3 → GAS2 → GAS1`), so **571 of 2,237 items hit a cycle under pure defaults**. The engine is right to refuse — SPEC-0001 REQ "Cycle Detection" treats cycles as a runtime condition rather than one to design away — so the fix belongs in the producer.

The substance table states the fact directly in **`PinObjective`**: six substances read `UI_REFINE_OBJ` and are refined only; the other 105 read some flavour of gather, find or process. Measured rather than assumed: setting `raw_obtainable` from `PinObjective != UI_REFINE_OBJ` resolves **all 2,237 items with zero cycles**.

The requirement now also says what to do at the edges, because they are where a rule like this goes wrong:

- A **substance** whose `PinObjective` is absent or holds a value the list does not cover → **fail**, do not guess.
- A **product** that has a recipe → not raw-obtainable; the product table carries no equivalent field.
- A **product with no recipe** → still raw-obtainable, by the existing rule. That is what gives the engine a leaf rather than an item it cannot terminate on.
- A raw-obtainable item → defaults to `raw` even where it also has recipes, with the engine's per-node override there for players who want the refine route.

Three scenarios added, including the one that would have caught this: resolving every item in a generated artifact under default methods reports no cycle.

## 2. Crop yields come in two reward shapes

REQ "Base Economy Data" named only `GcRewardSpecificSubstance`. **Four of the twelve** farmable plants yield a *product* instead, under `GcRewardSpecificProduct` with an identical `AmountMin`/`AmountMax` shape. Reading one form drops four crops and still produces an artifact that loads.

Separately: **the reward key is not the yielded item.** The table is keyed by the plant's interaction id and names the item inside the entry — `PLANT_BARREN` yields `PLANT_DUST`. The two agree for eleven of twelve crops, which is exactly the ratio that makes the wrong rule look right.

Two scenarios added for each.

## Also

- Design entries recording why gatherability is a fact about the item rather than about the recipe table, and why both reward shapes are read — including what neither generalisation proves from twelve cases.
- A new risk: `PinObjective` is an enum read from one build, so a value NMS 5.97 does not contain now fails generation. Deliberate, but it is a new way for a game update to stop the pipeline.
- Resolves the design's open question on refiner difficulty variants — the artifact carries both, which #24 already implements.

## The one code change

#24's comment and its emitted search note said *five* of twelve crops yield a product. It is **four** — creature pellets (`CREATURE1`) are a substance, and I miscounted when writing that PR. Nothing read or emitted changes, because both reward forms were already read; but the note ships **inside the artifact**, so the count is corrected here alongside the spec text it came from rather than left to drift.

## What this does not do

Neither amendment is implemented. `PinObjective` needs a story before #25's acceptance run — #32's opt-in test asserts the cycle today with a comment saying why, so that assertion changes when the implementation lands.

Part of #21

Amends SPEC-0004 REQ "Recipe Graph Construction", REQ "Base Economy Data".

### DIFF
--- a/docs/openspec/specs/tier1-normalizer/design.md
+++ b/docs/openspec/specs/tier1-normalizer/design.md
@@ -62,6 +62,20 @@ Four facts shape this design:
 
 **Rationale**: The artifact is committed. Nondeterministic ordering makes every regeneration an unreviewable diff, which means nobody reviews it, which means a real balance change lands unnoticed among reordering noise. Determinism is what keeps `git diff` meaningful as the review surface.
 
+### Refiner throughput carries both difficulty variants
+
+**Choice**: Emit the standard and Survival figures side by side rather than picking one.
+
+**Rationale**: This resolves what was an open question in this design. `gcgameplayglobals` states both (`RefinerSubsMadeInTime` 250 against 100 on Survival), the planner knows which save it is planning for and the normalizer does not, and carrying both costs four integers.
+
+### Crop yields are read from whichever reward shape the entry carries
+
+**Choice**: Read both `GcRewardSpecificSubstance` and `GcRewardSpecificProduct`, and take the yielded item from the reward's own `ID` rather than from the key the reward was looked up by.
+
+**Rationale**: Both were found by running the normalizer against the real reward table rather than by reading the requirement. Eight of twelve farmable plants yield a substance and four yield a product, so a spec naming only the substance form describes a normalizer that drops four crops and still produces a loadable artifact. Separately, the reward key and the yielded item agree for eleven of twelve crops — `PLANT_BARREN` yields `PLANT_DUST` — which is exactly the ratio that makes "the key is the item" look true.
+
+**Trade-off**: Neither generalisation is provable from twelve cases. Both are written as "read what the entry says" rather than as a rule about which plants use which shape, so a thirteenth crop in a future build needs no change here.
+
 ### Base-economy data as ranges, keyed by hotspot
 
 **Choice**: Preserve minimum and maximum where the source expresses a range, and attach class scaling to hotspot categories rather than to devices.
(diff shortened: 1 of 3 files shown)


jonstump (commented):
## Review: Approved

Spec-aware review against SPEC-0004 and ADR-0005. Verdict: **approve**. (Comment review — GitHub blocks `APPROVE` on a self-authored PR.)

Conflict-marker gate clean, CI green on all four checks.

Both amendments record things the real tables do that the spec's wording did not cover, and both were found by running against a whole install rather than a fixture — which is the mechanism SPEC-0003 established and this is the first time it has paid off on the *spec* rather than on the code.

### I verified the factual correction independently

The one code change corrects "five of twelve crops yield a product" to "four", removing creature pellets. That number ships **inside the artifact** in the emitted search note, so it is a claim the artifact makes about the game rather than a comment, and worth checking rather than taking.

Extracted `metadata/reality/tables/rewardtable.mbin` from `NMSARC.Precache.pak` on a real install, decompiled it with MBINCompiler 6.45.0.1, and enumerated the reward templates by plant key:

| Key | Template |
|---|---|
| `PLANT_PEARL` | Product |
| `PLANT_GRAV` | Product |
| `PLANT_SACVENOM` | Product |
| `PLANT_NIP` | Product |
| `CREATURE1` | **Substance** (every occurrence) |
| `PLANT_DUST`, `PLANT_LUSH`, `PLANT_SNOW`, `PLANT_HOT`, `PLANT_RADIO`, `PLANT_POOP`, `PLANT_BARREN`, … | Substance |

Four products, and creature pellets are a substance in every entry. The correction is right, and the four now named in the comment are exactly the four the table carries.

### Amendment 1 — raw-obtainability from `PinObjective`

The reasoning holds and is the important half of this PR. Deriving `raw_obtainable` from the absence of a recipe assumes gathering and refining are exclusive, and they are not — you mine Cobalt *and* can refine it. Since refining runs both ways between several such pairs, the old rule does not merely mislabel items, it makes 571 of them unresolvable.

Locating the source of truth in `PinObjective` and stating the measurement — six substances read `UI_REFINE_OBJ`, the other 105 read gather/find/process, and the change resolves all 2,237 with zero cycles — is the right standard of evidence for a modelling decision.

The edge cases are where a rule like this goes wrong, and the requirement now names all four: absent-or-unknown `PinObjective` fails rather than guesses; a product with a recipe is not raw; a product **without** one still is, which is what leaves the engine a terminable leaf; and a raw-obtainable item defaults to `raw` with the per-node override still available. That last one matters — without it the amendment would take the refine route away from players who want it.

The new risk entry is the honest counterpart: `PinObjective` is an enum read from one build, so this is a new way for a game update to stop the pipeline. Naming it as a deliberate trade rather than leaving it implicit is correct, and consistent with how the `0x20` storage flag was handled in SPEC-0003.

### Amendment 2 — two reward shapes, and the key is not the item

Both halves are the same class of error and worth having caught together. Reading only `GcRewardSpecificSubstance` drops four crops **and still produces an artifact that loads** — silent partial data, the failure mode this project keeps returning to.

The second half is subtler and the note about it is the useful part: the reward key is the plant's interaction id, not the yielded item, and the two agree for eleven of twelve. Eleven-of-twelve agreement is exactly the ratio that makes a wrong rule survive testing, and saying so in the design doc is worth more than the correction itself.

### Scope is right

The PR states plainly that neither amendment is implemented and that `PinObjective` needs a story before #25's acceptance run — #34 is queued for it. Amending the spec in a separate PR from the code that revealed the gap is the correct split given `docs/` is a protected path, and it keeps the reasoning reviewable on its own.

Resolving the design's open question on refiner difficulty variants — the artifact carries both, which #24 already implements — is good housekeeping to fold in here rather than leave stale.

### Merge ordering

No file overlap with #32 (`economy.go` and the spec here; `artifact.go` there). Independent; either order.

PR #36, opened 2026-08-20
feat(normalize): read raw-obtainability from PinObjective

### LINKED ISSUE #34: Raw-obtainability from PinObjective

**Blocked by #23** — modifies `parseSubstances` and `resolveMethods` in `internal/normalize/tables.go`, which that story introduces.

**Requires the spec amendment in #33**, which establishes the rule this story implements.

The generated artifact does not resolve today. SPEC-0004 originally derived `raw_obtainable` from the *absence* of a recipe, which assumes gathering and refining are exclusive. They are not — you mine Cobalt with a terrain manipulator and you can refine it — so every gatherable substance with a refine route comes out defaulting to `refine`. Refining runs both ways between several such pairs (`CAVE1 ⇄ CAVE2`, `CATALYST1 ⇄ CATALYST2`, `GAS1 → GAS3 → GAS2 → GAS1`), and **571 of 2,237 items hit a cycle** under pure defaults.

The substance table states gatherability directly in `PinObjective`. Measured against NMS 5.97: setting `raw_obtainable` from `PinObjective != UI_REFINE_OBJ` resolves all 2,237 items with zero cycles.

This unblocks #25, which cannot run the engine over generated output while most of it cycles.

## Requirements

- [ ] **Recipe Graph Construction** (SPEC-0004) — Raw-obtainability MUST be read from the source, not inferred from the absence of a recipe. The normalizer MUST set `raw_obtainable` true for every substance whose `PinObjective` is not `UI_REFINE_OBJ`, and MUST fail rather than guess where the field is absent or holds a value outside the set observed. The product table carries no equivalent field, so a product that has a recipe MUST NOT be marked raw-obtainable; a product with no recipe still is. A raw-obtainable item MUST default to `raw` even where it also has recipes.
  - WHEN a substance's `PinObjective` is `UI_GATHER_REFINE_OBJ` and the table also defines refine recipes for it THEN it is emitted `raw_obtainable` true with `default_method` `raw`, and its refine recipes are emitted alongside
  - WHEN a substance's `PinObjective` is `UI_REFINE_OBJ` THEN it is emitted `raw_obtainable` false, defaulting to `refine`
  - WHEN the rollup engine resolves each item in a generated artifact under default methods THEN every item resolves, and none reports a cycle

## Scope

**In:**
- Read `PinObjective` in `parseSubstances`; fail closed on an absent or unrecognized value
- `resolveMethods` yields `raw` for a raw-obtainable item even where it has recipes
- Extend the graph fixtures so a `UI_REFINE_OBJ` substance and a `UI_GATHER_REFINE_OBJ` substance are both covered — `CAVE1`/`CAVE2` are the readable pair and both are already in the closure
- Update the opt-in `NMS_SOURCE_DIR` test: it currently asserts `ErrCycleDetected` with a comment saying the assertion changes when this lands

**Out:**
- Products. There is no gatherability signal for them and inventing one is the mistake this story exists to undo
- Any per-item override or curated list. If the enum grows, generation fails and that is the intended failure

## Acceptance Criteria

- `CAVE1` (Cobalt) emits `raw_obtainable` true, `default_method` `raw`, and still carries its refine recipes
- `CAVE2` (Ionised Cobalt) emits `raw_obtainable` false, `default_method` `refine`
- A substance whose `PinObjective` holds an unrecognized value fails generation naming the substance and the value
- A substance with no `PinObjective` field fails generation rather than defaulting either way
- A product with no recipe is still emitted `raw_obtainable` true with `default_method` `raw`
- A product that has a recipe is not marked raw-obtainable
- Over a real install (`NMS_SOURCE_DIR`), all 2,237 items resolve under default methods and none reports a cycle
- The six `UI_REFINE_OBJ` substances are the only substances emitted non-raw

### Branch

`feat/34-raw-obtainability-from-pinobjective`

### PR Convention

```
Closes #34
Part of #21

Implements SPEC-0004 REQ "Recipe Graph Construction" (as amended for
PinObjective).
```

jonstump:
## What

Implements #34. Raw-obtainability now comes from the substance table's `PinObjective` instead of being inferred from the absence of a recipe.

**Over a real install, all 2,237 items now resolve under default methods with zero cycles** — 571 of them could not before.

## The rule

All 111 substances carry one of five `PinObjective` values. The map is **closed**: a sixth value fails generation rather than being classified by whichever default happened to be written here.

| Value | Count | Raw-obtainable |
|---|---:|---|
| `UI_FIND_OBJ` | 42 | yes |
| `UI_GATHER_OBJ` | 41 | yes |
| `UI_GATHER_REFINE_OBJ` | 18 | yes |
| `UI_PROCESS_OBJ` | 4 | yes — the gases, from an atmosphere harvester |
| `UI_REFINE_OBJ` | 6 | **no** |

A raw-obtainable item defaults to `raw` even where it has recipes, and keeps them as legal alternatives — `CAVE1` still reports both `raw` and `refine` as legal methods, so the per-node override does its job.

## Two things found while implementing

Both are recorded rather than smoothed over.

**1. The source contradicts itself once.** `WATERPLANT` (Cyto-Phosphate) reads `UI_REFINE_OBJ` while **no refiner recipe produces it**. It has to be emitted raw or the engine has an item it cannot terminate on — but silently letting the "no recipe → raw" fallback overwrite a source-stated fact is how you lose track of a data change. It is emitted raw and reported in `Graph.RawByNecessity`, and the opt-in test asserts that list is exactly `[WATERPLANT]`.

**2. I was wrong about the product table.** SPEC-0004 (as I wrote it in #33) says "The product table carries no equivalent field". It does — all 2,144 product rows carry `PinObjective`. I checked the substance table's shape and inferred the product table's without grepping it, which is the same bounded-search-reported-as-general-result this project has now recorded four times.

The rule itself is unaffected: the field is not a gatherability signal for products. 1,240 are empty, and the vocabulary is open-ended — `UI_BUY_OBJ` (288), `UI_CRAFT_OBJ` (112), and one-off mission strings like `UI_PIN_VENTGEM_OBJ`. Products still take raw-obtainability from the graph. Only the spec's *reason* is wrong, and it is deferred below.

## Tests

Fixture-level, against real sliced rows:

- `CAVE1` (Cobalt) — raw-obtainable, defaults `raw`, **still carries its refine recipes** and reports both methods legal
- `CAVE2` (Ionised Cobalt) — not raw, defaults `refine`
- Every item in the generated artifact resolves with no cycle; resolving `CAVE2` reaches `CAVE1` and **terminates there**, which is the loop that used to close
- An unrecognized `PinObjective` fails, naming the value and the field
- An absent `PinObjective` fails, naming the substance
- A product with no recipe is a raw terminal; a product with one is not

`CAVE1`/`CAVE2` were added as fixture seeds so the gatherable-and-refinable case is exercised against real rows rather than assumed — before that, `CAVE1` had no recipes in the fixture and the test passed for the wrong reason. `testdata/graph/gen.go` records why.

Opt-in whole-install pass (`NMS_SOURCE_DIR`) now asserts the systemic property: every one of the 2,237 items resolves, no cycles, and the five refine-only substances are the only substances emitted non-raw — checked in both directions so a sixth joining them is caught.

- `docs/openspec/specs/tier1-normalizer/spec.md`, REQ "Recipe Graph Construction": "The product table carries no equivalent field" is false — it carries `PinObjective` on all 2,144 rows. Restate the reason: the field exists but is not a gatherability signal for products (1,240 empty, open-ended vocabulary including `UI_BUY_OBJ` and per-item mission strings). The requirement itself does not change.
- Same requirement: add the contradiction case — a substance marked `UI_REFINE_OBJ` that no recipe produces MUST still be emitted raw-obtainable, and the override MUST be recorded.

Closes #34
Part of #21

Implements SPEC-0004 REQ "Recipe Graph Construction" (as amended for PinObjective).

### DIFF
--- a/internal/normalize/tables.go
+++ b/internal/normalize/tables.go
@@ -42,6 +42,19 @@ type Graph struct {
 	// their name key, sorted. Every one is checked to be unreferenced.
 	UnnamedOmitted []string
 
+	// RawByNecessity lists substances the source marks refined-only that no
+	// recipe in fact produces, sorted. They are emitted raw-obtainable
+	// anyway, because an item with neither a recipe nor a gathering route is
+	// one the engine cannot terminate on.
+	//
+	// Recorded rather than silently overridden: it is the source
+	// contradicting itself, and one of them is `WATERPLANT` (Cyto-Phosphate),
+	// whose PinObjective reads UI_REFINE_OBJ while no refiner recipe makes
+	// it. A change in this list is a change in the game data.
+	//
+	// Governing: SPEC-0004 REQ "Search Boundaries Are Recorded"
+	RawByNecessity []string
+
 	// LocalisationFiles records which tables the name resolution read.
 	//
 	// Governing: SPEC-0004 REQ "Search Boundaries Are Recorded"
@@ -65,7 +78,7 @@ func BuildGraph(src Sources) (*Graph, error) {
 		return nil, err
 	}
 
-	subItems, err := parseSubstances(src.Substances, loc)
+	subItems, refinedOnly, err := parseSubstances(src.Substances, loc)
 	if err != nil {
 		return nil, err
 	}
@@ -81,7 +94,7 @@ func BuildGraph(src Sources) (*Graph, error) {
 	items := append(subItems, prodItems...)
 	recipes := append(craft, refined...)
 
-	resolveMethods(items, recipes)
+	rawByNecessity := resolveMethods(items, recipes, refinedOnly)
 
 	// Omitting an unnamed item is only safe while nothing references it.
 	// Asserted rather than assumed: a game update could start using one, and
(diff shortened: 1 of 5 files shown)


jonstump (commented):
## Review: Approved, with two notes

Spec-aware review against issue #34's acceptance criteria and SPEC-0004 as amended by #37. Verdict: **approve**. (Comment review — GitHub blocks `APPROVE` on a self-authored PR.)

Conflict-marker gate clean, CI green. Verified locally on `a8b6165`: `go test -race ./...` passes across all packages.

No correctness findings. The two notes below are about durability and fixture semantics, not defects.

### Verified against the real tables

I checked the closed-vocabulary claim rather than trusting the comment, since the whole design rests on it — extracted and decompiled the substance table from a real install:

| Value | Count |
|---|---|
| `UI_FIND_OBJ` | 42 |
| `UI_GATHER_OBJ` | 41 |
| `UI_GATHER_REFINE_OBJ` | 18 |
| `UI_PROCESS_OBJ` | 4 |
| `UI_REFINE_OBJ` | 6 |
| **total / distinct** | **111 / 5** |

Exactly the five in `pinObjectives`, over exactly the 111 substances the comment claims. Closing the map and failing on a sixth value is the right call, and the comment explaining *why* it is closed — "a sixth value is a game update changing something this rule depends on" — is what stops someone adding a permissive default later.

### The logic is right, including the ordering

`resolveMethods` putting `case items[i].RawObtainable` **first** is the load-bearing line: a raw-obtainable item defaults to `raw` even when it also has recipes, which is what actually breaks the `CAVE1 ⇄ CAVE2` and `GAS1 → GAS3 → GAS2 → GAS1` loops. Reordering it below the craft/refine cases would silently restore the 571 cycles.

The `default:` branch handling `WATERPLANT` is also right: it has no recipes, so it falls through to raw-obtainable, and the `refinedOnly` check records that this overrode a source fact rather than merely applied the fallback.

### Note 1 — `RawByNecessity` has no route into the artifact, while its sibling does

`SelfReferentialOmitted` has `Builder.SetSelfReferentialOmitted` and lands in `provenance.self_referential_recipes_omitted`. `RawByNecessity` has no equivalent setter and no field on `Provenance` — it exists only on the in-memory `Graph`.

The amended requirement says the override "MUST record", without saying where, so this satisfies the text as written: `BuildGraph` returns the `Graph` and a caller can read it. But the requirement's stated *reason* is "a change in that set is a change in the game data" — and that only holds if the set is somewhere a change is visible. Today the only production path is `BuildGraph`'s return value, and the only consumers are tests.

The concrete risk is #25, which is the story that wires generation for real: it will naturally call the setter that exists and has no prompt to carry a field that has none. Adding `SetRawByNecessity` now — even unused until #25 — would make the two records symmetric and make the omission impossible to miss. Reasonable to defer to #25 instead; worth being a decision rather than an oversight.

### Note 2 — the fixture's `RawByNecessity` is mostly fixture-slicing, not source contradiction

Measured against the committed graph fixture:

```
RawByNecessity = [LAND3 STELLAR2 WATER2 WATERPLANT]   (4 entries)
```

Against the real tables, only `WATERPLANT` is genuinely refine-only-with-no-recipe. `LAND3`, `STELLAR2` and `WATER2` are all produced by refiner recipes in the full table — they land in the fixture's list only because those recipes are outside the slice.

Nothing is wrong here: the mechanism is exercised, the real behaviour is correct, and the opt-in test asserts the true `["WATERPLANT"]`. But it is worth knowing before someone writes the obvious fixture assertion, because pinning `[LAND3 STELLAR2 WATER2 WATERPLANT]` would encode fixture shape as though it were game truth — and the field's whole meaning is "the source contradicts itself", which is false of three of those four.

If the fixture is ever meant to cover this scenario in CI, it needs either the missing recipes sliced in so only `WATERPLANT` remains, or an assertion that deliberately asserts containment rather than equality with a comment saying why.

Related: the only assertion on `RawByNecessity` today lives in `tables_real_test.go`, which skips without `NMS_SOURCE_DIR`. So the new spec scenario — "a refine-only substance no recipe produces is still raw, and the override is recorded" — has no CI coverage. That is defensible given the fixture caveat above, but it is the one scenario in this area that CI does not check.

### Acceptance criteria

| Criterion (issue #34) | Status |
|---|---|
| `CAVE1` raw + `default_method` raw, refine recipes still emitted | Met |
| `CAVE2` non-raw, defaulting to refine | Met |
| Unrecognized `PinObjective` fails naming substance and value | Met — `TestUnrecognizedPinObjectiveFails` |
| Absent `PinObjective` fails rather than defaulting | Met — `TestAbsentPinObjectiveFails` |
| Product with no recipe still raw | Met — `TestProductRawObtainabilityComesFromTheGraph` |
| Product with a recipe not marked raw | Met — same test |
| All 2,237 items resolve, zero cycles, over a real install | Met — `TestGeneratedGraphResolvesUnderDefaults` (opt-in) |
| The six `UI_REFINE_OBJ` are the only non-raw substances | Met, with the `WATERPLANT` override now spec'd by #37 |

The opt-in test being updated to assert resolution where it previously asserted `ErrCycleDetected` — the flip #32 left pinned with a comment naming this exact change — closes that loop properly.

PR #37, opened 2026-08-20
docs: correct SPEC-0004 on the product table, and record the one contradiction

jonstump:
## What

The two deferred design-doc updates from [#36](https://github.com/jonstump/nms-base-planner/pull/36). Both were found by implementing the `PinObjective` requirement rather than by re-reading it. No code changes — #36 already implements all of this.

## 1. The product table does carry `PinObjective`

The requirement said "The product table carries no equivalent field." It does, on **all 2,144 rows**.

That claim came from reading the substance table's shape and inferring the product table's rather than looking. It is the fourth recorded instance in this project of a bounded search reported as a general result — and it happened in the same change that added a requirement to read facts from the source instead of inferring them.

**The rule survives the correction on its merits rather than on the false premise.** The field exists but classifies nothing: 1,240 values are empty, and the vocabulary is open-ended — `UI_BUY_OBJ` (288), `UI_CRAFT_OBJ` (112), and per-item mission strings like `UI_PIN_VENTGEM_OBJ`. The requirement now says the normalizer MUST NOT read it, and says why.

## 2. One substance contradicts itself, and the override must be recorded

Six substances read `UI_REFINE_OBJ`. One of them — `WATERPLANT`, Cyto-Phosphate — **is produced by no recipe in the table**. The requirement had no answer for that.

Honouring the flag would leave an item with neither a recipe nor a gathering route, which the engine cannot terminate on. So the graph wins and it is emitted raw-obtainable.

What the requirement now adds is that the override **MUST be recorded**. The "no recipe, therefore raw" fallback would have swallowed this without a trace, and the next reader would have no way to tell a rule being applied from a rule being contradicted. A silent fallback that happens to be right here is the same code that would be wrong somewhere else.

## Scenarios

- The refine-only scenario now states the precondition it always assumed — that a recipe producing it exists.
- **New**: an absent or unrecognized `PinObjective` fails naming the substance and the value, rather than classifying it either way.
- **New**: a refine-only substance no recipe produces is emitted raw, and the override is recorded.

## Design

- The raw-obtainability decision carries a dated correction recording the wrong claim, where it came from, and why the conclusion holds without it.
- A new decision entry on why the graph wins when two source facts disagree, and why recording the disagreement is the part that matters.

Part of #21

Amends SPEC-0004 REQ "Recipe Graph Construction".

### DIFF
--- a/docs/openspec/specs/tier1-normalizer/design.md
+++ b/docs/openspec/specs/tier1-normalizer/design.md
@@ -136,6 +136,18 @@ Measured rather than assumed: setting `raw_obtainable` from `PinObjective != UI_
 
 **Trade-off**: The rule is a small allow/deny list over an enum this project has seen exactly one build of, so a game update that adds a `PinObjective` value will fail generation rather than classify it. That is the intended direction — the alternative is a default that silently reclassifies items — but it does mean this is a place an update can break, and it is named in the risks below.
 
+**Correction, 2026-08-19**: the first version of this decision said the product table carries no `PinObjective`. It does — on all 2,144 rows. That claim came from reading the substance table's shape and inferring the product table's rather than looking, which is the fourth recorded instance in this project of a bounded search reported as a general result, and it happened in the same change that added a requirement to read facts from the source. The rule survives the correction on its merits rather than on the false premise: 1,240 of those values are empty and the vocabulary is open-ended, running from `UI_BUY_OBJ` and `UI_CRAFT_OBJ` to per-item mission strings like `UI_PIN_VENTGEM_OBJ`. It classifies nothing.
+
+### Where two source facts disagree, the graph wins and the disagreement is recorded
+
+**Choice**: A substance the source marks refined-only that no recipe produces is emitted raw-obtainable, and every such override is reported.
+
+**Rationale**: Found by implementing the rule above against the whole table. `WATERPLANT` (Cyto-Phosphate) reads `UI_REFINE_OBJ` while nothing in the recipe table makes it. One of the two facts has to give, and it has to be the flag: an item with neither a recipe nor a gathering route is one the engine cannot terminate on, so honouring the flag produces an artifact that does not resolve.
+
+What matters is that the override is not silent. The "no recipe, therefore raw" fallback would have swallowed this case without a trace, and the next reader would have had no way to tell a rule being applied from a rule being contradicted. Reporting it makes the disagreement a reviewable line in the output, and a change in that set a change in the game data.
+
+**Trade-off**: One item today, and a mechanism for it. Justified because the alternative is indistinguishable from the bug: a silent fallback that happens to produce the right answer here is the same code that would produce a wrong one somewhere else.
+
 ### Cooking is a flag, not a table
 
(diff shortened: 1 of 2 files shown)


jonstump (commented):
## Review: Approved

Spec amendment review against SPEC-0004 and the implementation in #36. Verdict: **approve**. (Comment review — GitHub blocks `APPROVE` on a self-authored PR.)

Conflict-marker gate clean, CI green.

**Both factual claims verified independently against the real tables**, rather than taken from the PR body — extracted `nms_reality_gcproducttable`, `nms_reality_gcsubstancetable` and `nms_reality_gcrecipetable` from `NMSARC.Precache.pak` and decompiled them with MBINCompiler 6.45.0.1.

### 1. The product table does carry `PinObjective` — confirmed, exactly

| Measured | PR claims |
|---|---|
| 2,144 product rows | 2,144 |
| 2,144 carrying a `PinObjective` field | all rows |
| 1,240 empty | 1,240 |
| `UI_BUY_OBJ` 288 | 288 |
| `UI_CRAFT_OBJ` 112 | 112 |
| 17 distinct non-empty values | "open-ended", per-item mission strings |

Every number matches. The prior claim — "the product table carries no equivalent field" — was wrong, and the correction is right.

The important part is the second-order handling: the rule **survives on its merits, not on the false premise**. The field exists but classifies nothing (58% empty, an open vocabulary running to per-item mission strings), so "MUST NOT read it" is the correct conclusion for a different reason than the one originally given. Rewriting the requirement to say why, rather than quietly leaving a right answer resting on a wrong fact, is the difference between a correction and a patch.

Naming it as the fourth recorded instance of a bounded search reported as a general result — in the very change that added a requirement to read facts from the source instead of inferring them — is worth the sting. That is the pattern this project keeps catching, and it caught itself this time.

**For the record: I reviewed and approved #33 with that claim in it and did not check the product table.** The claim was checkable in about two minutes with tooling this repo already ships. Worth noting so the count of "who verified what" stays honest.

### 2. `WATERPLANT` contradicts itself — confirmed

Six substances read `UI_REFINE_OBJ`: `LAUNCHSUB2`, `LAND3`, `CAVE2`, `WATER2`, `WATERPLANT`, `STELLAR2`. Cross-checked each against the 403 distinct outputs in the refiner table:

| Substance | Produced by a recipe? |
|---|---|
| LAUNCHSUB2, LAND3, CAVE2, WATER2, STELLAR2 | yes |
| **WATERPLANT** | **no** |

Exactly one contradiction, exactly the one described.

The resolution is right — an item with neither a recipe nor a gathering route is one the engine cannot terminate on, so the graph has to win. What makes the amendment worth its length is the second half: requiring the override to be **recorded**. The pre-existing "no recipe, therefore raw" fallback would have produced the same output silently, and the reasoning in the PR body is the correct one — *"a silent fallback that happens to be right here is the same code that would be wrong somewhere else."*

Adding the precondition to the existing refine-only scenario ("and the table defines a recipe producing it") is the small fix that matters: that scenario was always assuming it and now says so.

### Design entries

Both are the right things to record — the dated correction naming the wrong claim and where it came from, and a decision entry on why the graph wins when two source facts disagree. The second generalises past this one case, which is what makes it a decision rather than a note.

PR #46, opened 2026-08-20
feat: acceptance against generated output, and the generated dataset

### LINKED ISSUE #25: Acceptance against generated output

jonstump:
## What

Implements #25 — the last story in epic #21. Adds `cmd/nmstier1`, commits what it produces as `data/tier1.json`, and resolves the Stasis Device through the merged rollup engine from **that** artifact rather than from the hand-authored fixture.

Resolving the fixture verifies the engine, which #2 already did. Only an input the normalizer produced in the same run verifies the normalizer.

## The dataset

`data/tier1.json` — 2,237 items, 2,133 recipes, 108 parts, 3 hotspot families, 12 crops, refiner throughput. Generated in one step:

```bash
go run ./cmd/nmstier1 -src $SRC -out data/tier1.json -game-version 5.97
```

`-game-version` is required rather than defaulted: SPEC-0004 forbids emitting a guessed version, and nothing in the decompiled tree states it.

## Tests

| Test | Gate | What it proves |
|---|---|---|
| `TestCommittedArtifactReproducesTheGoldenTree` | always | The committed dataset resolves to ADR-0001's confirmed tree |
| `TestGeneratedTreeScalesLinearly` | always | ×10 is exactly ten times, over generated data |
| `TestGeneratedTreeIsThirtySixNodesNotThirtyFour` | always | The finding below, pinned |
| `TestRealInstallReproducesTheCommittedArtifact` | `NMS_SOURCE_DIR` | Regenerating from a real install produces the committed file **byte for byte** |

That last one is what SPEC-0004's "reproduce any committed artifact from the same install without manual steps" actually asserts, and it is the reason the always-on tests are meaningful rather than circular: they run against a file the gated test proves is the normalizer's own output. It passes.

## ADR-0001's quantities all hold

- Each gas product costs **250 gas and 50 Condensed Carbon** — `REACTION1`/`2`/`3` verified individually
- At ×1: **500 Sulphurine, 500 Nitrogen, 500 Radon, 300 Condensed Carbon**, each an exact integer
- Three branches: Quantum Processor (`MEGAPROD2`), Cryogenic Chamber (`MEGAPROD3`), Iridesite (`ALLOY8`), one of each

## Finding: the tree is 36 nodes, not 34

The two extra are `CAVE1` (Cobalt) and `OXYGEN`.

`CAVE2` (Ionised Cobalt) is **refined, not gathered** — its `PinObjective` reads `UI_REFINE_OBJ`, verified in #34 — so it expands into 2 Cobalt and 2 Oxygen instead of terminating. The hand-authored fixture models it as a raw leaf. That was a prototype simplification, not a fact about the game, and ADR-0001's 34 was counted from the prototype tree.

It is not a default a plan could opt out of: pinning `CAVE2` to `raw` on real data is rejected, because the artifact says it is not raw-obtainable. The test asserts that too, so "structural" is checked rather than claimed.

Asserted as its own named test rather than folded into the node count, so the difference stays a recorded finding instead of a number someone quietly changed.

- `docs/adrs/ADR-0001-two-tier-nms-data-ingestion.md`: the confirmed Stasis tree is 36 nodes, not 34. `CAVE2` is refined rather than gathered, so it expands into 2 Cobalt and 2 Oxygen. Every quantity in the confirmation holds unchanged.
- `docs/openspec/specs/tier1-normalizer/spec.md`, REQ "Acceptance Against the Golden Tree": same correction to the "34 distinct nodes" bullet and to the scenario's expected count.

## Also

- The hand-authored fixture is retained as a golden file. Its `note` now says so, points at `data/tier1.json` as the real dataset, and names the one place it differs from the real tree — it previously said "Replace via the extraction spike", which is done.
- `data/README.md` records the regeneration steps and that the file is never hand-edited.
- The artifact's own `note` carries the omission counts: 27 self-referential recipes, 18 unnamed items, and the `WATERPLANT` raw-by-necessity override.

Closes #25
Part of #21

Implements SPEC-0004 REQ "Acceptance Against the Golden Tree".


jonstump (commented):
## Review: Changes requested

Spec-aware review against issue #25's acceptance criteria and SPEC-0004. (Comment review — GitHub blocks `REQUEST_CHANGES` on a self-authored PR. Treat as changes-requested.)

Conflict-marker gate clean, CI green on all four checks. Verified locally on `f10de34`: `go test -race ./...` passes; the three always-on acceptance tests pass and the two `NMS_SOURCE_DIR` tests skip cleanly.

One finding, and it lands on the requirement this story is closest to.

---

### The provenance under-reports the source archives

`cmd/nmstier1/main.go:93` hardcodes the archive list:

```go
b, err := normalize.NewBuilder(*version, *mbin, []string{
    "NMSARC.Precache.pak", "NMSARC.globals.pak",
```

Every display name in the artifact comes from the localisation join — `ULTRAPROD2` → "Stasis Device", `FUEL2` → "Condensed Carbon" — and those tables are in neither of those archives. Measured against the install:

| Archive | Entries matching `english` |
|---|---|
| `NMSARC.MetadataEtc.pak` | **16** (`language/nms_loc1_english.mbin` and siblings) |
| `NMSARC.globals.pak` | 0 |
| `NMSARC.Precache.pak` | 0 |

I found this trying to reproduce the artifact from the recorded provenance: I extracted `metadata/reality/tables` (54), `metadata/simulation/scanning` (11) and `interactiveflora` (164) from Precache and `gcgameplayglobals` (1) from globals, and there was no `language/` anywhere — which is what `nmstier1` globs for at `main.go:71`.

SPEC-0004 REQ "Source Provenance and Version Stamping" requires the artifact to "record the source archives". Naming two of the three is a provenance that cannot reproduce its own artifact — follow it exactly and you get a graph with no names.

That matters more here than it would elsewhere, because this is the story whose entire subject is proving the artifact is what it claims. `TestRealInstallReproducesTheCommittedArtifact` is the strongest assertion in the PR and the reason the always-on tests aren't circular — but it takes `NMS_SOURCE_DIR` as given, so it verifies the generator is deterministic without verifying that the recorded provenance describes the tree it consumed. The one input the gated test cannot check is the one that is wrong.

**Suggested fix**: add `NMSARC.MetadataEtc.pak`. Deriving the list from the files actually read would be better than a third hardcoded string — the same argument the epic already makes for reading yields rather than defaulting them — but a correct constant is a fine first step, and the ADR-0001 note about localisation living in a different archive from the product table is the comment that belongs next to it.

Worth checking whether the economy sources pull from a fourth archive while you are in there; I confirmed the four paths above and did not audit `ReadEconomy`'s full input set.

---

### The 36-node finding is right, and I verified it independently

Rather than trust the assertion, I loaded the committed `data/tier1.json` and resolved `ULTRAPROD2` directly:

```
NODE COUNT = 36
CAVE1 present: true   OXYGEN present: true
CAVE2 terminal=false recipe=REFINERECIPE_126
    -> CAVE1  perUnit=2 yield=5
    -> OXYGEN perUnit=2 yield=5
```

And every ADR-0001 quantity holds against generated data:

| Leaf | Total |
|---|---|
| Sulphurine / Nitrogen / Radon | 500 each, exact |
| Condensed Carbon | 300, exact |
| Cobalt / Oxygen (the two new nodes) | 60 each |

`REACTION1`/`2`/`3` each carry `GAS × 250` and `FUEL2 × 50` — the "250 gas + 50 Condensed Carbon" claim, per reaction.

So the correction is exactly as narrow as the PR says: the node count was counted from the prototype tree, where `CAVE2` was modelled as a raw leaf. It is not, its `PinObjective` says so, and #34 already established that reading. Every quantity survives.

Asserting it as `TestGeneratedTreeIsThirtySixNodesNotThirtyFour` rather than quietly changing a 34 to a 36 is the right call — the name makes the finding un-loseable, and a future reader who knows ADR-0001 says 34 will find the explanation instead of a contradiction.

The deferred design-doc updates are correctly scoped: ADR-0001's confirmation section and SPEC-0004's "34 distinct nodes" bullet both need the same correction, and `docs/` is protected from this branch.

### The RawByNecessity note from #36 is addressed

`GenerationNote` carries it into the artifact's `note` field, and the committed artifact shows it:

> …27 self-referential recipes omitted; 18 items omitted for having no English name. Substances the source marks refined-only that no recipe produces, emitted raw-obtainable so the graph terminates: WATERPLANT.

That closes the gap I raised on #36 — the override is now visible to a reader of the artifact rather than only to a caller holding the `Graph`. The comment's reasoning is the right one: "an omission nobody can see is indistinguishable from data that was never there."

### Acceptance criteria

| Criterion (issue #25) | Status |
|---|---|
| Generates an artifact and resolves `ULTRAPROD2` through `internal/domain` in the same run | Met — gated test regenerates and compares; always-on tests resolve the committed file |
| Exactly 34 distinct nodes across the three named branches | **Superseded** — 36, with the difference proven structural and ADR-0001 queued for correction |
| Leaf totals 500/500/500/300 | Met — verified independently |
| `REACTION1`/`2`/`3` each 250 gas + 50 Condensed Carbon | Met — verified independently |
| Gated on a real install, skips cleanly so CI stays green | Met — both gated tests skip; CI green |
| Generated artifact committed; `stasis-device.tier1.json` retained as golden file | Met — 914 KB, and the fixture's `note` now says what it is and where it differs |

### What I did not verify

I did not run `TestRealInstallReproducesTheCommittedArtifact`. Building `NMS_SOURCE_DIR` needs a decompiled tree spanning three archives — including the 16 localisation tables at ~5 MB each — and I stopped when the missing `language/` directory turned into the finding above. So the byte-for-byte reproduction claim is unverified by me; it is asserted by a test that passes on your machine, and the finding above is precisely that the claim's inputs are not fully recorded.

jonstump:
## Review feedback addressed — provenance fixed in 2acc66d

### The archive list

`NMSARC.MetadataEtc.pak` added, and the constant now carries why it is a constant and what each archive supplies:

```go
//	NMSARC.Precache.pak     metadata/reality/tables, simulation/scanning,
//	                        models/planets/.../interactiveflora
//	NMSARC.globals.pak      gcgameplayglobals.global
//	NMSARC.MetadataEtc.pak  language/nms_*_english — every display name
```

It stays a constant rather than becoming derived, and the comment says why: `-src` is a *decompiled* tree, so the `.MXML` files carry no record of which `.pak` produced them. The generator cannot observe this — it has to be stated, which means it has to be kept true by hand, which is the thing the comment exists to make possible.

### `data/README.md` had the same error, in a worse form

The documented procedure said:

```
nmsextract extract NMSARC.Precache.pak $SRC language/
```

Precache contains no `language/` path at all, so following the README literally produces an empty directory and the generator dies on the missing glob. Corrected to five commands across three archives, and **every one was run against a real install and its entry count checked** rather than reasoned about:

| Archive | Filter | Entries |
|---|---|---|
| `NMSARC.Precache.pak` | `metadata/reality/tables/` | 54 |
| `NMSARC.Precache.pak` | `metadata/simulation/scanning/` | 11 |
| `NMSARC.Precache.pak` | `plantinteraction.entity` | 13 |
| `NMSARC.globals.pak` | `gcgameplayglobals.global` | 1 |
| `NMSARC.MetadataEtc.pak` | `_english.mbin` | 8 |

The README also now says plainly that it is three archives and why the third is easy to miss.

### The artifact is regenerated, not hand-edited

The provenance lives inside `data/tier1.json`, so correcting it means regenerating. The diff is **three lines** — the `source` string and the added archive. Everything else is byte-identical, which is the determinism requirement doing its job on a change that touches only metadata.

### On the reproduction claim I said I had not verified

I have now. I built a decompiled tree from all three archives — 77 `.mbin` extracted and decompiled with MBINCompiler 6.45.0.1 — and ran the gated test.

**It passed *before* this change**, which is what confirmed my source tree matched yours, and it passes after. So `TestRealInstallReproducesTheCommittedArtifact` is doing exactly what the PR claims, and the committed file is still precisely what the generator emits.

That ordering was deliberate: proving the baseline reproduces first meant that when the regenerated file differed by three lines, the difference was attributable to the one thing I changed rather than to a mismatched source tree.

| Check | Result |
|---|---|
| `NMS_SOURCE_DIR=... go test -run TestRealInstall` | both gated tests pass |
| `go vet ./...` | clean |
| `gofmt -l .` | no output |

### Not changed

I did not audit `ReadEconomy`'s full input set beyond the four paths it names as constants — those four are covered by the three archives above, and all five README commands are verified, but if a future economy source is added from a fourth archive this list needs the same attention. That is the maintenance cost of a constant the generator cannot check for itself, and it is now written down next to the constant.

PR #48, opened 2026-08-20
fix(hgpak): name the magic found even when it is not printable

jonstump:
Closes the last outstanding finding from the SPEC-0003 reviews — first raised on #13, carried through eight PRs, and never picked up because it only ever lived in review comments.

## The gap

SPEC-0003 REQ "Container Identification" requires a mismatch to fail *"with an error naming the magic found"*. `printableMagic` returned only the leading run of printable ASCII, so a file whose first byte is binary was refused with:

```
magic is "", want "HGPAK": not an HGPAK archive
```

That names neither what was opened nor that anything was read at all.

**Why it survived so long is the interesting part.** The requirement's own scenario uses a PSARC file, and a PSARC header starts with four printable bytes. So the implementation satisfied the one case the spec names and failed every case it does not — and the test written from that scenario passed. A spec scenario is an example, not a bound, and this is what it looks like when an implementation fits the example rather than the rule.

## The fix

`describeMagic` always carries the eight bytes as hex, and keeps the printable prefix in front of them when there is one:

```
PSARC  magic is "PSAR" (5053415200010004), want "HGPAK": not an HGPAK archive
gzip   magic is 1f8b080000000000, want "HGPAK": not an HGPAK archive
zstd   magic is "(" (28b52ffd00000000), want "HGPAK": not an HGPAK archive
```

The prefix stays because recognisability is the point of the spec's scenario — a reader who sees `PSAR` knows immediately what they opened. The hex is unconditional because the magic is eight bytes and a printable prefix is at most part of it: `"PSAR"` is the recognisable half of a PSARC header, not the whole of it, so even the case that "worked" was naming half the magic.

## Tests

- `TestNonPrintableMagicIsStillNamed` — gzip, all-binary, and zstd-frame leads, each asserting the message is not `magic is ""` and carries the actual bytes
- `TestPrintableMagicIsReportedWithItsFullBytes` — PSARC keeps its recognisable prefix *and* gains the other four bytes

Both **fail against the old implementation**, reporting the literal `magic is ""`. The pre-existing `TestPSARCIsRejectedNamingTheMagic` is untouched and still passes.

## One cosmetic wart, left alone

A magic whose first byte happens to be printable but whose second is not yields a one-character prefix — zstd's `0x28` renders as `"("`. It is noise rather than signal, but suppressing short prefixes would mean inventing a minimum length the spec does not have, and the hex is authoritative in every case. Flagging it rather than papering over it with an arbitrary threshold.

## Verification

| Check | Result |
|---|---|
| `NMS_PCBANKS=... go test` | **97/97 archives** still open |
| `go vet ./...` | clean |
| `gofmt -l .` | no output |

The 97-archive run matters here: this changes the rejection path, and the acceptance path shares `Open`.

### DIFF
--- a/internal/hgpak/hgpak.go
+++ b/internal/hgpak/hgpak.go
@@ -209,8 +209,8 @@ func Open(r io.ReaderAt, size int64) (*Archive, error) {
 	// what was actually found — a PSARC archive must be diagnosable as such
 	// rather than as generic corruption.
 	if !bytes.Equal(hdr[:8], magic[:]) {
-		return nil, fmt.Errorf("magic is %q, want %q: %w",
-			printableMagic(hdr[:8]), "HGPAK", ErrNotHGPAK)
+		return nil, fmt.Errorf("magic is %s, want %q: %w",
+			describeMagic(hdr[:8]), "HGPAK", ErrNotHGPAK)
 	}
 
 	a := &Archive{r: r, size: size}
@@ -475,9 +475,34 @@ func (a *Archive) Close() error {
 	return nil
 }
 
-// printableMagic renders a magic value for an error message, keeping ASCII
-// readable so "PSAR" is recognizable at a glance.
-func printableMagic(b []byte) string {
+// describeMagic renders a magic value for an error message.
+//
+// Governing: SPEC-0003 REQ "Container Identification" — a mismatch MUST fail
+// "with an error naming the magic found", so this must never come back
+// empty.
+//
+// It previously returned only the leading run of printable ASCII, which
+// named nothing at all for a file whose first byte is binary: a gzip archive
+// was refused with `magic is ""`. The PSARC scenario the requirement gives
+// happens to start with four printable bytes, which is why the gap survived
+// — the one case the spec names was the one case that worked.
+//
+// The hex is always present because the magic is eight bytes and a printable
+// prefix is at most part of it: "PSAR" is the recognisable half of a PSARC
+// header, not the whole of it. The prefix stays in front because that
+// recognisability is the point — a reader who sees PSAR knows immediately
+// what they opened, and the requirement's own scenario is about exactly that
+// diagnosis.
+func describeMagic(b []byte) string {
+	if p := printablePrefix(b); p != "" {
+		return fmt.Sprintf("%q (%x)", p, b)
+	}
+	return fmt.Sprintf("%x", b)
+}
+
+// printablePrefix is the leading run of printable ASCII in b, which may be
+// empty.
+func printablePrefix(b []byte) string {
 	end := len(b)
 	for i, c := range b {
 		if c < 0x20 || c > 0x7e {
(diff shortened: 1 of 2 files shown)


jonstump (commented):
Approved on review — see the verification comment. (GitHub refuses self-approval on this account, so this is filed as a COMMENT review.)

jonstump:
## Review: approved — SPEC-0003 REQ "Container Identification"

(GitHub refuses self-approval on this account, so this is a comment rather than an APPROVE review.)

Verified independently rather than taken from the PR body.

**The claim that both tests fail against the old implementation holds.** I reverted `describeMagic` to the previous one-line behaviour and reran:

```
gzip        magic is "", want "HGPAK": not an HGPAK archive
all binary  magic is "", want "HGPAK": not an HGPAK archive
zstd frame  magic is "(", want "HGPAK"   — missing 28b52ffd
PSAR        magic is "PSAR", want "HGPAK" — missing 5053415200010004
```

All four subtests fail, with exactly the `magic is ""` symptom described. The tests are load-bearing, not decorative.

**Messages, checked end to end** against `hgpak.Open`:

| Input | Message |
|---|---|
| PSARC | `magic is "PSAR" (5053415200010004), want "HGPAK"` |
| gzip | `magic is 1f8b080000000000, want "HGPAK"` |
| zstd | `magic is "(" (28b52ffd00000000), want "HGPAK"` |
| all-ASCII | `magic is "ABCDEFGH" (4142434445464748), want "HGPAK"` |
| `HGPAKX` | `magic is "HGPAKX" (484750414b580000), want "HGPAK"` |

The near-miss case is an unintended win: `HGPAKX` against `HGPAK\0\0\0` is legible only because the hex shows where the nulls should be.

The requirement says reject "with an error naming the magic found". The old path named nothing whenever the first byte was binary, so this closes a real gap rather than polishing a message.

**Verification**

| Check | Result |
|---|---|
| Conflict-marker gate | Clean |
| CI (4 checks) | All pass |
| `go test ./...` | ok |
| `NMS_PCBANKS=… go test ./internal/hgpak/` | `TestFullArchivesFromRealInstall` and `TestFullArchivePathResolutionFromRealInstall` both pass |
| `go vet`, `staticcheck v0.8.0-rc.1`, `gofmt -l` | Clean |

The whole-install run matters because this changes the rejection path and `Open` is shared with the acceptance path — confirmed it still opens every archive.

**On the flagged wart** — zstd rendering as `"("` — agreed on leaving it. Suppressing short prefixes means inventing a minimum length the spec does not have, and the hex is authoritative in every row of that table. Flagging it beat picking a threshold.

### One follow-up, not blocking

The PR's own diagnosis is that the implementation fitted the scenario instead of the rule, because the scenario's PSARC header happens to start with four printable bytes. **SPEC-0003 still has only that one scenario.** The code is now correct and the tests bound it, but the spec is still bounded by the same single example that hid this for eight PRs.

Worth adding alongside "A PSARC file is rejected clearly":

> **WHEN** a file whose magic begins with a non-printable byte is opened
> **THEN** the read fails naming the magic found, rather than naming an empty string

A docs change, not a change to this PR.

Issue #41, opened 2026-08-20
Power computation: generation, draw and deficit


jonstump:
Part of #38. Governed by [SPEC-0001](docs/openspec/specs/rollup-engine/spec.md).

**Blocked by #39** — needs the base grouping and constants input. Independent of #40: power reads draws from the same per-base assignment rather than from producer rows, so the two can proceed in parallel once #39 lands.

Stage 3: per base, total generation against total draw, and what to build when draw wins.

## Settle the open question first

design.md's open questions say of mixed power grids:

> The base-planner handoff models one power type per base and notes production likely needs a generator list instead of a type toggle. This spec follows the mock (single type per base). **Revisit before implementing stage 3 in earnest.**

That revisit is a prerequisite for this story, not a detail inside it. Building single-type-per-base and later moving to a generator list changes the shape of the result type, which the WASM boundary (SPEC-0002) then has to carry. Resolve it — as an amendment or a recorded decision to stay with the mock — before implementation starts.

## Requirements

- [ ] **Power Computation** (SPEC-0001) — For each base the engine MUST compute total generation, total draw, and the resulting surplus or deficit, using constants supplied at call time. Generation MUST support two source types: electromagnetic generators MUST apply a class multiplier to a base output; solar panels MUST be classless and MUST additionally require batteries at a configured ratio for night coverage. When draw exceeds generation, the engine MUST report the deficit and MUST report the number of additional generation units required to clear it, so the view can present the fix as an action rather than a warning.
  - WHEN a base has 3 electromagnetic generators at class A, with a class-B base output of 110 kPs and a class-A multiplier of 1.5 THEN the engine reports 495 kPs of generation
  - WHEN a base is powered by 5 solar panels at a ratio of 1 battery per 2 panels THEN the engine reports 3 batteries as required
  - WHEN a base draws 400 kPs against 330 kPs of generation from class-B electromagnetic generators producing 110 kPs each THEN the engine reports a deficit of 70 kPs and reports that 1 additional generator clears it
  - WHEN a base at surplus has its generator class downgraded such that generation falls below draw THEN the engine reports a deficit and the required additional unit count

## Notes

**The deficit must carry its fix.** "Report the deficit" alone satisfies half the requirement — the point is the additional-unit count, so the view renders an action rather than a warning. The fourth scenario exists to prove the fix recomputes on a class change rather than being computed once.

[...]

Issue #40, opened 2026-08-20
Producer rollup: farm, extractor, ranch and kitchen


jonstump:
Part of #38. Governed by [SPEC-0001](docs/openspec/specs/rollup-engine/spec.md).

**Blocked by #39** — needs the base grouping and constants input that story establishes.

Stage 2 of the engine: turn each base's assigned leaf totals into producer requirements across four types — `farm`, `extractor`, `ranch`, `kitchen`. This is the story that makes the planner a base planner rather than a dependency tree.

The largest story in the epic by some margin — one requirement, ten scenarios, four producer types. It is not split because SPEC-0001 carries it as a single requirement and the four types share the sizing, per-base aggregation and byproduct logic; splitting them would put coupled code in separate PRs. Expect a PR above the usual 200–500 line target.

design.md points at `docs/design/base-planner/handoff.md` for worked examples, and its build order says to develop this stage against them.

## Requirements

- [ ] **Producer Rollup** (SPEC-0001) — For each base, the engine MUST convert assigned leaf totals into producer requirements by producer type: `farm`, `extractor`, `ranch`, and `kitchen`. Producer counts MUST derive from constants supplied at call time and MUST NOT be hardcoded. Farm rows MUST report plant count and biodome count. Extractor rows MUST report extractor count sized so the required quantity is produced within a configured fill duration at the site's configured class, together with the resulting fill time; above the configured depot threshold the row MUST additionally report `ceil(required quantity / depot capacity)` supply depots, and below it MUST report none. Ranch rows MUST report the fauna count required to yield the required quantity within one configured collection cycle, together with the cycle time; where any fauna product at a base requires feeding, the base MUST report a pellet feeder once per base rather than once per row. Kitchen rows MUST report each step's input-to-output ratio and process duration and MUST distinguish intermediate steps from the step producing the plan target; nutrient processor count MUST be reported once per base as `ceil(step count / steps per processor)` and MUST NOT be the sum of per-row counts. Extractor class MUST be configured per site, not per row. Items whose demand is satisfied by a byproduct of another producer at the same base MUST be reported as requiring no construction, contributing no producer count and no power draw.
  - WHEN a base is assigned a crop requiring 200 units at a yield of 25 per plant and a dome capacity of 16 THEN the engine reports 8 plants and 1 biodome for that crop
[...]

Issue #38, opened 2026-08-20
Implement Rollup Engine: producer and power stages


jonstump:
Governed by [SPEC-0001](docs/openspec/specs/rollup-engine/spec.md) ([design](docs/openspec/specs/rollup-engine/design.md)).

Completes the rollup engine. **Stage 1 is already built and merged** — `Resolve` covers dependency graph resolution, method and recipe selection, quantity propagation, cycle detection, exact arithmetic, provenance and determinism. This epic covers the two stages that turn a dependency tree into a build plan.

Per design.md, the engine is three sequential stages, each taking the prior stage's output plus its own configuration:

```
resolve (graph) → rollup (producers) → power (budget)
```

The tree canvas needs only stage 1. The base planner needs all three.

## Scope

| SPEC-0001 requirement | Status |
|---|---|
| Dependency Graph Resolution, Method Resolution, Recipe Selection, Quantity Propagation, Cycle Detection, Exact Arithmetic, Provenance Propagation, Determinism, Error Handling Standards | Already implemented in `internal/domain` |
| **Leaf Assignment to Bases** | This epic |
| **Producer Rollup** | This epic |
| **Power Computation** | This epic |

The nine implemented requirements are deliberately not re-planned. Only the three unbuilt ones become stories.

## The constants input is not what design.md assumes

design.md lists as Tier 2 — "injected, never hardcoded" — crop yields, dome capacity, extractor rates per class, generator outputs, class multipliers, draws, depot thresholds and capacities, battery ratios, fauna yield per cycle, cycle durations, and steps per nutrient processor.

Since that was written, the 2026-08-18 confirmation (ADR-0001) found most of those extractable, and SPEC-0004's normalizer now emits them in the Tier 1 artifact's `Economy` section:

| Constant | Source today |
|---|---|
| Crop yields, growth times | Tier 1 — `Economy.Crops` |
| Extractor and generator rates, storage | Tier 1 — `Economy.Parts.Primary` |
| Class multipliers | Tier 1 — `Economy.Hotspots.Strengths` / `Weightings` |
| Power draws | Tier 1 — `Economy.Parts.Dependencies` |
| Refiner throughput | Tier 1 — `Economy.Refining` |
| Biodome crop-slot count | Tier 2 — the one remaining curated constant per ADR-0001 |
| Battery ratio, fauna yield and cycle time, steps per nutrient processor, depot threshold | Not in `Economy` — curated, or needs a source |

[...]

PR #49, opened 2026-08-20
feat(domain): stage-2 inputs, base grouping, and constant sources

### LINKED ISSUE #39: Rollup inputs: base assignment and economy constants

**Foundation story** — both #40 and #41 depend on the types this establishes. Merge before either begins.

Stage 2 takes stage 1's `ResolvedGraph` plus its own configuration. This story defines that configuration and the leaf-to-base grouping every producer row is computed from — the boundary between `resolve` and `rollup`.

Extends `internal/domain` (currently `resolve.go`, `select.go`, `tier1.go`, `economy.go` — stage 1 only, no producer or power code exists).

## Where the constants come from

design.md says all economy constants arrive as a call-time parameter and names them "Tier 2". Most are now Tier 1: SPEC-0004's normalizer emits them in `Economy` (crop yields and growth times, part rates and storage, hotspot class strengths, power draws, refiner throughput). What remains genuinely curated is smaller — biodome crop-slot count per ADR-0001, plus battery ratio, fauna yield and cycle time, steps per nutrient processor, and the depot threshold, none of which `Economy` carries.

Settling that split is this story's substance. The injection principle from design.md holds either way — nothing is hardcoded — but the input type should read from `Economy` where the data exists rather than asking a caller to re-supply it.

## Requirements

- [ ] **Leaf Assignment to Bases** (SPEC-0001) — The engine MUST accept an assignment of leaf items to bases and MUST group leaf totals by their assigned base. Leaves with no assignment MUST be reported in a distinct unassigned group rather than being silently dropped or attributed to a default base.
  - WHEN crops are assigned to one base and gases to another THEN each base's rollup contains only its assigned leaf items
  - WHEN a leaf item has no base assignment THEN it appears in the unassigned group with its full required total
  - WHEN a leaf is reassigned from one base to another THEN both the origin and destination base rollups are recomputed, including their power budgets

## Scope

**In:**
- The stage-2 input type: base assignment (leaf item ID → base ID) plus the economy constants the later stages read
- Grouping stage 1's leaf totals by assigned base, with a distinct unassigned group
- Resolving each constant to its source — `Economy` field or curated input — and documenting the ones with no source yet
- Per-site configuration the later stories need: extractor class per base (SPEC-0001 REQ "Producer Rollup" requires class be configured per site, not per row)

**Out:**
- Producer counts and power math — #40 and #41

## The SPEC-0001 design amendment belongs to this story

`/sdd:check` confirmed this as a live ADR-vs-spec contradiction rather than a note for later:

> `docs/openspec/specs/rollup-engine/design.md:88-92` — "Tier 2 constants injected, never hardcoded" lists crop yields, dome capacity, extractor rates per class, generator outputs, class multipliers, draws, depot thresholds and capacities, battery ratios, fauna yield per cycle, cycle durations and steps per nutrient processor as "provisional and hand-curated".
>
> `docs/adrs/ADR-0001` now states the curated tier holds **one** entry — biodome crop-slot count — since generator and consumer rates, class scaling, crop yields and growth times all turned out extractable, and SPEC-0004's normalizer emits them in `Economy`.

Two accepted artifacts disagree about where half this story's inputs come from. Amending the design **is** part of this story rather than a follow-up, because the amendment is exactly the mapping this story has to settle to build the input type at all — and writing it down before building it is how the contradiction got here.

The amendment should record, per constant, whether it reads from `Economy` or arrives curated, and name the ones with no source yet. `docs/` is a protected path for `/sdd:work`, so it lands as a companion docs PR alongside the implementation, not in the same branch.

## Acceptance Criteria

- Leaves assigned to different bases group into their own rollups, with no cross-contamination
- An unassigned leaf appears in the unassigned group carrying its full total, and is neither dropped nor attributed to a default base
- Reassigning a leaf recomputes both the origin and destination base
- Extractor class is settable per base and readable by a later stage
- Every constant the later stories need either resolves to an `Economy` field or is an explicit curated input — none is hardcoded, and any with no source is recorded rather than invented
- The stage boundary keeps stage 1 free of economy constants: resolving a graph still requires no rollup configuration

### Branch

`feat/39-rollup-inputs-base-assignment-and-constants`

### PR Convention

```
Closes #39
Part of #38

Implements SPEC-0001 REQ "Leaf Assignment to Bases".
```

jonstump:
## What

Implements #39, the foundation story for #40 and #41. Establishes the boundary between resolving a graph and rolling it up: `RollupInput`, `SiteConfig`, `Curated`, `Constants`, and `GroupLeaves`.

## Settling the ADR-vs-spec contradiction

The story's substance was resolving where each constant comes from. SPEC-0001's design lists a dozen as "provisional and hand-curated"; ADR-0001 says the curated tier holds one entry. **Both were wrong, in opposite directions** — and I checked the generated artifact rather than trusting either list.

| Constant | Source | Design said |
|---|---|---|
| Crop yield, growth time | `Economy.Crops[]` | curated |
| Part rates, storage, power draws | `Economy.Parts[]` | curated |
| Class strengths C/B/A/S | `Economy.Hotspots[]` | curated |
| Refiner throughput | `Economy.Refining` | curated |
| **Depot capacity** | **`U_SILO_S.primary.storage` = 1,440,000** | curated |
| **Battery capacity** | **`U_BATTERY_S.primary.storage` = 45,000** | curated |
| Biodome crop slots | curated (ADR-0001's entry) | curated |
| Fauna yield, cycle time | curated — no source found | curated |
| Steps per nutrient processor | curated — planner policy | curated |
| Depot **threshold** | curated — planner policy | curated |

Depot and battery capacity have sources. The threshold — *when* a plan should start reporting depots — does not, because the game has no opinion about when hoarding starts. That distinction is now in the code: `Constants.DepotCapacity()` reads the artifact, `Curated.DepotThreshold` is supplied.

Two of the remaining five are planner policy rather than missing game data, which is worth separating: nobody will ever find them in a table.

## Design

- **Unassigned is a distinct group, not a default base.** It sorts last because it is a remainder. A plan that quietly builds Frost Crystal at whichever base sorted first is worse than one that says it does not know.
- **Reassignment needs no special handling.** `GroupLeaves` is a pure function of the graph and the assignment, so moving a leaf and re-running recomputes both bases. The test asserts the *origin* shrank, not just that the destination grew.
- **A stale assignment is an error.** Naming something that is not a leaf of this graph — because a method changed and it now expands — is reported rather than silently ignored.
- **An unconfigured base is refused.** An extractor class picked for the caller is a number they never chose appearing in their plan.
- **Every curated constant is required.** A zero is refused, not defaulted.
- **Part IDs are enumerated as constants** so the set of buildables this engine depends on is listable, and a rename fails in one place.

## One thing worth knowing about exactness

`ClassStrength` returns `*big.Rat`, and the conversion is exact for every value NMS 5.97 carries — 1, 1.5, 2, 2.5, 150, 220, 250, 300 are all binary-exact.

The exposure is upstream: `ClassValues` is `float64` in the schema, so a source value that is *not* binary-representable would already have been rounded at JSON decode, before this code sees it. No such value exists today, so this is latent rather than live. Noted in the accessor where it would bite. If one ever appears, the fix is in the schema, not here.

## Tests

Grouping tests run against the hand-authored fixture; constant-source tests run against `data/tier1.json`, because "it reads from `Economy`" is only meaningful against the `Economy` the normalizer actually emits.

- Crops at one base and gases at another, with no cross-contamination and sorted output
- An unassigned leaf carries its **full** 300 Condensed Carbon, is in a distinct group, and every leaf lands somewhere (counted against `len(g.Leaves())`)
- Reassignment: origin drops to 1 demand, destination rises to 2
- Extractor class differs per site and is readable from the group
- Four refusal cases: unconfigured base, class outside C/B/A/S, the empty base id, and a non-leaf assignment
- Every curated constant refused at zero, named individually
- `Resolve` still works on the fixture, **which carries no economy section at all** — if stage 1 had started depending on one, that test could not pass

Per the issue, the SPEC-0001 design amendment is a companion docs PR rather than part of this branch:

- `docs/openspec/specs/rollup-engine/design.md`, "Tier 2 constants injected, never hardcoded": replace the list of hand-curated constants with the per-constant source mapping above. Record that depot capacity and battery capacity read from `Economy`, that steps-per-processor and depot threshold are planner policy rather than missing game data, and that fauna yield and cycle time are the only genuinely unsourced game values.
- Same file: note that the injection principle is unchanged — nothing is hardcoded — but that "injected" now means "read from the artifact" for most of the list.

Closes #39
Part of #38

Implements SPEC-0001 REQ "Leaf Assignment to Bases".

### DIFF
--- a/internal/domain/rollup.go
+++ b/internal/domain/rollup.go
@@ -1,0 +1,457 @@
+package domain
+
+import (
+	"fmt"
+	"math/big"
+	"sort"
+)
+
+// Stage 2's inputs: which leaves belong to which base, how each site is
+// configured, and where every economy constant comes from.
+//
+// Governing: SPEC-0001 REQ "Leaf Assignment to Bases", ADR-0001 (two-tier
+// NMS data ingestion)
+//
+// This file is the boundary between resolving a graph and rolling it up.
+// Stage 1 takes a Tier 1 artifact and a target; it reads no economy
+// constant and needs no base assignment. Stage 2 takes stage 1's output and
+// everything below.
+
+// BaseID identifies a base within a plan. The empty value is not a base —
+// see Unassigned.
+type BaseID string
+
+// Unassigned is the group a leaf lands in when the plan does not say where
+// it comes from.
+//
+// It is deliberately a distinct group rather than a default base: SPEC-0001
+// requires unassigned leaves be "reported in a distinct unassigned group
+// rather than being silently dropped or attributed to a default base",
+// because a plan that quietly builds Frost Crystal at whichever base sorted
+// first is worse than one that says it does not know.
+const Unassigned BaseID = ""
+
+// HotspotClass is the C/B/A/S grade of a hotspot.
+//
+// The class belongs to the hotspot, never to the device: the game has one
+// extractor, not four. Per SPEC-0001 REQ "Producer Rollup" it is configured
+// per site, so every extractor row at a base recomputes together.
+type HotspotClass string
+
+const (
+	ClassC HotspotClass = "C"
+	ClassB HotspotClass = "B"
+	ClassA HotspotClass = "A"
+	ClassS HotspotClass = "S"
+)
+
+var validClasses = map[HotspotClass]bool{ClassC: true, ClassB: true, ClassA: true, ClassS: true}
+
+// Valid reports whether c is one of the four grades.
+func (c HotspotClass) Valid() bool { return validClasses[c] }
+
+// SiteConfig is the per-base configuration the producer stages read.
+type SiteConfig struct {
+	// ExtractorClass is the hotspot grade at this base. Per site rather
+	// than per row, so changing it recomputes every extractor at the base
+	// and nothing anywhere else.
+	//
+	// Governing: SPEC-0001 REQ "Producer Rollup" — "Extractor class MUST be
+	// configured per site, not per row."
+	ExtractorClass HotspotClass
+}
+
+// Curated holds the economy constants no game table states.
+//
+// Governing: SPEC-0001 design.md "Tier 2 constants injected, never
+// hardcoded"; ADR-0001 (curated tier)
+//
+// The list is short and getting shorter. SPEC-0004's normalizer now emits
+// most of what that design entry called Tier 2 — crop yields and growth
+// times, part rates, storage buffers, power draws, hotspot class strengths,
+// refiner throughput — and Constants reads those from the artifact rather
+// than asking a caller to supply them again. What remains here is what a
+// search of the tables did not find.
+//
+// Every field is required. A zero is refused rather than defaulted,
+// because a silently-assumed constant is indistinguishable from a measured
+// one once it reaches a number the user reads.
+type Curated struct {
+	// BiodomeCropSlots is how many plants fit in a biodome. ADR-0001's one
+	// surviving curated entry; the community wiki reports 16 and no table
+	// searched states it. It may be geometric — snap points in the scene —
+	// and therefore extractable after all.
+	BiodomeCropSlots int64
+
+	// FaunaYieldPerCycle is how many units one creature yields per
+	// collection cycle, and FaunaCycleSeconds how long that cycle is.
+	// Neither appears in the reality tables; fauna products are generated
+	// from creature data this pipeline does not read.
+	FaunaYieldPerCycle int64
+	FaunaCycleSeconds  int64
+
+	// StepsPerProcessor is how many nutrient processor steps one processor
+	// covers. A planner policy about how hard to work one machine rather
+	// than a game constant.
+	StepsPerProcessor int64
+
+	// DepotThreshold is the required quantity above which a row reports
+	// supply depots. Also a planner policy — the game does not have an
+	// opinion about when hoarding starts.
+	//
+	// Depot *capacity* is not here: it is U_SILO_S's storage buffer, which
+	// the artifact carries. See Constants.DepotCapacity.
+	DepotThreshold int64
+}
+
+// validate refuses a partially-specified constant set.
+func (c Curated) validate() error {
+	for _, f := range []struct {
+		name string
+		v    int64
+	}{
+		{"biodome crop slots", c.BiodomeCropSlots},
+		{"fauna yield per cycle", c.FaunaYieldPerCycle},
+		{"fauna cycle seconds", c.FaunaCycleSeconds},
+		{"steps per processor", c.StepsPerProcessor},
+		{"depot threshold", c.DepotThreshold},
+	} {
+		if f.v <= 0 {
+			return fmt.Errorf("%w: curated constant %q is %d; it must be supplied, not defaulted",
+				ErrInvalidArtifact, f.name, f.v)
+		}
+	}
+	return nil
+}
+
+// Part IDs the rollup reads by name. Named here rather than inline so the
+// set of buildables this engine depends on is enumerable, and so a game
+// update that renames one fails in a single place.
+const (
+	PartExtractorMineral = "U_EXTRACTOR_S"
+	PartExtractorGas     = "U_GASEXTRACTOR"
+	PartSupplyDepot      = "U_SILO_S"
+	PartBattery          = "U_BATTERY_S"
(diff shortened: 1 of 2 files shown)


jonstump (commented):

Spec-aware review against issue #39's acceptance criteria, SPEC-0001, and ADR-0001. Verdict: **approve**. (Comment review — GitHub blocks `APPROVE` on a self-authored PR.)

Conflict-marker gate clean, CI green. Verified locally on `3f30cff`: `go test -race ./...` passes across all packages.

### The constant-source finding is right, and I checked it

The story's job was settling where each constant comes from, and "both were wrong, in opposite directions" is the correct verdict. I read the two claims that move the mapping rather than trusting the table, straight out of the merged `data/tier1.json`:

| Part | Field | Value |
|---|---|---|
| `U_SILO_S` | `primary.storage` | **1,440,000** |
| `U_BATTERY_S` | `primary.storage` | **45,000** |

Both are there, both are what the PR says. So depot capacity and battery capacity are Tier 1 values that ADR-0001 and SPEC-0001 both had filed as curated.

The distinction drawn around the depot is the part worth keeping: **capacity** has a source and **threshold** does not, because the game has no opinion about when hoarding starts. `Constants.DepotCapacity()` reading the artifact while `Curated.DepotThreshold` is supplied puts that difference in the type system rather than in a comment. Separating "planner policy" from "missing game data" among the remaining curated five is the same instinct — nobody will ever find steps-per-processor in a table, and saying so stops the next person looking.

### The exactness note is honest about where the exposure actually is

`ClassStrength` returning `*big.Rat` is right, and so is the observation that the risk sits upstream: `ClassValues` is `float64` in the schema, so a non-binary-representable source value would already be rounded at decode, before this code sees it. Verified against the artifact — 1, 1.5, 2, 2.5 and 150, 220, 250, 300 are all binary-exact, so this is latent rather than live, exactly as claimed. Recording it at the accessor where it would bite, rather than as a TODO, is the right placement.

`SetFloat64` returning nil on NaN/Inf is handled, and the sign check catches a zero or negative strength.

### Acceptance criteria

| Criterion (issue #39) | Status |
|---|---|
| Leaves group by base with no cross-contamination | Met |
| Unassigned leaf carries its full total in a distinct group | Met — asserted at 300 Condensed Carbon, and every leaf counted against `len(g.Leaves())` |
| Reassignment recomputes both bases | Met — and the test asserts the **origin shrank**, not merely that the destination grew |
| Extractor class settable per base, readable by a later stage | Met |
| Every constant resolves to `Economy` or an explicit curated input; unsourced ones recorded | Met, and improved on — two moved from curated to sourced |
| Stage 1 stays free of economy constants | Met — and proven by the fixture carrying no economy section at all, so a new dependency could not pass silently |

That last test is the one I would keep if I could keep only one: it fails for the right reason if stage 1 ever starts needing an `Economy`.

---

### Note — the deferred amendment corrects one direction of the error, not both

The `### Deferred Design Doc Updates` section covers `rollup-engine/design.md`, which **over**-counted: it lists a dozen constants as hand-curated when most read from `Economy`.

It does not cover `ADR-0001:38`, which **under**-counts in the opposite direction:

> As of the 2026-08-18 confirmation that is one entry — biodome crop-slot count

This PR's own table shows **five** curated values: biodome crop slots, fauna yield, fauna cycle time, steps per nutrient processor, and depot threshold. Two of those are planner policy and three are game data, but none of them is biodome crop slots, and the ADR says there is only one.

So a reader who follows the amendment as scoped ends up with a corrected design doc and an ADR that still says the curated tier holds one entry — which is the claim this PR disproves most directly. Worth adding to the deferred list.

Related: `ADR-0001:207` already lists battery capacity as having moved to Tier 1, so the ADR is not merely stale but internally inconsistent — line 207 knows something line 38 does not.

**On the deferral itself**: the split is correct per the issue and `docs/` is protected from this branch, so this is not a complaint about the mechanism. But this PR says `Closes #39`, so the story closes and the amendment lives in a merged PR body until someone opens the companion. That is the precise shape of the drift `/sdd:check` found last run — the 34-node count, the localisation archive, and this very Tier 2 list were all once deferred updates. The pattern works when the follow-up gets opened; it is worth opening it now rather than after the story is closed.

Issue #45, opened 2026-08-20
Module lifecycle, surface, and domain purity


jonstump:
Part of #42. Governed by [SPEC-0002](docs/openspec/specs/wasm-boundary/spec.md), implementing [ADR-0003](docs/adrs/ADR-0003-go-wasm-domain-core.md).

**Blocked by #43 and #44** — wires the envelope and the error codes into actual entry points.

The `syscall/js` half, and the only story in this epic that needs a WASM build to exercise fully. Registers one namespace, gates it on an explicit readiness signal, wires `resolve` to `domain.Resolve`, and turns ADR-0003's purity claim into an assertion.

## Requirements

- [ ] **Boundary Surface** (SPEC-0002) — The adapter MUST expose one named entry point per domain stage on a single namespace object, coarse-grained so one call performs one complete stage, exposing no state-mutating function. `resolve` is REQUIRED; `rollup` and `power` are RESERVED for stages 2 and 3 and MUST follow the same contract when added.
  - WHEN the module has initialized THEN exactly one namespace object is registered, carrying the stage entry points, and no domain function is reachable from the global scope directly
  - WHEN the view resolves a plan and renders 34 nodes THEN exactly one boundary crossing occurred, and node data was read from the single returned value
  - WHEN `rollup` or `power` is added in a later stage THEN it accepts and returns the same envelope shape defined by REQ "Result Envelope", with no bespoke calling convention

- [ ] **Module Lifecycle and Readiness** (SPEC-0002) — Module instantiation and Tier 1 artifact loading MUST be distinct steps, with an explicit readiness signal gating entry points and separate diagnostics for each failure.
  - WHEN an entry point is called before the readiness signal has resolved THEN it returns a failure envelope with the not-ready code, and does not hang or return undefined
  - WHEN the module instantiates successfully but the supplied Tier 1 artifact fails validation THEN the failure is reported as an invalid-artifact error, not as a module load failure

- [ ] **Event Loop Safety** (SPEC-0002) — A call MUST NOT block the JavaScript event loop such that the page becomes unresponsive, and MUST NOT deadlock between the Go runtime and the event loop.
  - WHEN a resolve call executes THEN the page does not become unresponsive, and no deadlock occurs between the Go runtime and the JavaScript event loop

- [ ] **Domain Purity Preservation** (SPEC-0002) — `syscall/js` MUST be reachable only through the adapter package, the adapter MUST hold no domain logic, and the encoding paths MUST run under plain `go test`.
[...]

Issue #44, opened 2026-08-20
Error codes and the failure contract


jonstump:
Part of #42. Governed by [SPEC-0002](docs/openspec/specs/wasm-boundary/spec.md).

**Blocked by #43** — the error payload is the other half of the envelope that story defines.

The failure half of the contract. `errors.Is` does not survive a boundary crossing, and matching on message text is the classic brittle coupling — so every domain sentinel maps to a stable code that the view branches on, while the wrapped Go message crosses alongside as unstructured diagnostic text.

Still plain `go test` — no WASM build needed.

## Requirements

- [ ] **Sentinel Error Preservation** (SPEC-0002) — Each domain sentinel MUST map to a stable identifier carried in the error payload, decoupled from Go error text, with a reserved code for errors matching no sentinel.
  - WHEN a plan names an item absent from the Tier 1 artifact THEN the error payload carries the code for unknown item, and the consumer branches on that code alone
  - WHEN a domain error's wrapped message text is reworded THEN the code it crosses with is unchanged
  - WHEN an error matches none of the defined sentinels THEN it crosses with the reserved unclassified code, not with the code of an unrelated sentinel

- [ ] **Contract Versioning** (SPEC-0002) — The module MUST report a contract version the consumer can check, and a mismatch MUST be reported rather than consumed.
  - WHEN the view is built against one contract version and loads a module reporting another THEN it reports a version mismatch naming both versions, and does not attempt to consume the payload

- [ ] **Error Handling Standards** (SPEC-0002) — Decoding failures MUST name what could not be decoded and MUST NOT attempt computation, and distinct failure kinds MUST carry distinct codes.
  - WHEN the view passes input that cannot be decoded into a valid plan THEN the failure names what could not be decoded, and no computation is attempted
  - WHEN a call fails because the module is not ready, versus because the input was malformed THEN the two failures carry different codes

## Notes

**The unclassified code is the point, not a fallback.** Silently mapping an unrecognized error onto the nearest sentinel makes the view branch confidently on a wrong kind — a worse failure than admitting ignorance. The third scenario exists to pin that, and it is the one an implementation is most likely to get wrong by being helpful.

**The reworded-message scenario is a test about coupling.** It passes trivially the day it is written and only earns its keep later, when someone improves an error message. Worth writing as an explicit test that changes a message and asserts the code is stable, rather than trusting the design.

[...]

Issue #42, opened 2026-08-20
Implement WASM Boundary


jonstump:
Governed by [SPEC-0002](docs/openspec/specs/wasm-boundary/spec.md) ([design](docs/openspec/specs/wasm-boundary/design.md)), implementing [ADR-0003](docs/adrs/ADR-0003-go-wasm-domain-core.md) and [ADR-0004](docs/adrs/ADR-0004-react-view-layer.md).

The adapter between the Go domain core and the JavaScript view. **Nothing exists yet** — no `syscall/js` anywhere in the tree, no adapter package. All 11 requirements and 27 scenarios are unbuilt.

This is where SPEC-0001's guarantees are most easily lost. That spec spends real effort on exact integer and rational arithmetic; JavaScript's only numeric type is IEEE-754 double, exact for integers only to 2^53−1 against Go's int64 range. A naive marshalling step discards at the last moment exactly what the engine was built to preserve.

## Only `resolve` is buildable now, and that is by design

SPEC-0002 REQ "Boundary Surface" marks `rollup` and `power` as **RESERVED**: they must follow the same envelope contract when added, but the boundary ships without them. Stage 1 (`domain.Resolve`) is merged and is all the tree canvas needs.

So this epic does **not** block on #38 (SPEC-0001 stages 2 and 3). When #40 and #41 land, their entry points slot into the envelope this epic defines — no bespoke calling convention, per that requirement's third scenario.

## The design decisions the stories implement

| Decision | Why it is not negotiable |
|---|---|
| JSON strings across the boundary, not `js.Value` construction | A JSON boundary is exercisable under plain `go test`. `js.Value` trees are testable only inside a running WASM instance, which would make the most correctness-critical code the least testable. |
| Every quantity crosses as a decimal string, uniformly | "Number when it fits, string when it doesn't" creates a branch that never runs in testing and breaks in production on the one large value. |
| Stable error codes decoupled from Go error text | `errors.Is` does not survive the crossing, and matching on message text is brittle. A reserved unclassified code beats mapping an unknown error onto the nearest sentinel. |
| The adapter holds no domain logic | The concrete temptation is rounding. Rounding is a domain rule — SPEC-0001 names which physical boundaries round and in which direction — and it stays there. |
| Readiness separate from artifact load | They fail for unrelated reasons. A malformed artifact is a data problem; a failed instantiation is a build problem. Collapsing them produces a message that helps with neither. |

## Stories

[...]

PR #50, opened 2026-08-20
feat(bridge): result envelope, exact quantity encoding, determinism

### LINKED ISSUE #43: Envelope, quantity encoding, and determinism
Part of #42. Governed by [SPEC-0002](docs/openspec/specs/wasm-boundary/spec.md).

**Foundation story** — #44 and #45 both build on the envelope and encoding this establishes. Merge before either begins.

The success half of the boundary contract: what a result looks like on the wire, how quantities survive the crossing, and that the same input encodes identically every time. All of it testable under plain `go test` — no browser, no WASM build.

Greenfield. Creates the adapter package; `internal/domain` is consumed, not modified.

## Requirements

- [ ] **Result Envelope** (SPEC-0002) — Every entry point MUST return a single envelope carrying either a result payload or an error payload, never both and never neither.
  - WHEN a call succeeds THEN the envelope reports success, carries the result payload, and carries no error payload
  - WHEN graph resolution fails THEN the envelope reports failure, carries the error payload, and carries no result payload — not an empty one

- [ ] **Exact Quantity Encoding** (SPEC-0002) — Every quantity crossing the boundary MUST encode as a decimal string, uniformly, with no conversion through `float64`.
  - WHEN Condensed Carbon resolves to a total of 300 THEN the encoded value is the string `"300"`, not the number `300`
  - WHEN a total exceeds 2^53−1, which JavaScript cannot represent exactly as a number THEN the encoded string carries the exact value, and parsing it as a `BigInt` on the consuming side yields that value unchanged
  - WHEN a later stage produces a total that is exactly one and a half THEN the encoded value represents that quantity exactly, and no rounding or truncation is applied at the boundary
  - WHEN the adapter encodes any quantity THEN no conversion through `float64` occurs, verified by the absence of such conversions in the encoding code

- [ ] **Recipe Selection Crossing** (SPEC-0002) — Per ADR-0005 a method no longer identifies a recipe, so the plan payload MUST carry recipe selection alongside method, and resolved nodes MUST carry their legal recipes.
  - WHEN a plan specifying a non-default recipe for one node crosses the boundary and is returned THEN that node reports the specified recipe, and its expansion reflects it
  - WHEN a plan in which every node uses its default recipe is encoded THEN the payload contains no recipe selections
  - WHEN the view receives a resolved graph THEN each node carries its legal recipes for the chosen method, without the view reading the artifact

- [ ] **Determinism Across the Boundary** (SPEC-0002) — Encoding MUST be deterministic, preserving the domain's ordering and provenance.
  - WHEN the same plan input is resolved and encoded twice in the same process THEN the two encoded outputs are byte-identical
  - WHEN an encoded graph is decoded by the consumer THEN node order matches the domain's order — terminals first, target last
  - WHEN a node is marked unverified by the domain THEN it arrives at the consumer marked unverified

## Notes

**The `float64` scenario wants a real check, not a claim.** "No conversion through float64 occurs, verified by the absence of such conversions in the encoding code" is asking for something greppable or vet-able, not a comment saying so. `big.Rat` has `FloatString` and `RatString`; only one of those is safe here.

**A total of exactly one and a half is not hypothetical.** `Applications()` already returns `5/2` for the 125-at-yield-50 case in SPEC-0001, and stages 2–3 produce more of them. The encoding has to carry a rational exactly, which is what makes "decimal string" a decision about *representation* rather than formatting.

**Empty is not absent.** The Result Envelope scenario is explicit that a failure carries *no* result payload — "not an empty one". A zero-valued struct that marshals to `{}` fails this.

**Defaults cost no payload.** The plan encoding mirrors SPEC-0001's `PlanInput.Recipes`, where absence is meaningful: a node using its default records nothing. That property is what keeps the URL hash small, and it is asserted domain-side already.

## Acceptance Criteria

- All twelve scenarios above pass as named tests, under plain `go test` with no WASM build
- Every quantity field in the encoded payload is a string; none is a JSON number
- A value above 2^53−1 round-trips exactly through the string encoding
- A rational total encodes without rounding or truncation
- The encoding path contains no `float64` conversion, and this is verified mechanically rather than asserted in a comment
- A failure envelope carries no result payload, empty or otherwise
- Encoding the same input twice is byte-identical, and node order matches the domain's

### Branch

`feat/43-envelope-quantity-encoding-and-determinism`

### PR Convention

```
Closes #43
Part of #42

Implements SPEC-0002 REQ "Result Envelope", REQ "Exact Quantity Encoding",
REQ "Recipe Selection Crossing", REQ "Determinism Across the Boundary".
```

jonstump:
## What

Implements #43, the foundation story for #44 and #45. New `internal/bridge` package — the encoding half of ADR-0003's adapter, deliberately free of `syscall/js` so **all twelve scenarios run under plain `go test` with no WASM build**.

## Encoding decisions

**Quantities cross as strings, never JSON numbers.** `big.Rat.RatString` is the only formatter used: integral values render as digits (`"300"`), fractional ones as rationals (`"3/2"`), and both round-trip exactly. `FloatString` rounds to a fixed number of places and `Float64` leaves the exact domain, so neither appears anywhere in the package.

**`Data` and `Error` are pointers**, so exactly one is present. A failure marshals with **no `data` key at all** rather than an empty object. The requirement says "carries no result payload — not an empty one", and `{}` reads to a consumer as a call that succeeded and produced nothing.

**Node order is the domain's, unchanged.** The adapter does not re-sort — the tab order is a domain decision per SPEC-0001, and re-deriving it here would be a second place for it to drift.

## The float64 scenario, verified rather than claimed

The issue flagged that "verified by the absence of such conversions in the encoding code" wants something mechanical. `TestNoFloatInTheEncodingPath` parses this package's own source with `go/ast` and fails on a `float64`/`float32` conversion, a `FloatString` call, or a `Float64`/`Float32`/`SetFloat64` call — naming file and line.

I checked it has teeth by swapping `RatString` for `FloatString(6)`:

```
bridge_test.go:319: bridge.go:61:18: FloatString call in the encoding path
```

## A test that was silently passing by not running

The safe-integer scenario first used a graph total of 5×10¹⁷. That is past 2^53−1, but it happens to be **double-representable** — and the guard I wrote against exactly that case called `t.Skip`, so the whole test was skipped and none of its assertions ran. Vacuous.

It now also encodes **2^53+1** — the smallest integer a double cannot hold, and precisely what a consumer parsing with `Number` instead of `BigInt` would silently round — and asserts the string round-trips exactly while the `float64` round trip does not.

Being past 2^53−1 is the criterion the scenario cares about; double-representability was the wrong property to branch on.

## Tests — twelve scenarios, one per spec scenario

| Requirement | Scenarios |
|---|---|
| Result Envelope | success carries payload and no `error` key; failure carries error and **no `data` key**, checked both as a substring and structurally |
| Exact Quantity Encoding | `"300"` as a string, plus a structural walk asserting *every* `total`/`quantity`/`perUnit`/`yield`/`applications` in the payload is a JSON string; 2^53+1 survives; `3/2` is not rounded; no float in the path |
| Recipe Selection Crossing | a pinned non-default recipe round-trips **and its expansion changes** (250 Frost Crystal, not 40); an all-defaults plan carries neither key; every non-terminal node carries legal recipes containing its selection |
| Determinism | 25 encode runs byte-identical, graph and plan; node order matches the domain index-for-index with a terminal first and the target last; unverified nodes arrive unverified |

The provenance and legal-recipe tests both `t.Fatal` if the fixture stops exercising them, so neither can pass by finding nothing to check.

`go vet`, `staticcheck v0.8.0-rc.1`, `gofmt -l` clean. No `syscall/js` import anywhere in `internal/`.

## Scope note

Sentinel error codes are #44. Everything crosses as `UNCLASSIFIED` for now — the reserved code SPEC-0002 requires exist so an unmatched error is never silently mapped onto an unrelated sentinel. #44 fills in the mapping and will likely extend `ErrorPayload`; that is expected of a foundation story rather than a guess I should have made here.

Closes #43
Part of #42

Implements SPEC-0002 REQ "Result Envelope", REQ "Exact Quantity Encoding", REQ "Recipe Selection Crossing", REQ "Determinism Across the Boundary".

### DIFF
--- a/internal/bridge/bridge.go
+++ b/internal/bridge/bridge.go
@@ -1,0 +1,310 @@
+// Package bridge encodes domain values for the boundary the view consumes.
+//
+// Governing: ADR-0003 (Go domain, thin adapter), SPEC-0002 REQ "Result
+// Envelope", REQ "Exact Quantity Encoding", REQ "Recipe Selection Crossing",
+// REQ "Determinism Across the Boundary"
+//
+// ADR-0003 splits the application into a domain package that imports no
+// syscall/js and a thin adapter that is the only code permitted to touch
+// js.Value. This package is the encoding half of that adapter, and it is
+// deliberately free of syscall/js too: everything here is testable under
+// plain `go test` with no WASM build, and the js-facing shim is a later
+// story that marshals what this produces.
+//
+// The one rule that shapes every type below: a quantity never becomes a
+// JSON number. JavaScript's number type cannot hold 2^53 exactly, and the
+// engine's whole exactness commitment would be spent one step before the
+// value a user reads.
+package bridge
+
+import (
+	"encoding/json"
+	"fmt"
+	"math/big"
+
+	"github.com/jonstump/nms-base-planner/internal/domain"
+)
+
+// ContractVersion is the boundary contract this package implements.
+//
+// Governing: SPEC-0002 REQ "Contract Versioning" — "The version MUST change
+// whenever the envelope shape, the encoding of quantities, or the sentinel
+// code set changes." The consuming view checks it and refuses a mismatch
+// rather than parsing an unexpected shape.
+const ContractVersion = "1.0.0"
+
+// Quantity is an exact quantity on the wire.
+//
+// Governing: SPEC-0002 REQ "Exact Quantity Encoding" — "Every quantity
+// crossing the boundary MUST be encoded as a decimal string, never as a
+// JavaScript number ... Where a total is not an exact integer, it MUST be
+// encoded as an exact decimal or rational string that round-trips without
+// loss."
+//
+// Integral values render as decimal digits ("300"); a fractional value
+// renders as a rational ("5/2"). Both round-trip exactly. The alternative
+// for the fractional case — a fixed number of decimal places — has to
+// choose a precision, and any choice truncates some value the engine
+// computed exactly.
+type Quantity string
+
+// QuantityOf renders an exact rational.
+//
+// big.Rat.RatString is the only formatter used: it emits "a/b", or just "a"
+// when the denominator is one, and it is lossless in both cases.
+// FloatString rounds to a fixed number of places, and Float64 is a
+// conversion this package must not make.
+func QuantityOf(r *big.Rat) Quantity {
+	if r == nil {
+		return "0"
+	}
+	return Quantity(r.RatString())
+}
+
+// QuantityOfInt renders an integer quantity.
+func QuantityOfInt(v int64) Quantity { return Quantity(new(big.Int).SetInt64(v).String()) }
+
+// Rat parses a Quantity back to an exact rational, which is what makes
+// "round-trips without loss" checkable rather than asserted.
+func (q Quantity) Rat() (*big.Rat, bool) {
+	r, ok := new(big.Rat).SetString(string(q))
+	return r, ok
+}
+
+// Envelope is what every entry point returns.
+//
+// Governing: SPEC-0002 REQ "Result Envelope" — "a single envelope carrying
+// an explicit success flag, exactly one of a result payload or an error
+// payload, and the contract version."
+//
+// Data and Error are pointers so that exactly one is present on the wire. A
+// failure marshals with no data key at all rather than with an empty object:
+// the requirement's scenario says "carries no result payload — not an empty
+// one", and a zero-valued struct that renders as {} reads to a consumer as
+// a successful call that produced nothing.
+type Envelope struct {
+	OK              bool           `json:"ok"`
+	ContractVersion string         `json:"contractVersion"`
+	Data            *ResultPayload `json:"data,omitempty"`
+	Error           *ErrorPayload  `json:"error,omitempty"`
+}
+
+// ResultPayload is the success half.
+type ResultPayload struct {
+	Graph *Graph `json:"graph,omitempty"`
+}
+
+// ErrorPayload is the failure half.
+//
+// Code is a stable machine-readable identifier; the sentinel mapping that
+// fills it is #44's story, and until then everything crosses as
+// CodeUnclassified. Message carries the domain's prose, including the
+// wrapped resolution path, with no contractual guarantee of format.
+type ErrorPayload struct {
+	Code    string `json:"code"`
+	Message string `json:"message"`
+}
+
+// CodeUnclassified is the code reserved for a failure matching no known
+// sentinel. SPEC-0002 REQ "Sentinel Error Preservation" requires the
+// reserved code exist so an unmatched error is never silently mapped onto
+// an unrelated sentinel; the rest of the code set arrives with #44.
+const CodeUnclassified = "UNCLASSIFIED"
+
+// Success builds an envelope carrying a payload and no error.
+func Success(data ResultPayload) Envelope {
+	return Envelope{OK: true, ContractVersion: ContractVersion, Data: &data}
+}
+
+// Failure builds an envelope carrying an error and no payload.
+func Failure(code, message string) Envelope {
+	if code == "" {
+		code = CodeUnclassified
+	}
+	return Envelope{
+		OK:              false,
+		ContractVersion: ContractVersion,
+		Error:           &ErrorPayload{Code: code, Message: message},
+	}
+}
+
+// Plan is the plan state crossing in both directions.
+//
+// Governing: SPEC-0002 REQ "Recipe Selection Crossing"
+//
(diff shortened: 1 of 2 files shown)


jonstump (commented):
## Review: Approved

Spec-aware review against issue #43's acceptance criteria, SPEC-0002, and ADR-0003. Verdict: **approve**. (Comment review — GitHub blocks `APPROVE` on a self-authored PR.)

Conflict-marker gate clean, CI green. Verified locally on `b9cbd84`: `go test -race ./...` passes across all packages, all twelve scenarios under plain `go test` with no WASM build.

No findings. Two things I checked rather than took on trust.

### The float64 guard has real teeth — I tried to defeat it

REQ "Exact Quantity Encoding"'s fourth scenario asks for "no conversion through `float64` ... verified by the absence of such conversions in the encoding code", and issue #43 flagged that this wants something mechanical rather than a comment claiming compliance. `TestNoFloatInTheEncodingPath` parses the package with `go/ast`.

I injected a conversion into `bridge.go` to see whether it would notice:

```go
func reviewerProbe(r float64) float64 { return float64(r) * 1.0 }
```

```
bridge_test.go:330: bridge.go:313:48: float64 conversion in the encoding path
bridge_test.go:358: bridge.go: "float64" appears in code: func reviewerProbe(...)
```

Caught at the exact line, by both the AST walk and the text-level check. That is the scenario genuinely satisfied rather than asserted — and the belt-and-braces second check means a construction the AST walk does not model still gets flagged.

Checking it yourself by swapping `RatString` for `FloatString(6)` before shipping it is the right instinct; a guard nobody has seen fail is a guard nobody knows works.

### `syscall/js` purity confirmed

`go list -deps ./internal/bridge` returns zero matches. The only mentions in `internal/` are the governing comments in `bridge.go` and `domain/errors.go` explaining why it is absent. REQ "Domain Purity Preservation" belongs to #45, but this package is the one that will host the temptation, and it starts clean.

### The vacuous-test catch is the most valuable thing here

> The safe-integer scenario first used a graph total of 5×10¹⁷. That is past 2^53−1, but it happens to be double-representable — and the guard I wrote against exactly that case called `t.Skip`, so the whole test was skipped and none of its assertions ran.

A test that skips is a test that reports success. Finding one you wrote yourself, in the scenario whose entire purpose is precision, and then fixing the *criterion* rather than the value — "being past 2^53−1 is what the scenario cares about; double-representability was the wrong property to branch on" — is exactly the reasoning that keeps a suite meaningful.

The replacement is better than the original in a way worth naming: **2^53+1 is the smallest integer a double cannot hold**, so it is precisely the value that a consumer parsing with `Number` instead of `BigInt` would silently round. Asserting the string round-trips while the `float64` round trip does not tests the consumer hazard, not just the encoder. And the guard is now `t.Fatalf` — "does not exceed 2^53-1, so this proves nothing" — so it cannot go quiet again.

The same defensive shape appears in the provenance and legal-recipe tests, which `t.Fatal` if the fixture stops exercising them. Three tests that refuse to pass by finding nothing to check.

### Acceptance criteria

| Criterion (issue #43) | Status |
|---|---|
| All twelve scenarios pass under plain `go test`, no WASM build | Met |
| Every quantity field a string, none a JSON number | Met — structural walk over `total`/`quantity`/`perUnit`/`yield`/`applications` |
| A value above 2^53−1 round-trips exactly | Met — 2^53+1, with the float round trip asserted to fail |
| A rational total encodes without rounding | Met — `3/2` |
| No `float64` conversion, verified mechanically | Met — AST guard, teeth confirmed above |
| Failure envelope carries no result payload, empty or otherwise | Met — pointer fields, so the `data` key is absent rather than `{}`, checked both as substring and structurally |
| Same input encodes byte-identically; node order matches the domain | Met — 25 runs, and order compared index-for-index with a terminal first and target last |

### Design calls worth keeping

**`Data` and `Error` as pointers.** The requirement says "carries no result payload — not an empty one", and the reasoning is right: `{}` reads to a consumer as a call that succeeded and produced nothing, which is a different and worse claim than failure.

**Node order is not re-derived.** The tab order is a SPEC-0001 decision; re-sorting here would be a second place for it to drift. Passing it through unchanged is what makes the determinism scenario about the boundary rather than about a reimplementation of the domain's ordering.

**`RatString` as the only formatter.** `FloatString` rounds and `Float64` leaves the exact domain, so naming both as banned — and enforcing that with the AST guard — closes the gap between "we chose the right function" and "nobody can quietly choose a wrong one later".

### Merge ordering

No file overlap with #49 — this is `internal/bridge`, that is `internal/domain`. Independent, either order.

PR #52, opened 2026-08-20
feat(domain): producer rollup for farm, extractor, ranch and kitchen

### LINKED ISSUE #40: Producer rollup: farm, extractor, ranch and kitchen

jonstump:
## What

Implements #40 — the largest story in the epic. Turns each base's assigned leaf totals into construction instructions across four producer types. **All ten Producer Rollup scenarios have a named test** (20 passing test cases including subtests).

## The three traps, each with a test that fails under the obvious implementation

**Feeders are per base.** Two fed fauna products at one base report **one** feeder. A per-row implementation reports two and passes every single-row test, so the test asserts two ranch rows exist first — otherwise it would pass by finding nothing.

**Processors are `ceil(total / per processor)`, not the sum of per-row ceilings.** Four steps at two per processor is **2**. Summing per-row ceilings gives 4, since `ceil(1/2)` is 1 four times over. The two answers differ whenever a base has more than one step.

**Extractor class belongs to the site.** The test builds two bases with two extractor rows each, flips one base from B to S, and asserts *every* row at that base recomputed **and** that no row at the other base moved.

## Rounding

Only at SPEC-0001's enumerated boundaries, always up, via integer `QuoRem` on numerator and denominator — no float at the one place the engine is allowed to lose precision.

Biodomes round from **plants**, not from the requirement. 17 plants at a capacity of 16 is two domes; rounding once from the requirement would give one. Tested directly.

Plants size on the **pessimistic bound** of the yield range. A plan sized on the best case under-builds, and the failure mode is a farm that never fills the order.

## Four more curated constants, and why

The producer stage needs four things no table states. They went into `Curated` rather than being inferred:

| Constant | Why it is curated |
|---|---|
| `ProcessSeconds` | `Economy.Refining` gives throughput per cycle but not the cycle's duration |
| `FaunaProducts` | Nothing says Wild Milk comes from a creature |
| `ResourceHotspots` | Nothing says Sulphurine sits on a Gas hotspot rather than a Mineral one |
| `FaunaYieldPerCycle`, `FaunaCycleSeconds` | already curated from #39 |

Crops classify themselves — `Economy.Crops` names the substance each plant yields — but the other two do not. Guessing a hotspot category from an item's name is exactly the inference that has cost this project before, so **an unclassified resource fails naming itself** rather than defaulting to Mineral. There is a test for that.

## One thing I could not resolve from the source

`ExtractorRate` scales the part's rate by the hotspot class strength, both read from the artifact. **The unit of that rate is not stated anywhere.** `U_EXTRACTOR_S` carries rate 100 against storage 360000, which fills in 3,600 of whatever unit the rate is denominated in — an hour, if the rate is per second.

The rollup is consistent either way, because `SiteConfig.FillSeconds` is supplied in the same unit and only the product `rate × window` matters. But **the plan's absolute times are only as right as that reading**, and it is a reading rather than a fact. Documented on the accessor; worth confirming in game before any duration is shown to a user.

The spec's scenario ("200 units per hour over 1.5 hours → 2 extractors") is reproduced with the identical ratio in integer inputs — rate 100 over a window of 3 — with the scaling explained in the test.

## Also

- The unassigned group produces no construction instructions: there is no site to build at. It stays visible in the `Grouping` so a caller can report it.
- A byproduct-covered item gets **no row of any kind** — not a zero-count row — and is reported under `NoBuild` carrying its full demand, so it is visible rather than dropped.
- `TestNoConstantIsHardcoded` varies crop yield, dome capacity, depot capacity and fauna yield independently and asserts each changes the answer.

Closes #40
Part of #38

Implements SPEC-0001 REQ "Producer Rollup".

### DIFF
--- a/internal/domain/producer.go
+++ b/internal/domain/producer.go
@@ -1,0 +1,496 @@
+package domain
+
+import (
+	"fmt"
+	"math/big"
+	"sort"
+)
+
+// Stage 2's producer half: each base's assigned leaf totals become concrete
+// construction instructions.
+//
+// Governing: SPEC-0001 REQ "Producer Rollup", REQ "Exact Arithmetic and
+// Rounding Discipline"
+//
+// Three of this requirement's scenarios exist because the obvious
+// implementation gets them wrong, and each has its own test: feeders are per
+// base rather than per row, nutrient processors are ceil(total steps / steps
+// per processor) rather than the sum of per-row ceilings, and extractor
+// class is a property of the site.
+//
+// Every quantity below stays a big.Rat until the moment it becomes a thing
+// you build. Rounding happens only at the boundaries SPEC-0001 enumerates —
+// plants, biodomes, extractors, depots, fauna, processors — and always up,
+// because half a biodome is not a biodome.
+
+// ProducerType is the kind of thing a row tells you to build.
+type ProducerType string
+
+const (
+	ProducerFarm      ProducerType = "farm"
+	ProducerExtractor ProducerType = "extractor"
+	ProducerRanch     ProducerType = "ranch"
+	ProducerKitchen   ProducerType = "kitchen"
+)
+
+// FarmRow is one crop at one base.
+type FarmRow struct {
+	ItemID string
+	Name   string
+
+	// Plants and Biodomes are counts you build, each rounded up.
+	Plants   int64
+	Biodomes int64
+
+	// YieldPerPlant is the source's range. Plants are sized on the
+	// pessimistic bound: a plan that assumes the best case under-builds,
+	// and the failure is a farm that never fills the order.
+	YieldPerPlant Range
+
+	// GrowthSeconds is time to maturity, from the artifact.
+	GrowthSeconds int64
+
+	required *big.Rat
+}
+
+// Required returns the row's requirement as an exact rational.
+func (r FarmRow) Required() *big.Rat { return new(big.Rat).Set(r.required) }
+
+// ExtractorRow is one extracted resource at one base.
+type ExtractorRow struct {
+	ItemID string
+	Name   string
+
+	// Class is the site's, not the row's.
+	// Governing: SPEC-0001 REQ "Producer Rollup" — "Extractor class MUST be
+	// configured per site, not per row."
+	Class HotspotClass
+
+	// Extractors is the count sized so the requirement is produced within
+	// the site's configured fill duration.
+	Extractors int64
+
+	// Depots is ceil(required / depot capacity) above the configured
+	// threshold, and zero below it.
+	Depots int64
+
+	required *big.Rat
+	rate     *big.Rat
+	fill     *big.Rat
+}
+
+// Required returns the row's requirement.
+func (r ExtractorRow) Required() *big.Rat { return new(big.Rat).Set(r.required) }
+
+// RatePerSecond is one extractor's output at the site's class.
+func (r ExtractorRow) RatePerSecond() *big.Rat { return new(big.Rat).Set(r.rate) }
+
+// FillSeconds is how long the built extractors actually take to produce the
+// requirement — the resulting fill time, not the target.
+//
+// It is at most the configured duration, and usually less, because the count
+// was rounded up.
+func (r ExtractorRow) FillSeconds() *big.Rat { return new(big.Rat).Set(r.fill) }
+
+// RanchRow is one fauna product at one base.
+type RanchRow struct {
+	ItemID string
+	Name   string
+
+	// Fauna is the creature count required to yield the requirement within
+	// one collection cycle.
+	Fauna int64
+
+	// CycleSeconds is that cycle's duration.
+	CycleSeconds int64
+
+	required *big.Rat
+}
+
+// Required returns the row's requirement.
+func (r RanchRow) Required() *big.Rat { return new(big.Rat).Set(r.required) }
+
+// KitchenStep is one nutrient processor step.
+type KitchenStep struct {
+	// ItemID is what the step produces, Recipe the route it takes.
+	ItemID string
+	Name   string
+	Recipe string
+
+	// Inputs are the step's ingredients, sorted by item ID.
+	Inputs []KitchenInput
+
+	// ProcessSeconds is the step's duration.
+	ProcessSeconds int64
+
+	// Final marks the step that produces the plan target; every other step
+	// is intermediate.
+	//
+	// Governing: SPEC-0001 REQ "Producer Rollup" — Scenario "Final kitchen
+	// step is distinguished".
+	Final bool
+
+	required *big.Rat
+}
+
+// Required returns how many units this step must produce.
+func (s KitchenStep) Required() *big.Rat { return new(big.Rat).Set(s.required) }
+
+// KitchenInput is one ingredient of a step, with its ratio to the output.
+type KitchenInput struct {
+	ItemID string
+
+	perOutput *big.Rat
+}
+
+// PerOutput is how many of this input one unit of the step's output costs —
+// the input-to-output ratio, exact.
+func (i KitchenInput) PerOutput() *big.Rat { return new(big.Rat).Set(i.perOutput) }
+
+// NoBuild is an item whose demand another producer at the same base already
+// satisfies.
+//
+// Governing: SPEC-0001 REQ "Producer Rollup" — "Items whose demand is
+// satisfied by a byproduct of another producer at the same base MUST be
+// reported as requiring no construction, and MUST NOT contribute a producer
+// count or a power draw."
+type NoBuild struct {
+	ItemID string
+	Name   string
+	// From names the producer whose byproduct covers it.
+	From string
+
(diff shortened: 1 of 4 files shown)


jonstump (commented):

Spec-aware review against issue #40's acceptance criteria and SPEC-0001. Verdict: **approve**. (Comment review — GitHub blocks `APPROVE` on a self-authored PR.)

Conflict-marker gate clean, CI green on all four checks. The largest story in the epic — one requirement, ten scenarios, four producer types — and all ten have named tests.

### The three traps are each caught by a test named after them

Issue #40 flagged three places the obvious implementation goes wrong. Each is handled, and structurally rather than by convention:

- **`PelletFeeders` and `NutrientProcessors` are fields on `BaseBuild`, not on rows.** The type makes per-row counting unexpressible rather than merely discouraged, which is the difference between a rule and a comment. `TestFeederIsReportedOncePerBase` and `TestKitchenSizesProcessorsPerBase` pin it.
- **`TestProcessorsRoundUpFromTheTotal`** is the one that would catch the subtler version: `ceil(total / per)` and the sum of per-row ceilings differ whenever a base carries more than one step, and a test that only checked one step would pass under either.
- **`TestSiteClassAppliesToAllRowsAndOnlyThatBase`** covers both halves of the third trap — that a class change recomputes every row at that base, *and* that it touches no other base.

`TestNoConstantIsHardcoded` and `TestUnclassifiedResourceIsRefused` are the right kind of guard to add beyond the scenario list.

### `ExtractorRate` reads the model correctly

`part.Primary.Rate × ClassStrength(category, class)` is the right reading of ADR-0001's finding — the part declares a rate, the hotspot carries the class. Verified against the artifact: `U_EXTRACTOR_S` rate 100 on Mineral, `U_GASEXTRACTOR` rate 100 on Gas, with Mineral and Gas strengths of 1 / 1.5 / 2 / 2.5.

Worth saying explicitly because **#53 reads the same pattern differently** for `U_GENERATOR_S` — class strength alone, dropping the part rate. I have raised that there rather than here; this PR is the one that has it right, and the two should not land disagreeing.

### The fill-duration unit note is the right kind of honesty

> `U_EXTRACTOR_S` carries rate 100 against storage 360000, which fills in 3,600 of those units — an hour if the rate is per second. Callers therefore supply `SiteConfig.FillSeconds` in the same unit; the arithmetic is consistent either way, and the plan's absolute times are only as right as that reading. Worth confirming in game before any duration is shown to a user.

The arithmetic being unit-agnostic while the *displayed* number is not is exactly the distinction worth recording, and putting it at the function rather than in a TODO means the next person to show a duration finds it.

---

### Note — the feeder condition is stronger than the spec's

REQ "Producer Rollup" says:

> **Where any fauna product assigned to a base requires feeding**, the base MUST report a pellet feeder.

The implementation reports one whenever the base has any ranch row at all:

```go
if len(build.Ranches) > 0 {
    build.PelletFeeders = 1
}
```

There is no "requires feeding" concept anywhere — `Curated.FaunaProducts` is a set of *which items come from creatures*, which is a different question. So the spec's conditional is collapsed into an unconditional, and a base whose only fauna product does not need feeding would be told to build a feeder it does not need.

That may well be right — if every farmable fauna product requires feeding, the two are behaviourally identical. But nothing records that, so a reader cannot tell whether the conditional was considered and found always-true, or simply not modelled. This project has a specific history with exactly that ambiguity: `CAVE2` was modelled as a raw leaf for the same kind of reason, looked right for months, and was wrong.

Either is a fine resolution:

- Model it — a curated classification alongside `FaunaProducts`, in the same spirit as `ResourceHotspots`
- Or state that all farmable fauna require feeding, cite where that was checked, and note that the conditional collapses as a result

The scenario the spec gives ("two distinct fauna products that both require feeding → 1 feeder") passes either way, which is why the test cannot settle it.

### Merge ordering

This PR and #53 both extend `Curated` in `internal/domain/rollup.go`, and both define an identical `ceilRat`. I confirmed git **cannot** auto-merge them — content conflicts in both `rollup.go` and `rollup_test.go`, plus a duplicate-declaration build failure from the two `ceilRat` copies. Whichever lands second needs real conflict resolution and must drop its copy of the helper, as #53's body says.

### Acceptance criteria

All ten scenarios named and passing; producer counts come from injected constants; feeders and processors asserted per base with tests that would fail under per-row summation; class change scoped to one base; byproduct-satisfied items report no construction, no count, no draw.

PR #53, opened 2026-08-20
feat(domain): power generation, draw and the deficit fix

### LINKED ISSUE #41: Power computation: generation, draw and deficit

jonstump:
## What

Implements #41. Each base's generation, draw, and the resulting surplus or deficit. All four Power Computation scenarios have a named test.

## Two generation source types

**Electromagnetic** generators take the Power hotspot's class strength. The spec states EM output as a class-B base times a class multiplier; the artifact states absolute per-class outputs, which is the same model expressed once instead of twice. The scenario's "class-B base of 110 with a class-A multiplier of 1.5" is a class-A strength of 165, and 3 generators is 495 — asserted directly.

**Solar** is classless: the panel's own rate, plus `ceil(panels / panelsPerBattery)` batteries for night coverage. `PanelsPerBattery` joins the curated set — the battery's *capacity* is in the artifact, but how much darkness to plan for is a choice.

## Draw reads both shapes

A part can cost power two ways, and the tests cover each:

- **Directly** — `BIOROOM` has `primary.rate` −20 on the power network
- **Through a dependency** — `U_EXTRACTOR_S` has a `-55` power dependency

Reading only dependencies would report every biodome as free. A part that does neither (a silo) contributes zero.

Draw is a pure function of counts and part costs, so this stage takes **counts** rather than producer rows — which is what keeps it independent of #40, as the issue said it should be.

## A design flaw the tests found

A base with draw and **no generation configured at all** used to fail the whole computation: it hit the deficit branch, found no generator class to size the fix against, and returned an error.

That is the wrong trade. A base you cannot yet cost is not a base you should be unable to *see*. The deficit is now reported with `FixUnsized` set and the count left at zero, so the view can show "you are 60 short" before the user has picked a hotspot class.

## Merge-order note — the two stories are not quite independent

The issue said #40 and #41 could proceed in parallel, and functionally they did. But **both need `ceilRat`**, and I could not import #40's copy from an unmerged branch.

This PR defines it in `power.go`; #52 defines an identical helper in `producer.go`. **Whichever lands second should drop its copy rather than rename it** — one ceiling rule, one implementation. The note is in the code as well as here.

Both PRs also touch `internal/domain/rollup.go` to extend `Curated` (this one adds `PanelsPerBattery`, #52 adds four classification constants), so they are **Tier 1 rather than Tier 0** — the second to merge needs a rebase.

Suggested order: **#52 first** (larger, more constants), then rebase this one and drop the duplicate helper.

## Tests

- EM generation by class, plus the same generators at B to show the class is what moves it
- 5 panels → 3 batteries, classless output, and a base with no panels needing none
- Deficit of 70 against 330 generation → exactly 1 more generator, plus a surplus case reporting no fix
- Downgrade A → C reopens the deficit at 165 and asks for 3
- Direct vs dependent draw, and a part that draws nothing
- Provenance taint per base; determinism over 25 runs with sorted output
- An unknown part fails naming both the part and the base rather than counting as free

Closes #41
Part of #38

Implements SPEC-0001 REQ "Power Computation".

### DIFF
--- a/internal/domain/power.go
+++ b/internal/domain/power.go
@@ -144,13 +144,32 @@ func powerFor(base BaseID, c *Constants, in PowerInput) (PowerBudget, error) {
 		perEM:      new(big.Rat),
 	}
 
-	// Electromagnetic generation: a class multiplier on a base output, both
-	// read from the artifact's Power hotspot.
+	// Electromagnetic generation: the generator's own rate scaled by its
+	// hotspot's class strength.
+	//
+	// Governing: ADR-0001 — "A part declares a base Rate and a
+	// DependsOnHotspots category; the hotspot carries the class." Both
+	// factors are read, the same way Constants.ExtractorRate reads them for
+	// U_EXTRACTOR_S and the solar branch below reads U_SOLAR_S's rate.
+	//
+	// U_GENERATOR_S carries rate 1 in NMS 5.97, so dropping the factor would
+	// agree with this for every value the game currently ships — which is
+	// exactly why it is read rather than assumed. A rate of 1 that is
+	// multiplied in documents itself; a rate of 1 that is ignored is
+	// indistinguishable from a rate nobody checked.
 	if cfg.EMGenerators > 0 {
-		perEM, err := c.ClassStrength("Power", cfg.EMClass)
+		gen, err := c.Part(PartGenerator)
 		if err != nil {
 			return PowerBudget{}, fmt.Errorf("base %q: %w", base, err)
 		}
+		if gen.Primary.Rate <= 0 {
+			return PowerBudget{}, fmt.Errorf("%w: %s states no output", ErrInvalidArtifact, PartGenerator)
+		}
+		strength, err := c.ClassStrength("Power", cfg.EMClass)
+		if err != nil {
+			return PowerBudget{}, fmt.Errorf("base %q: %w", base, err)
+		}
+		perEM := new(big.Rat).Mul(new(big.Rat).SetInt64(gen.Primary.Rate), strength)
 		budget.perEM = perEM
 		budget.generation.Add(budget.generation,
 			new(big.Rat).Mul(perEM, new(big.Rat).SetInt64(cfg.EMGenerators)))
--- a/internal/domain/power.go
+++ b/internal/domain/power.go
@@ -1,0 +1,229 @@
+package domain
+
+import (
+	"fmt"
+	"math/big"
+	"sort"
+)
+
+// Stage 2's power half: what a base generates, what it draws, and what to
+// build when the second exceeds the first.
+//
+// Governing: SPEC-0001 REQ "Power Computation", REQ "Exact Arithmetic and
+// Rounding Discipline", REQ "Provenance Propagation"
+//
+// Draw is a pure function of what is built and what each part draws, so this
+// stage takes counts rather than producer rows. That keeps it independent of
+// the producer stage's row shapes — the same numbers whether they came from
+// a rollup or from a caller sketching a base by hand.
+
+// PowerUnit is a count of one buildable at a base.
+type PowerUnit struct {
+	PartID string
+	Count  int64
+}
+
+// PowerConfig is a base's generation setup.
+type PowerConfig struct {
+	// EMGenerators sit on a power hotspot and take its class; SolarPanels
+	// are classless and need batteries for the night.
+	//
+	// Governing: SPEC-0001 REQ "Power Computation" — "Generation MUST
+	// support two source types."
+	EMGenerators int64
+	EMClass      HotspotClass
+	SolarPanels  int64
+}
+
+// PowerBudget is one base's power position.
+type PowerBudget struct {
+	Base BaseID
+
+	// Batteries are required for solar night coverage, at the configured
+	// ratio, rounded up.
+	Batteries int64
+
+	// AdditionalGenerators is how many more electromagnetic generators at
+	// the base's class would clear a deficit — zero when there is none.
+	//
+	// Reported rather than left to the view so a deficit presents as an
+	// action rather than a warning.
+	// Governing: SPEC-0001 REQ "Power Computation" — Scenario "Deficit
+	// reports the fix".
+	AdditionalGenerators int64
+
+	// FixUnsized is true when the base is in deficit but no generator class
+	// is configured, so the fix has no size. The deficit is still reported
+	// — a base you cannot yet cost is not a base you should be unable to
+	// see.
+	FixUnsized bool
+
+	// Verified is false when any contributing figure is.
+	// Governing: SPEC-0001 REQ "Provenance Propagation".
+	Verified bool
+
+	generation *big.Rat
+	draw       *big.Rat
+	perEM      *big.Rat
+}
+
+// Generation is the base's total output, exact.
+func (b PowerBudget) Generation() *big.Rat { return new(big.Rat).Set(b.generation) }
+
+// Draw is the base's total consumption, exact and positive.
+func (b PowerBudget) Draw() *big.Rat { return new(big.Rat).Set(b.draw) }
+
+// Balance is generation minus draw: positive is surplus, negative deficit.
+func (b PowerBudget) Balance() *big.Rat { return new(big.Rat).Sub(b.generation, b.draw) }
+
+// Deficit is how much the base is short, or zero when it is not.
+func (b PowerBudget) Deficit() *big.Rat {
+	balance := b.Balance()
+	if balance.Sign() >= 0 {
+		return new(big.Rat)
+	}
+	return balance.Neg(balance)
+}
+
+// InDeficit reports whether draw exceeds generation.
+func (b PowerBudget) InDeficit() bool { return b.Balance().Sign() < 0 }
+
+// PerGenerator is one electromagnetic generator's output at the base's class.
+func (b PowerBudget) PerGenerator() *big.Rat { return new(big.Rat).Set(b.perEM) }
+
+// PowerInput is the power stage's per-base input.
+type PowerInput struct {
+	// Config is each base's generation setup.
+	Config map[BaseID]PowerConfig
+
+	// Draws are the buildables at each base that consume power, as counts
+	// by part ID. Each part's draw comes from the artifact.
+	Draws map[BaseID][]PowerUnit
+
+	// Unverified marks bases whose contributing figures are not verified,
+	// so the budget carries the taint forward.
+	Unverified map[BaseID]bool
+}
+
+// ComputePower produces each base's power budget.
+//
+// Governing: SPEC-0001 REQ "Power Computation"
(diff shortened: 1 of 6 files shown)


jonstump (commented):
## Review: Changes requested

Spec-aware review against issue #41's acceptance criteria, SPEC-0001, and ADR-0001. (Comment review — GitHub blocks `REQUEST_CHANGES` on a self-authored PR. Treat as changes-requested.)

Conflict-marker gate clean, CI green on all four checks.

The four Power Computation scenarios each have a named test and pass. One finding, and it is about the model rather than the arithmetic.

---

### EM generation drops the generator's own rate

`power.go` computes electromagnetic generation as class strength times count:

```go
perEM, err := c.ClassStrength("Power", cfg.EMClass)
budget.generation.Add(budget.generation, new(big.Rat).Mul(perEM, ...EMGenerators))
```

`U_GENERATOR_S.Primary.Rate` is never read. #52 models the identical source pattern the other way — `ExtractorRate` is `part.Primary.Rate` **times** `ClassStrength`:

```go
return new(big.Rat).Mul(new(big.Rat).SetInt64(p.Primary.Rate), strength), nil
```

Same table, same `DependsOnHotspots` shape, two different readings. ADR-0001's finding describes one model: *"A part declares a base `Rate` and a `DependsOnHotspots` category; the hotspot carries the class."* Measured against the committed artifact, the parts line up exactly that way:

| Part | `primary.rate` | hotspot |
|---|---|---|
| `U_EXTRACTOR_S` | 100 | Mineral |
| `U_GASEXTRACTOR` | 100 | Gas |
| `U_GENERATOR_S` | **1** | Power |
| `U_SOLAR_S` | 50 | — (classless, as the spec says) |

**The two readings agree today only because the generator's rate is 1.** At class B, `1 × 220` and `220` are the same number. They stop being the same the moment a game update changes that rate, or a second EM generator part with a different rate is modelled — and the failure would be silent, because nothing asserts the factor is being read.

That `U_SOLAR_S` carries rate 50 and no hotspot, exactly matching the spec's "solar panels MUST be classless", is good evidence the rate field is meaningful on these parts rather than vestigial.

Two ways to close it, and either is fine — what is not fine is the two stages disagreeing:

- Read the rate, matching #52: `U_GENERATOR_S.Primary.Rate × ClassStrength("Power", class)`. A rate of 1 then documents itself.
- Or establish that the generator's rate is genuinely not a factor, and say so where it is dropped — with the same standard of evidence #52 applied to the extractor.

The PR body says the artifact "states absolute per-class outputs, which is the same model expressed once instead of twice". Against `U_GENERATOR_S` that reading is indistinguishable from the multiplicative one, so it is not yet established — it is the hypothesis that happens to fit a rate of 1.

---

### The `FixUnsized` change is the right call

A base with draw and no generation configured previously failed the whole computation. Reporting the deficit with the fix left unsized is better: *"A base you cannot yet cost is not a base you should be unable to see."* Failing a whole plan because one base has no hotspot class picked yet would make the engine unusable during exactly the exploratory phase the planner exists for.

### Draw reads both shapes, and the test proves it matters

`BIOROOM` costs power through `primary.rate` −20; `U_EXTRACTOR_S` costs it through a −55 dependency. Reading only dependencies would report every biodome as free. Covering both, with a part that does neither contributing zero, is the right shape — and it is the kind of thing that is invisible until a plan quietly under-reports draw.

### Merge ordering — confirmed worse than the PR body says

The body notes both PRs touch `rollup.go` and that the second needs a rebase. I verified it, and git cannot auto-merge at all:

```
CONFLICT (content): Merge conflict in internal/domain/rollup_test.go
internal/domain/rollup.go: syntax error: unexpected <<
```

Conflicts in **both** `rollup.go` and `rollup_test.go` — the `Curated` struct edits collide directly, not just the `ceilRat` duplicate. So the second to land needs genuine conflict resolution, and flagging the shared helper was right: `ceilRat` is byte-identical in `producer.go` and `power.go`, same package, so keeping both is a duplicate-declaration build failure rather than a style issue.

Calling that out in the PR body rather than leaving it for whoever merges second is the correct handling of a constraint you cannot fix from inside one branch.

### Acceptance criteria

| Criterion (issue #41) | Status |
|---|---|
| Four scenarios pass as named tests | Met |
| Generation for both source types, solar classless | Met — but see the finding on how EM output is derived |
| Battery count rounds up | Met — `ceil(panels / panelsPerBattery)` |
| Deficit reports both shortfall and additional units | Met, plus the unsized case |
| Class change recomputes; downgrade into deficit reports the fix | Met |
| No power constant hardcoded | Met |
| Mixed-power-grid open question resolved before implementation | **Not addressed** — see below |

The issue made design.md's open question a prerequisite: *"The base-planner handoff models one power type per base and notes production likely needs a generator list instead of a type toggle. Revisit before implementing stage 3 in earnest."* This PR implements one EM class plus solar per base — the mock's model — without recording that the question was considered and settled that way. Worth a line in the design doc either way, since the shape of `PowerBudget` is what SPEC-0002's boundary will carry.

jonstump:
## Review feedback addressed — generation model fixed in 4a96e29

EM generation now reads both factors:

```go
gen, err := c.Part(PartGenerator)
if gen.Primary.Rate <= 0 {
    return PowerBudget{}, fmt.Errorf("%w: %s states no output", ErrInvalidArtifact, PartGenerator)
}
strength, err := c.ClassStrength("Power", cfg.EMClass)
perEM := new(big.Rat).Mul(new(big.Rat).SetInt64(gen.Primary.Rate), strength)
```

That matches `Constants.ExtractorRate` and, more to the point, matches **the solar branch eight lines below it in this same file**, which already read and validated `U_SOLAR_S.Primary.Rate`. The generator was the one place in the engine reading this table a different way, and the inconsistency was internal to `power.go` rather than only across PRs.

### Why it was worth fixing rather than noting

`U_GENERATOR_S` carries rate 1, so `1 × strength` and `strength` are the same number for every value NMS 5.97 ships. The old code was not producing a wrong figure — it was producing the right figure for a reason that did not hold. A factor that stays invisible until the data changes is a factor nobody checked, and a rate of 1 that is multiplied in documents itself.

### A second gap the fix closed

A generator with `rate: 0` previously produced generation from the class strength alone — output from a part stating it has none. It is now refused with `ErrInvalidArtifact` naming the part, which is the standard the solar branch already applied.

### Tests

**`TestGeneratorRateScalesGeneration`** uses a fixture where the generator's rate is **3**, precisely so the factor cannot hide. Class B strength 110, two generators:

- reading both factors: `3 × 110 × 2` = 660
- dropping the rate: `110 × 2` = 220

**`TestGeneratorWithNoStatedOutputIsRefused`** covers the zero-rate case.

Both **fail against the previous implementation**:

```
power_test.go:374: generation = 220, want 660 — the generator's own rate is not being read
power_test.go:400: Power accepted a generator stating no output
```

The existing fixture keeps rate 1, so **every pre-existing test passes unmodified** — including all four Power Computation scenarios. That is the point worth noting: the old tests could not have caught this, because the fixture was built from the same assumption as the code.

### Verification

| Check | Result |
|---|---|
| `go test -race ./...` | all four packages ok |
| Four Power Computation scenarios | pass, unmodified |
| `go vet ./...` | clean |
| `gofmt -l .` | no output |

### Unchanged, and still open

The two review notes I have not acted on, since neither is mine to settle from inside this branch:

- **Merge ordering.** Still Tier 1 with #52 — I confirmed git cannot auto-merge them: content conflicts in both `rollup.go` and `rollup_test.go`, plus the duplicate `ceilRat`. Whichever lands second needs real conflict resolution and must drop its copy of the helper.
- **The mixed-power-grid open question.** design.md says "Revisit before implementing stage 3 in earnest", and this PR implements the mock's single-type-per-base model without recording that the question was considered. Worth a line in the design doc either way, because `PowerBudget`'s shape is what SPEC-0002's boundary will carry.

PR #54, opened 2026-08-20
feat(bridge): stable error codes and the failure contract

### LINKED ISSUE #44: Error codes and the failure contract

jonstump:
## What

Implements #44. Every domain sentinel maps to a stable identifier the view switches on; the Go message crosses alongside as diagnostic text with no contractual shape. All six scenarios have named tests, still under plain `go test`.

## The coupling this removes

`errors.Is` does not survive a boundary crossing, and matching on message text breaks the day someone improves a message. `CodeFor` matches by `errors.Is` **alone** and reads no text — so the reworded-message requirement holds by construction rather than by discipline.

The test pins it both ways:

1. It rewords a wrapped message (asserting first that the two strings genuinely differ) and checks the code is unchanged.
2. It parses `CodeFor`'s own AST and fails on any `strings.*` call or `.Error()` inside it — so a future "helpful" text check fails the test rather than passing review.

## The unclassified code is the point, not a fallback

Silently mapping an unrecognized error onto the nearest sentinel makes the view branch **confidently on a wrong kind** — worse than admitting ignorance. The test feeds errors deliberately worded to look like sentinels they do not wrap:

```go
errors.New("unknown item: this message resembles a sentinel but wraps none")
errors.New("cycle detected somewhere, allegedly")
errors.New("invalid tier 1 artifact")
```

Each must cross as `UNCLASSIFIED`.

## Exhaustiveness is checked against the source

`TestEverySentinelHasADistinctCode` parses `internal/domain/errors.go` and asserts every `Err*` it declares has a code. **A sentinel added to the domain without a code fails this test** rather than quietly crossing as unclassified — which is the failure mode a hand-written list in the test would have allowed.

It also asserts no two sentinels share a code, and that none maps to the reserved one.

## Three boundary-native sentinels

The module fails in ways the domain has no vocabulary for, because the domain never had a lifecycle or a caller across a wire: `ErrNotReady`, `ErrMalformedInput`, `ErrVersionMismatch`.

They are matched **after** the domain's, so a domain error wrapped by one of ours keeps the more specific code. There is a test for that: `fmt.Errorf("%w: %w", ErrMalformedInput, domain.ErrIllegalMethod)` crosses as `ILLEGAL_METHOD`.

`ErrNotReady` is defined here with its siblings and returned by #45 — the same split `hgpak` uses for `ErrUnsafePath`.

**Malformed input is deliberately distinct from an invalid artifact.** The artifact is ours; the input is theirs. Collapsing the two sends a user hunting through their own payload for our bug, or the reverse.

## Decode failures

`DecodePlanStrict` returns a **zero** `PlanInput` on any failure, so a caller cannot accidentally compute against a half-decoded plan — asserted in every one of the six malformed cases (no target, non-numeric quantity, fractional quantity, zero, negative, and a method outside the vocabulary).

## Note

The worktree was created under the issue title rather than the `### Branch` name; I corrected the branch to `feat/44-error-codes-and-failure-contract` as the issue specifies.

Closes #44
Part of #42

Implements SPEC-0002 REQ "Sentinel Error Preservation", REQ "Contract Versioning", REQ "Error Handling Standards".

### DIFF
--- a/internal/bridge/bridge.go
+++ b/internal/bridge/bridge.go
@@ -96,20 +96,14 @@ type ResultPayload struct {
 
 // ErrorPayload is the failure half.
 //
-// Code is a stable machine-readable identifier; the sentinel mapping that
-// fills it is #44's story, and until then everything crosses as
-// CodeUnclassified. Message carries the domain's prose, including the
-// wrapped resolution path, with no contractual guarantee of format.
+// Code is a stable machine-readable identifier from the set in errors.go;
+// see CodeFor for the mapping. Message carries the domain's prose,
+// including the wrapped resolution path, with no contractual guarantee of
+// format.
 type ErrorPayload struct {
 	Code    string `json:"code"`
 	Message string `json:"message"`
 }
-
-// CodeUnclassified is the code reserved for a failure matching no known
-// sentinel. SPEC-0002 REQ "Sentinel Error Preservation" requires the
-// reserved code exist so an unmatched error is never silently mapped onto
-// an unrelated sentinel; the rest of the code set arrives with #44.
-const CodeUnclassified = "UNCLASSIFIED"
 
 // Success builds an envelope carrying a payload and no error.
 func Success(data ResultPayload) Envelope {
--- a/internal/bridge/errors.go
+++ b/internal/bridge/errors.go
@@ -1,0 +1,171 @@
+package bridge
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/jonstump/nms-base-planner/internal/domain"
+)
+
+// The failure half of the boundary contract.
+//
+// Governing: SPEC-0002 REQ "Sentinel Error Preservation", REQ "Contract
+// Versioning", REQ "Error Handling Standards"
+//
+// errors.Is does not survive a boundary crossing, and matching on message
+// text is the classic brittle coupling: an improved error message silently
+// breaks a consumer's branch. So each domain sentinel maps to a stable
+// identifier the view switches on, and the Go message crosses alongside as
+// diagnostic text with no contractual shape.
+
+// Codes carried in an ErrorPayload.
+//
+// These are contract, not implementation detail: changing one is a contract
+// version change per SPEC-0002 REQ "Contract Versioning". They are
+// deliberately independent of the Go error text, so rewording a message
+// never moves a code.
+const (
+	// From internal/domain's sentinel vocabulary.
+	CodeUnknownItem     = "UNKNOWN_ITEM"
+	CodeIllegalMethod   = "ILLEGAL_METHOD"
+	CodeCycleDetected   = "CYCLE_DETECTED"
+	CodeMissingConstant = "MISSING_CONSTANT"
+	CodeInvalidArtifact = "INVALID_ARTIFACT"
+
+	// Added by the boundary itself. The module can fail in ways the domain
+	// has no vocabulary for, because the domain never had a lifecycle or a
+	// caller on the other side of a wire.
+	CodeNotReady        = "NOT_READY"
+	CodeMalformedInput  = "MALFORMED_INPUT"
+	CodeVersionMismatch = "VERSION_MISMATCH"
+
+	// CodeUnclassified is reserved for an error matching no sentinel.
+	//
+	// The point, not a fallback: silently mapping an unrecognized error
+	// onto the nearest sentinel makes the view branch confidently on a
+	// wrong kind, which is worse than admitting ignorance.
+	CodeUnclassified = "UNCLASSIFIED"
+)
+
+// sentinelCodes is the mapping, in match order.
+//
+// Ordered rather than a map because errors.Is is a predicate: an error can
+// satisfy more than one sentinel if a future one wraps another, and a map's
+// iteration order would make which code won depend on the run.
+var sentinelCodes = []struct {
+	sentinel error
+	code     string
+}{
+	{domain.ErrUnknownItem, CodeUnknownItem},
+	{domain.ErrIllegalMethod, CodeIllegalMethod},
+	{domain.ErrCycleDetected, CodeCycleDetected},
+	{domain.ErrMissingConstant, CodeMissingConstant},
+	{domain.ErrInvalidArtifact, CodeInvalidArtifact},
+
+	// Boundary-native failures, matched after the domain's so a wrapped
+	// domain error is never shadowed by one of ours.
+	{ErrNotReady, CodeNotReady},
+	{ErrMalformedInput, CodeMalformedInput},
+	{ErrVersionMismatch, CodeVersionMismatch},
+}
+
+// CodeFor returns the stable code for an error.
+//
+// Governing: SPEC-0002 REQ "Sentinel Error Preservation" — "Codes MUST be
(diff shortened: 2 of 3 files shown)


jonstump (commented):

Spec-aware review against issue #44's acceptance criteria and SPEC-0002. Verdict: **approve**. (Comment review — GitHub blocks `APPROVE` on a self-authored PR.)

Conflict-marker gate clean, CI green on all four checks. All six scenarios have named tests, plus three guards beyond the list.

### The coupling test earns its keep

`TestRewordingAMessageDoesNotChangeTheCode` is the one that only pays off later — it passes trivially the day it is written, and becomes load-bearing the first time someone improves an error message. Writing it as an explicit reword-then-assert rather than trusting the design is what the requirement is actually asking for: *"a stable code lets the view branch on failure kind while the message stays free to improve."*

`TestUnmatchedErrorCrossesAsUnclassified` covers the scenario most likely to be got wrong by being helpful. Mapping an unrecognised error onto the nearest sentinel makes the view branch confidently on a wrong kind, which is worse than admitting ignorance — and it is exactly the kind of "reasonable" fallback a later change might introduce.

### Two guards beyond the scenario list, both worth having

**`TestDomainCodeIsNotShadowedByABoundarySentinel`** — the boundary defines its own codes (not-ready, version mismatch, malformed input) alongside the domain's, and nothing structurally stops one masking the other. Testing that a domain error still crosses with its own code is the sort of interaction bug that only appears once both vocabularies exist.

**`TestFailureFromNilIsHonest`** — `Failure(CodeUnclassified, "an error envelope was requested with no error")` is the right handling. Fabricating a plausible failure, or returning a success envelope, would both be worse than saying what happened.

### Not-ready defined here, returned by #45

Defining the not-ready code in this vocabulary while the lifecycle story returns it mirrors how `hgpak` handled `ErrUnsafePath` — defined with the sentinel set, returned by the extraction layer. Keeping one vocabulary in one place is right, and the sixth scenario ("not ready versus malformed carry different codes") is testable here without the module existing.

---

### Note — exhaustiveness is asserted against a hand-maintained list

Issue #44's criterion reads: *"the mapping is exhaustive over the sentinels `internal/domain` exports"*. `TestEverySentinelHasADistinctCode` names five:

```go
"ErrUnknownItem":     {domain.ErrUnknownItem, bridge.CodeUnknownItem},
"ErrIllegalMethod":   {domain.ErrIllegalMethod, bridge.CodeIllegalMethod},
...
```

`internal/domain` exports exactly those five today, so the mapping **is** exhaustive and the test is accurate. What it does not do is *detect* a sixth. A new domain sentinel would cross as `UNCLASSIFIED`, every test here would still pass, and the view would branch on unclassified for a failure kind that has a name.

That degrades safely — unclassified is the honest answer, not a wrong code — which is why it would be quiet rather than loud. But it makes "exhaustive" a property that was true when written rather than one the suite maintains.

**#50 set the bar for this in the same package**: `TestNoFloatInTheEncodingPath` parses the package's own source with `go/ast` rather than trusting a list, and I confirmed it catches an injected conversion. The same technique applies directly — parse `internal/domain/errors.go` for its `Err*` declarations and assert each appears in the mapping. That turns the criterion into something the suite enforces rather than something a reviewer has to re-check whenever `domain` grows.

Not blocking: the mapping is correct today, and the failure mode is a safe degradation rather than a wrong answer.

### Acceptance criteria

| Criterion (issue #44) | Status |
|---|---|
| Six scenarios pass as named tests under plain `go test` | Met |
| Every domain sentinel maps to a distinct code; mapping exhaustive | Met today — see the note on how it is asserted |
| Unmatched error crosses as unclassified, never a wrong sentinel | Met |
| Rewording a message does not change the code | Met — proven by a test that rewords |
| Version mismatch names both versions and refuses the payload | Met |
| Malformed-input and not-ready carry different codes | Met |

### Merge ordering

`internal/bridge` only — no overlap with #52 or #53, which both touch `internal/domain/rollup.go` and conflict with each other. This one is Tier 0 and can merge in any order relative to them.

PR #55, opened 2026-08-20
feat(bridge): module lifecycle, boundary surface, and mechanical purity

### LINKED ISSUE #45: Module lifecycle, surface, and domain purity

jonstump:
## What

Implements #45, the last story in the WASM epic. `cmd/planner` is the only place in the tree that touches `syscall/js`, and it does one thing: hand strings between JavaScript and `bridge.Module`. Everything the boundary *decides* — what the entry points are, when they are callable, what comes back — lives in `internal/bridge` and tests under plain `go test`.

44 passing test cases. The WASM half builds and vets under `GOOS=js GOARCH=wasm`.

## Two things the tests found

**1. A load failure was classified by the wrong context.** A recipe naming an item no table defines is wrapped as *both* `ErrInvalidArtifact` and `ErrUnknownItem`, and `CodeFor` returned the more specific one.

That is right when a **plan** names a missing item — the caller's problem. It is wrong when loading **our own artifact**: a view told `UNKNOWN_ITEM` sends the user looking at their plan for our data bug. `Load` now classifies by context, because anything failing there is our artifact being inconsistent whatever else it wraps.

**2. My first purity check proved nothing.** The ADR-0003 note suggests `go list -deps ./internal/domain | grep syscall/js`, so that is what I wrote. **Every package fails it** — under `GOOS=js` the standard library's own `syscall` package imports `syscall/js`, so it is in every transitive closure.

The check now compares **direct imports**, which is what the ADR actually forbids, and asserts that `cmd/planner` *does* import it — so a pass is the absence of a dependency that exists elsewhere in the module, not a query that never finds anything. (Belt and braces: importing it into a non-adapter package also breaks the host build outright.)

## Event-loop safety: the operational definition the story asked for

The issue flagged this as "the requirement with the least specified acceptance — worth settling what counts as a pass rather than shipping a test that asserts nothing". Two parts:

**Bounded.** A resolve on the largest supported plan returns in **under 50 ms** — three frames at 60 Hz, which is where "unresponsive" starts to mean something. Measured in Go, because the work being timed is Go's. It runs in **141 µs** at quantity 1,000,000.

**Non-blocking.** No entry point may contain a channel operation, `select`, goroutine, sleep, or wait — checked by parsing the call paths. Deadlocking against a single-threaded event loop requires *waiting inside a call it is making*, so this is a property of the code rather than of a run. The `select {}` that parks `main` is outside every entry point, which is why it is exempt and why the check is scoped to the call paths.

## Surface

One namespace (`nmsPlanner`), five entry points enumerated in `bridge.EntryPoints` so the surface is listable from a plain test rather than scattered through registration code.

`rollup` and `power` are declared **now** with `resolve`'s signature, returning a well-formed not-ready envelope. RESERVED means the shape is fixed before there is an implementation to bend it around — and a reserved name returning `undefined` would be indistinguishable from a typo on the view's side.

## Tests worth calling out

- **Not-ready does not hang**: the test races the call against a 5-second timeout, so a deadlock fails by timing out rather than being assumed away.
- **One crossing**: resolving the 34-node plan returns everything in a single value, and the test walks every node asserting no field is missing that would force a second call.
- **Adapter holds no domain logic**: parses the adapter's AST and fails on `Add`/`Sub`/`Mul`/`Quo`/`QuoRem`/`Ceil`/`Floor`/`FloatString`. `design.md` names the concrete temptation — rounding a rational to something display-friendly on the way out — and SPEC-0001 says none of its rounding boundaries is "on the way to the view".
- **Encoding paths need no build tag**: fails if any non-test adapter source acquires one, which would move that behaviour out of reach of plain `go test`.
- **A failed load does not poison the module**: a valid artifact still loads afterwards.

`go vet` (host and WASM), `staticcheck v0.8.0-rc.1`, `gofmt -l` clean.

Closes #45
Part of #42

Implements SPEC-0002 REQ "Boundary Surface", REQ "Module Lifecycle and Readiness", REQ "Event Loop Safety", REQ "Domain Purity Preservation".

### DIFF
--- a/cmd/planner/main.go
+++ b/cmd/planner/main.go
@@ -1,0 +1,61 @@
+//go:build js && wasm
+
+// Command planner is the WASM entry point: the only place in the tree that
+// touches syscall/js.
+//
+// Governing: ADR-0003 (Go domain, thin adapter), SPEC-0002 REQ "Boundary
+// Surface", REQ "Module Lifecycle and Readiness", REQ "Event Loop Safety"
+//
+// It does one thing: hand strings between JavaScript and bridge.Module. Every
+// decision about what the entry points are, when they are callable, and what
+// comes back lives in internal/bridge, which builds and tests without WASM.
+// This file is deliberately too small to hide a bug in.
+package main
+
+import (
+	"syscall/js"
+
+	"github.com/jonstump/nms-base-planner/internal/bridge"
+)
+
+func main() {
+	module := bridge.NewModule()
+
+	// One namespace object carrying the stage entry points. No domain
+	// function is reachable from the global scope: the only exported names
+	// are these, and each returns a complete envelope.
+	//
+	// Governing: SPEC-0002 REQ "Boundary Surface".
+	namespace := js.Global().Get("Object").New()
+	for _, name := range bridge.EntryPoints {
+		namespace.Set(name, entryPoint(module, name))
+	}
+	namespace.Set("contractVersion", bridge.ContractVersion)
+	js.Global().Set(bridge.Namespace, namespace)
+
+	// Park forever. Returning from main tears the Go runtime down and every
+	// registered function with it.
+	//
+	// This is the only blocking construct in the module, and it is outside
+	// every entry point: the calls themselves are synchronous and
+	// straight-line, so nothing can deadlock against the event loop by
+	// waiting on it from inside a call.
+	// Governing: SPEC-0002 REQ "Event Loop Safety".
+	select {}
+}
+
+// entryPoint wraps one named call.
+//
+// Synchronous and straight-line by construction: it reads one string
+// argument, dispatches, and returns one string. No goroutine, no channel, no
+// callback into JavaScript — so a call can neither block the event loop
+// waiting for JavaScript nor deadlock against the Go scheduler.
+func entryPoint(module *bridge.Module, name string) js.Func {
+	return js.FuncOf(func(_ js.Value, args []js.Value) any {
+		var arg string
+		if len(args) > 0 && args[0].Type() == js.TypeString {
+			arg = args[0].String()
+		}
+		return module.CallJSON(name, arg)
+	})
+}
--- a/internal/bridge/module.go
+++ b/internal/bridge/module.go
@@ -1,0 +1,186 @@
+package bridge
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/jonstump/nms-base-planner/internal/domain"
+)
+
+// The module's surface and lifecycle, in a form that needs no WASM build.
+//
+// Governing: ADR-0003 (Go domain, thin adapter), SPEC-0002 REQ "Boundary
+// Surface", REQ "Module Lifecycle and Readiness", REQ "Event Loop Safety"
+//
+// Everything the boundary decides lives here: what the entry points are,
+// when they are callable, and what comes back. The syscall/js shim in
+// cmd/planner does nothing but hand strings across, so the interesting
+// behaviour is testable under plain go test and the js half stays too small
+// to hide a bug in.
+
+// Module is the boundary's stateful half: it holds the loaded artifact and
+// gates the entry points on it.
+//
+// Instantiation and artifact loading are distinct steps, per SPEC-0002 REQ
+// "Module Lifecycle and Readiness" — a module that exists but has no data is
+// a different failure from a module that failed to load, and conflating them
+// sends a user to the wrong place.
+type Module struct {
+	mu       sync.RWMutex
+	artifact *domain.Tier1
+}
+
+// NewModule creates an unloaded module. It is not ready until Load succeeds.
+func NewModule() *Module { return &Module{} }
+
+// Ready reports whether an artifact has been loaded.
+func (m *Module) Ready() bool {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+	return m.artifact != nil
+}
+
+// Load validates a Tier 1 artifact and makes the module ready.
+//
+// A validation failure is reported as an invalid artifact rather than as a
+// module load failure: the module instantiated fine, and the data is what is
+// wrong.
+//
+// Governing: SPEC-0002 REQ "Module Lifecycle and Readiness" — Scenario "A
+// bad artifact is not a module failure".
+func (m *Module) Load(artifactJSON string) Envelope {
+	a1, err := domain.LoadTier1(strings.NewReader(artifactJSON))
+	if err != nil {
+		// Classified as an artifact problem by *context*, not by whichever
+		// sentinel the error happens to wrap first.
+		//
+		// The domain double-wraps some validation failures — a recipe
+		// naming an item no table defines is both ErrInvalidArtifact and
+		// ErrUnknownItem — and CodeFor would return the more specific one.
+		// That is right when a *plan* names a missing item, which is the
+		// caller's problem. It is wrong here: anything that fails while
+		// loading our own artifact is our data being inconsistent, and a
+		// view told UNKNOWN_ITEM would send the user looking at their plan.
+		//
+		// Governing: SPEC-0002 REQ "Module Lifecycle and Readiness" — "the
+		// failure is reported as an invalid-artifact error".
+		return Failure(CodeInvalidArtifact, err.Error())
+	}
+	m.mu.Lock()
+	m.artifact = a1
+	m.mu.Unlock()
+	return Success(ResultPayload{})
+}
+
+// Resolve runs stage 1 and returns one envelope carrying the whole graph.
+//
+// Governing: SPEC-0002 REQ "Boundary Surface" — "coarse-grained so one call
+// performs one complete stage". The view gets every node from this single
+// value; there is no per-node accessor to cross the boundary again for.
+func (m *Module) Resolve(planJSON string) Envelope {
+	artifact, ok := m.loaded()
+	if !ok {
+		return FailureFrom(fmt.Errorf("resolve: %w", ErrNotReady))
+	}
+
+	var plan Plan
+	if err := json.Unmarshal([]byte(planJSON), &plan); err != nil {
+		return FailureFrom(fmt.Errorf("%w: %v", ErrMalformedInput, err))
+	}
+	in, err := DecodePlanStrict(plan)
+	if err != nil {
+		return FailureFrom(err)
+	}
+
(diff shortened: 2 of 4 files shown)


jonstump (commented):
## Review: Approved

Spec-aware review against issue #45's acceptance criteria, SPEC-0002, and ADR-0003. Verdict: **approve**. (Comment review — GitHub blocks `APPROVE` on a self-authored PR.)

Conflict-marker gate clean, CI green. Verified locally on `b99c297`: `go test -race ./...` passes across all four packages, 44 test cases in `internal/bridge`, `GOOS=js GOARCH=wasm go build ./cmd/planner` produces a 3.7 MB module, and `GOOS=js GOARCH=wasm go vet ./...` is clean.

No findings. Three things I checked rather than took on trust, and one of them corrects something I got wrong.

---

### The purity finding is right, and it invalidates a check I ran in an earlier review

> My first purity check proved nothing. The ADR-0003 note suggests `go list -deps ./internal/domain | grep syscall/js` … **Every package fails it** — under `GOOS=js` the standard library's own `syscall` package imports `syscall/js`.

Verified:

| Query | Result |
|---|---|
| host: `go list -deps ./internal/domain \| grep -c syscall/js` | **0** |
| `GOOS=js`: same query | **1** |
| `GOOS=js`: `go list -deps ./internal/hgpak` | **1** |

`internal/hgpak` is a zstd archive reader. Under the build the rule exists to police, the ADR's suggested check flags it as a `syscall/js` dependant — so it distinguishes nothing.

**I reported that check as evidence in my review of #50**, where I ran it on the host, got zero, and wrote that the package "does NOT depend on `syscall/js`". That was true of the host build and useless as a statement about the WASM one. Worth recording plainly: the check I offered as verification was the check this PR found to be vacuous.

The replacement compares **direct** imports under `GOOS=js`, which is what ADR-0003 actually forbids. I tested whether it has teeth by adding a `//go:build js && wasm` file importing `syscall/js` into `internal/domain` — a violation the host build cannot see, since `syscall/js` does not exist there:

```
purity_test.go:56: github.com/jonstump/nms-base-planner/internal/domain imports syscall/js;
                   ADR-0003 confines it to the adapter
--- FAIL: TestSyscallJSIsReachableOnlyThroughTheAdapter
```

It fires. Two details make it more than a grep: the `len(imports) < 5` guard refuses to pass when `go list` returns too little to be meaningful, and asserting that `cmd/planner` **does** import it means a pass is the absence of a dependency that demonstrably exists elsewhere in the module — not a query that never finds anything. That is the same standard `TestNoFloatInTheEncodingPath` set in #50, applied to a rule whose canonical formulation turned out to be broken.

### Event-loop safety now has the operational definition it was missing

Issue #45 flagged this as the weakest requirement in SPEC-0002 — one scenario, no numbers, and "does not become unresponsive" easy to satisfy with a test that asserts nothing. The answer splits it correctly:

**Bounded** — under 50 ms for the largest supported plan, justified as three frames at 60 Hz, which is where "unresponsive" starts to mean something rather than being a number picked to pass. Measured on my machine at **129.794 µs** at quantity 1,000,000, roughly 400× inside budget.

**Non-blocking** — no channel op, `select`, goroutine, sleep or wait inside any entry point's call paths, checked by parsing. The reasoning is the part that makes it right: deadlocking a single-threaded event loop requires *waiting inside a call it is making*, which is a property of the code rather than of a run — so a static check is the stronger instrument here, not a weaker substitute for a timing test. Scoping it to call paths, so the `select {}` parking `main` is exempt by construction rather than by exception, is the correct boundary.

### The load-classification finding is subtle and correctly resolved

A recipe naming an item no table defines wraps as both `ErrInvalidArtifact` and `ErrUnknownItem`, and `CodeFor` returned the more specific. Right when a **plan** names a missing item — the caller's problem. Wrong when loading **our own artifact**, where `UNKNOWN_ITEM` sends the user hunting through their plan for our data bug.

Classifying by context at `Load` rather than reordering the sentinel precedence is the right fix: the same error genuinely means different things depending on who supplied the input, and precedence has no way to express that.

### Surface

`EntryPoints = []string{"load", "ready", "resolve", "rollup", "power"}` under one `nmsPlanner` namespace, enumerated so the surface is listable from a plain test rather than inferred from registration code.

Declaring `rollup` and `power` **now**, with `resolve`'s signature and a well-formed not-ready envelope, is the right reading of RESERVED — and the justification is better than the requirement's: *"a reserved name returning `undefined` would be indistinguishable from a typo on the view's side"*. That is a real failure mode the spec does not name.

Now that SPEC-0001 stages 2 and 3 are merged, those two are a wiring job into a contract that already fits them, which is what REQ "Boundary Surface"'s third scenario was protecting.

### Acceptance criteria

| Criterion (issue #45) | Status |
|---|---|
| Nine scenarios pass; encoding under plain `go test`, registration/lifecycle under a WASM build | Met — 44 cases, WASM build and vet verified |
| Exactly one namespace; no domain function reachable globally | Met |
| 34-node plan crosses exactly once | Met — the test walks every node asserting no field would force a second call |
| Not-ready returns a failure envelope, does not hang or return undefined | Met — raced against a 5-second timeout, so a deadlock fails by timing out |
| Invalid artifact reported as such, not as module-load failure | Met, and sharpened by the classification finding |
| `syscall/js` reachable only through the adapter, enforced mechanically | Met — teeth confirmed above |
| Adapter holds no quantity arithmetic, traversal, or provenance rules | Met — AST check on `Add`/`Sub`/`Mul`/`Quo`/`QuoRem`/`Ceil`/`Floor`/`FloatString`, targeting the rounding temptation design.md names |
| Operational definition of "does not become unresponsive" recorded | Met — 50 ms budget with its rationale, plus the static non-blocking check |

`TestEncodingPathsNeedNoBuildTag` deserves a mention: it fails if any non-test adapter source acquires a build tag, which would quietly move that behaviour out of reach of plain `go test`. That is the erosion ADR-0003 warns about, guarded rather than trusted.

PR #69, opened 2026-08-20
feat(bridge): wire rollup and power into the boundary

### LINKED ISSUE #64: Wire rollup and power into the boundary
Governed by [SPEC-0002](docs/openspec/specs/wasm-boundary/spec.md), consuming [SPEC-0001](docs/openspec/specs/rollup-engine/spec.md) stages 2 and 3.

`internal/bridge` declares `rollup` and `power` but does not wire them:

```go
func (m *Module) Rollup(planJSON string) Envelope { return m.reserved("rollup") }
func (m *Module) Power(planJSON string) Envelope  { return m.reserved("power") }
```

That was correct when they were written — SPEC-0001's stages 2 and 3 did not exist. They do now (epic #38, PRs #49/#52/#53), so the reserved slots can be filled.

**This is not an unmet SPEC-0002 requirement.** REQ "Boundary Surface" marks both RESERVED and its scenario is conditional on their being added. The spec is satisfied; this is the "when added" case arriving.

## Scope

**In:**
- `Rollup` calls through to `GroupLeaves` → producer rollup; `Power` to the power budget
- Both return the existing `Envelope` — REQ "Boundary Surface" requires "the same envelope shape defined by REQ 'Result Envelope', with no bespoke calling convention"
- Every quantity crosses as a decimal string per REQ "Exact Quantity Encoding" — producer counts, fill and cycle times, generation, draw, deficit, and the additional-unit fix
- The stage-2 and stage-3 inputs cross the boundary: base assignments, per-site config (extractor class, EM class, generator and panel counts), and the curated constant set
- Failures map to stable codes per REQ "Sentinel Error Preservation" — `ErrMissingConstant` in particular, which stages 2 and 3 raise and `resolve` never did

**Out:**
- Any change to the envelope, error vocabulary, or namespace. If wiring needs one, that is a finding about the contract rather than a task in this story.

## Requirements

- [ ] **Boundary Surface** (SPEC-0002) — `rollup` and `power` MUST accept and return the same envelope shape as `resolve`, with no bespoke calling convention.
  - WHEN `rollup` or `power` is added in a later stage THEN it accepts and returns the same envelope shape defined by REQ "Result Envelope", with no bespoke calling convention

## Acceptance Criteria

- `rollup` and `power` return real results rather than a not-ready envelope
- Every emitted quantity is a JSON string; the existing structural walk in the encoding tests extends to the new payloads
- A plan resolving, rolling up and costing power crosses the boundary once per stage — three calls, not three per node
- A missing curated constant crosses with a distinct stable code, not `UNCLASSIFIED`
- `TestSyscallJSIsReachableOnlyThroughTheAdapter` and `TestAdapterHoldsNoDomainLogic` still pass — the rounding these stages do belongs to the domain and must not migrate into the adapter
- The `EntryPoints` list and the one-namespace assertion are unchanged

## Note

The adapter-holds-no-domain-logic guard matters more here than it did for `resolve`. Stages 2 and 3 produce counts that round up at physical boundaries, and `ceilRat` lives in the domain for that reason. The temptation design.md names — rounding a rational on the way out — has its first real opportunity in this story.

### Branch

`feat/64-wire-rollup-and-power-into-the-boundary`

### PR Convention

```
Closes #64

Implements SPEC-0002 REQ "Boundary Surface" reserved-stages scenario.
```

jonstump:
Fills the two reserved slots in `internal/bridge`. Both stages now return real results using `resolve`'s signature, envelope and error vocabulary unchanged.

This unblocks the whole of SPEC-0007 (#67) and leaf assignment in SPEC-0006 (#66).

## What crosses

**`rollup`** takes a plan, assignments, per-site config, byproducts, kitchen steps and the curated constant set; resolves, groups leaves onto bases, rolls up producers, and returns every base's construction instructions in one crossing. It resolves the plan itself rather than accepting a graph the caller hands back — REQ "Boundary Surface" requires one call to perform one complete stage, and a caller obliged to return a previous result would be assembling this one out of two crossings.

**`power`** takes generation sources, draws and constants. **It takes no plan.** The domain's power stage costs a set of counts rather than producer rows — the same figures whether they came from a rollup or from a caller sketching a base by hand — so requiring one would invent a coupling the engine does not have. It still needs a loaded artifact, because the parts and hotspot strengths it reads live in the economy section.

## The adapter still performs no arithmetic

This mattered more here than for `resolve`, as the issue predicted. Stages 2 and 3 round at every physical boundary SPEC-0001 enumerates — plants, biodomes, extractors, depots, fauna, processors, batteries, the additional-generator fix — and all of it stays in the domain. This code reads already-rounded `int64`s and renders them.

The sharpest case is power: `Balance` and `Deficit` are read from `PowerBudget` rather than derived from the `generation` and `draw` sitting beside them in the same struct. Subtracting two numbers I already have would have been the natural thing to write, and it is exactly the arithmetic-in-the-adapter that `TestAdapterHoldsNoDomainLogic` exists to catch. `FixUnsized` crosses as its own field for the same reason — a view told only `additionalGenerators: 0` would render a base in deficit as healthy.

## Contract version 1.0.0 → 1.1.0

`ResultPayload` gains `build` and `power` alongside `graph`, all `omitempty`. The envelope's four keys, the exactly-one-of invariant, the quantity encoding and the sentinel code set are unchanged, so the addition is backward compatible in shape — but the result payload is part of what a consumer parses, and leaving two different payloads claiming one version is the drift REQ "Contract Versioning" exists to prevent.

**Worth knowing:** `CheckVersion` is exact string equality, so a minor bump behaves like a major one — there is no way for the contract to express a backward-compatible addition. No consumer exists yet so nothing breaks, but that is a real property of the versioning design rather than something this PR introduces. Flagged rather than changed; the issue puts contract changes out of scope.

## Finding: `ErrMissingConstant` is dead vocabulary

The issue's scope says failures map to stable codes, "`ErrMissingConstant` in particular, which stages 2 and 3 raise."

They don't. `grep -n ErrMissingConstant internal/domain/*.go` returns only its own definition — nothing raises it. Every missing-constant path in stages 2 and 3 uses `ErrInvalidArtifact`: `Curated.validate()`, the site fill duration, panels-per-battery, the hotspot category, part rates, storage buffers. The sentinel's own comment says "Stage 1 never returns it; it is defined here so the sentinel set is complete and stable across stages" — the stages arrived and kept using `ErrInvalidArtifact`.

The acceptance criterion still holds: a missing constant crosses as `INVALID_ARTIFACT`, which is distinct, stable, and not `UNCLASSIFIED`, with a message naming the constant. `TestMissingCuratedConstantCrossesWithAStableCode` asserts that.

But it means a view cannot distinguish *"you didn't configure panels-per-battery"* (the caller's problem) from *"our artifact is broken"* (ours) — and that is precisely the distinction the `Load` path went out of its way to preserve, with a long comment explaining why classifying by context beats classifying by whichever sentinel wraps first. Changing which sentinel the domain raises is a SPEC-0001 change and out of scope here, so this is reported rather than fixed.

## Two wire names diverge from the domain's

Both because a key meaning two things in one document is worse than a name that differs from its Go field:

- `ExtractorRow.Extractors` → `extractorCount`. `extractors` is already the array holding the row, so `bases[0].extractors[0].extractors` would make a reader work out which is which. (The structural walk caught this, which is a decent argument for checking by key name.)
- `PowerRequest.Generation` → `sources`. The response's `generation` is a quantity; one word meaning a configuration inbound and a figure outbound is a needless thing to hold in mind.

## Unassigned leaves are reported

`Build.unassigned` carries the leaves the plan places nowhere. The producer stage skips them — there is no site to build them at — but the grouping deliberately keeps them, and dropping them at the boundary would make an unplaced requirement indistinguishable from one that does not exist. Both view specs have an unassigned bin.

## Tests

New `stages_test.go` builds a self-contained artifact carrying both a recipe graph and an economy section — neither domain fixture has an economy section, and the generated artifact is not a test dependency. Every number an assertion rests on is visible in the file.

Against the acceptance criteria:

| Criterion | Test |
|---|---|
| Real results, not not-ready | `TestWiredStagesShareTheEnvelopeShape`, `TestEachStageFillsItsOwnPayloadField` |
| Every quantity a JSON string; structural walk extended | `TestEveryStageQuantityCrossesAsAString` — 17 build keys, 7 power keys |
| Three calls, not three per node | `TestThreeStagesCostThreeCrossings` |
| Missing constant gets a distinct code | `TestMissingCuratedConstantCrossesWithAStableCode` |
| Purity guards still pass | verified below |
| `EntryPoints` and one-namespace assertion unchanged | untouched |

Plus deficit-with-fix, the unsized-fix state, mixed EM+solar generation, decode failures naming the field, byte-identical re-encoding, and not-ready naming the stage.

`TestReservedStagesShareTheEnvelopeShape` asserted the behaviour this story removes; replaced by `TestWiredStagesShareTheEnvelopeShape`.

## Verification

```
```

Purity guards run explicitly and passing: `TestSyscallJSIsReachableOnlyThroughTheAdapter`, `TestAdapterHoldsNoDomainLogic`, `TestEncodingPathsNeedNoBuildTag`, `TestResolveIsBoundedAndNonBlocking`, `TestEveryEnvelopeCarriesTheContractVersion`.

staticcheck is not installed locally — CI will run it.

No protected paths modified; no deferred design doc updates.

Closes #64

Implements SPEC-0002 REQ "Boundary Surface" reserved-stages scenario.

### DIFF
--- a/internal/bridge/bridge.go
+++ b/internal/bridge/bridge.go
@@ -31,7 +31,14 @@ import (
 // whenever the envelope shape, the encoding of quantities, or the sentinel
 // code set changes." The consuming view checks it and refuses a mismatch
 // rather than parsing an unexpected shape.
-const ContractVersion = "1.0.0"
+//
+// 1.1.0 added the build and power result payloads when stages 2 and 3 were
+// wired. The envelope's four keys, the exactly-one-of invariant, the
+// quantity encoding and the code set are all unchanged, so the addition is
+// backward compatible in shape — but the result payload is part of what a
+// consumer parses, and leaving two different payloads claiming one version
+// is the drift the requirement exists to prevent.
+const ContractVersion = "1.1.0"
 
 // Quantity is an exact quantity on the wire.
 //
@@ -90,8 +97,14 @@ type Envelope struct {
 }
 
 // ResultPayload is the success half.
+//
+// One field per stage, each omitempty, so a stage's result is present or
+// absent rather than present-and-empty. Exactly one is set by any entry
+// point this package exposes.
 type ResultPayload struct {
 	Graph *Graph `json:"graph,omitempty"`
+	Build *Build `json:"build,omitempty"`
+	Power *Power `json:"power,omitempty"`
 }
 
 // ErrorPayload is the failure half.
--- a/internal/bridge/module.go
+++ b/internal/bridge/module.go
@@ -105,24 +105,112 @@ func (m *Module) Resolve(planJSON string) Envelope {
 	return Success(ResultPayload{Graph: wire})
 }
 
-// Rollup and Power are RESERVED for stages 2 and 3.
+// Rollup runs stage 2 and returns one envelope carrying every base's
+// construction instructions.
 //
 // Governing: SPEC-0002 REQ "Boundary Surface" — "WHEN rollup or power is
 // added in a later stage THEN it accepts and returns the same envelope
 // shape defined by REQ 'Result Envelope', with no bespoke calling
-// convention."
+// convention." The signature, the envelope and the error vocabulary are
+// Resolve's, unchanged.
 //
-// Declared now, with the same signature as Resolve, so the shape is fixed
-// before there is an implementation to bend it around. Each returns a
-// not-ready failure until its stage is wired: a reserved name that returns
-// undefined would be indistinguishable from a typo on the view's side.
-func (m *Module) Rollup(planJSON string) Envelope { return m.reserved("rollup") }
+// The stage resolves the plan itself rather than accepting a graph the
+// caller hands back. REQ "Boundary Surface" requires one call to perform
+// one complete stage, and a caller obliged to return a previous result
+// would be assembling this one out of two crossings.
+func (m *Module) Rollup(requestJSON string) Envelope {
+	artifact, ok := m.loaded()
+	if !ok {
+		return FailureFrom(fmt.Errorf("rollup: %w", ErrNotReady))
+	}
 
-// Power is RESERVED for stage 3. See Rollup.
-func (m *Module) Power(planJSON string) Envelope { return m.reserved("power") }
+	var req RollupRequest
+	if err := json.Unmarshal([]byte(requestJSON), &req); err != nil {
+		return FailureFrom(fmt.Errorf("%w: %v", ErrMalformedInput, err))
+	}
+	plan, err := DecodePlanStrict(req.Plan)
+	if err != nil {
+		return FailureFrom(err)
+	}
+	curated, err := DecodeCurated(req.Constants)
+	if err != nil {
+		return FailureFrom(err)
+	}
+	grouping, err := DecodeRollupInput(req)
+	if err != nil {
+		return FailureFrom(err)
+	}
+	producers, err := DecodeProducerInput(req)
+	if err != nil {
+		return FailureFrom(err)
+	}
 
-func (m *Module) reserved(name string) Envelope {
-	return FailureFrom(fmt.Errorf("%s is reserved and not yet wired: %w", name, ErrNotReady))
+	constants, err := domain.NewConstants(artifact, curated)
+	if err != nil {
+		return FailureFrom(err)
+	}
+	graph, err := domain.Resolve(artifact, plan)
+	if err != nil {
+		return FailureFrom(err)
+	}
+	grouped, err := domain.GroupLeaves(graph, grouping)
+	if err != nil {
+		return FailureFrom(err)
+	}
+	build, err := domain.RollupProducers(grouped, artifact, constants, producers)
+	if err != nil {
+		return FailureFrom(err)
+	}
+	wire, err := EncodeBuild(build, grouped)
+	if err != nil {
+		return FailureFrom(err)
+	}
+	return Success(ResultPayload{Build: wire})
+}
+
+// Power runs stage 3 and returns one envelope carrying every base's power
+// position.
+//
+// Governing: SPEC-0002 REQ "Boundary Surface"
+//
+// It takes no plan. The domain's power stage costs a set of counts, not a
+// resolved graph — the same figures whether they came from a rollup or from
+// a caller sketching a base by hand — so requiring a plan here would invent
+// a coupling the engine does not have. It still requires a loaded artifact,
+// because the parts and hotspot strengths it reads live in the artifact's
+// economy section.
+func (m *Module) Power(requestJSON string) Envelope {
+	artifact, ok := m.loaded()
+	if !ok {
+		return FailureFrom(fmt.Errorf("power: %w", ErrNotReady))
+	}
+
+	var req PowerRequest
+	if err := json.Unmarshal([]byte(requestJSON), &req); err != nil {
+		return FailureFrom(fmt.Errorf("%w: %v", ErrMalformedInput, err))
+	}
+	curated, err := DecodeCurated(req.Constants)
+	if err != nil {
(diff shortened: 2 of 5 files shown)


jonstump:
## Review: approved — SPEC-0002 REQ "Boundary Surface"

(GitHub refuses self-approval on this account, so this is a comment rather than an APPROVE review.)

Verified independently rather than read from the PR body.

### Acceptance criteria

| Criterion (#64) | Verified |
|---|---|
| `rollup` and `power` return real results | Both gate on `loaded()` first, decode, call the domain, encode — no not-ready stub remains |
| Every emitted quantity is a JSON string | No numeric `json` tag anywhere in `stages.go`; the structural walk covers 17 build keys and 7 power keys |
| Three stages, three crossings | One `rollup` crossing carries every base and both row kinds |
| Missing constant → distinct stable code | Crosses `INVALID_ARTIFACT`, not `UNCLASSIFIED` |
| Purity guards still pass | CI green; independently, `grep -E '\.(Add\|Sub\|Mul\|Quo\|QuoRem\|Ceil\|Floor\|FloatString\|Float64\|SetFloat64)\('` over `stages.go`, `bridge.go`, `module.go` returns **nothing** |
| `EntryPoints` and namespace unchanged | Identical to main, byte for byte |

Conflict-marker gate clean. CI 4/4 green.

### The `ErrMissingConstant` finding is correct

I checked rather than took it:

```
internal/domain/errors.go:31  (comment)
internal/domain/errors.go:34  ErrMissingConstant = errors.New("missing constant")
internal/bridge/errors.go:62  {domain.ErrMissingConstant, CodeMissingConstant}
```

Its own definition and its code mapping. **Nothing raises it.** The issue's scope statement — "`ErrMissingConstant` in particular, which stages 2 and 3 raise" — was wrong about the code, and reporting that rather than quietly satisfying the criterion a different way is the right call.

The consequence you name is the real one: a view cannot separate *"you didn't configure panels-per-battery"* from *"our artifact is broken"*, which is exactly the distinction the `Load` path went out of its way to preserve. Agreed it is a SPEC-0001 change and out of scope here.

### The power-struct restraint is the right call and worth naming

`Balance` and `Deficit` read from `PowerBudget` rather than being subtracted from the `generation` and `draw` sitting beside them in the same struct. That is the single most tempting arithmetic in this diff and it would have looked completely reasonable. `FixUnsized` crossing as its own field is the same instinct — a view told only `additionalGenerators: 0` renders a base in deficit as healthy.

### Two non-blocking notes

**1. `rollup` resolves the graph a second time.** A view doing `resolve` → `rollup` resolves the same plan twice, because `rollup` calls `domain.Resolve` itself. Your reasoning for that is right — one call performs one complete stage, and a caller obliged to hand back a previous result would be assembling a stage from two crossings. At 36 nodes and ~141 µs it is immaterial.

It is worth naming as an accepted cost rather than leaving it to be rediscovered as a bug, because it stops being immaterial if rollup is ever called per keystroke on a large plan.

**2. `TestThreeStagesCostThreeCrossings` counts its own wrapper.** `countingModule.call` increments a counter the test itself drives, so it measures what the test does rather than what the module does. The half that has teeth is the assertion below it — one crossing carried every base and both row kinds — which is what would actually fail if a per-row accessor appeared. The comment is honest about the intent; just noting the counter is decorative.

### On the version bump

1.0.0 → 1.1.0 is right, and the flagged property is real: `CheckVersion` is exact string equality, so the contract has no way to express a backward-compatible addition. No consumer exists, so nothing breaks — but it means every future payload addition is a breaking change by construction. Correctly left out of scope.

Merging.

PR #72, opened 2026-08-21
docs: restore SPEC-0007 base planner card

jonstump:
Restores SPEC-0007, which GitHub reports as merged but which is not on `main`.

## What happened

PR #67 was stacked on `docs/spec-0005-check-fixes` and merged **into that branch**, not into `main`. That base branch was then superseded by #68 — which brought the tree-canvas files and the SPEC-0005 fixes, but not `base-planner-card`. So #67 merged, closed, and its 485 lines never reached `main`.

Verifiable:

```
git ls-tree -r main --name-only | grep base-planner-card   # nothing
git merge-base --is-ancestor 8454fca main                  # false
git branch -r --contains 8454fca                           # origin/docs/spec-0005-check-fixes only
```

Nothing prompts anyone to look, because the tracker says delivered. Found by `/sdd:audit`.

I set this up, so the lesson is mine to record: **a stacked PR merged into a base that is later superseded rather than merged loses its content silently.** Merge bottom-up in order, or rebase the child onto `main` before merging the parent.

## Two commits

**`cd3927c`** is the cherry-pick of `8454fca`, unmodified, preserving the original authorship and message. Twelve requirements, 26 scenarios, both files paired.

**`763100c`** refreshes what went stale while it was orphaned. The spec was drafted before #69, and its context, risks, migration plan and architecture diagram all described a surface blocked on `rollup` and `power` being reserved stubs. That inverted: #69 wired both, so this surface is now startable rather than blocked.

Two of its requirements turn out **better supported than they were written to be**:

- The boundary carries EM generator count, class and solar panel count as independent values, so REQ "Power Configuration Supports Mixed Sources" describes a configuration the wire actually accepts rather than one it might.
- `fixUnsized` crosses as its own field, so REQ "Deficit Is an Action, Including When It Cannot Be Sized" reads a real signal rather than inferring the state from `additionalGenerators: 0`.

Non-Goals gains the freighter and settlement cards. ADR-0006 and ADR-0007 both landed while this was orphaned, and both chose a dedicated card variant over hiding or substituting sections of this one — for the reason this spec would give: a card defined by what it suppresses accumulates a conditional at every future change. Both are `proposed` and each wants its own spec; what they inherit here is the card's composition rules, not its producer sections. Leaving SPEC-0007 silent about two ADRs that directly elaborate it would be the same staleness in reverse.

## What did not change

`spec.md` is untouched by the second commit. Its one candidate stale claim was a false positive — leaf assignment belongs to SPEC-0006, not here, so no requirement referenced #64.

The substantive content stands as reviewed: the card scoped against the panel, the three places the design and domain disagree (mixed power sources, `FixUnsized`, the uncomputed solar fix), REQ "Absent Data Is Absent" as a refusal rather than a feature, and durations specified as estimates because `Constants.ExtractorRate` records that its time unit is unconfirmed.

## Verification

- 12 requirements, each with at least one scenario; 26 scenarios all at `####`
- All four frontmatter edge targets resolve (`ADR-0004`; `SPEC-0001`, `SPEC-0002`, `SPEC-0005`)
- All relative links resolve against `main`
- `design.md` retains `## Open Questions`
- Docs only — no code, nothing to run

Part of SPEC-0007


jonstump:
## Review — one interaction with #74, addressed

This PR restores a spec whose REQ "Provenance on Displayed Figures" was deliberately hedged: it could only forbid the card from *implying* a verified/unverified distinction, because stage 2 dropped the provenance stage 1 computed and there was nothing to surface.

#74 closed that gap while this PR sat open. Merging the spec unchanged would have put a requirement on `main` that is weaker than the code now supports — the same staleness this PR exists to correct, in a new place.

Rebased onto post-#74 `main` and restated the requirement in its positive form: both the row flag and the base flag must be surfaced, neither substituting for the other, and the card must not compute provenance itself (SPEC-0005's no-arithmetic rule applied to a boolean).

**The constraint that replaces the hedge is the more useful half.** None of this project's curated constants carries a verified date yet, so a card rendered against today's set marks *every producer row*. The design tuned that treatment against two subtle chips in a prototype. A marker that relies on rarity for its restraint will not survive being on everything, so the requirement now says so and a scenario tests for it.

Verified after rebase: 12 requirements each with at least one scenario, 28 scenarios all at `####`, all frontmatter edges and relative links resolve.

PR #73, opened 2026-08-21
fix(domain): raise the missing-constant sentinel for absent Tier 2 constants

jonstump:
First of two fixes from `/sdd:audit`. SPEC-0001 REQ "Error Handling Standards" carries a scenario the engine did not satisfy:

> **WHEN** a Tier 2 constant required by a producer calculation is absent
> **THEN** the engine returns an error matching the missing-constant sentinel, naming the constant

Every absent-constant path raised `ErrInvalidArtifact`. `errors.Is(err, ErrMissingConstant)` was false everywhere — the sentinel was defined in the vocabulary, mapped to `CodeMissingConstant` at the boundary, asserted in the boundary's code-mapping test, and **raised by nothing in the repository**.

Demonstrated before fixing, with a `Curated` set missing `panelsPerBattery`:

```
error: invalid tier 1 artifact: curated constant "panels per battery" is 0; it must be supplied, not defaulted
errors.Is(err, ErrMissingConstant) = false   <-- spec requires true
errors.Is(err, ErrInvalidArtifact) = true
```

## Why it matters

The distinction is the whole value of the sentinel. A view previously received `INVALID_ARTIFACT` for both *"you did not supply panels-per-battery"* (the caller's problem, fixable by them) and *"our artifact is broken"* (ours, not fixable by them).

That is precisely the classification the boundary's `Load` path goes out of its way to preserve — it has a long comment explaining why it classifies by *context* rather than by whichever sentinel wraps first, so that a view is never sent hunting through its own payload for our bug. The producer stages were collapsing the same distinction in the other direction.

SPEC-0007's base planner card is built almost entirely on curated constants — biodome capacity, fauna yield and cycle, steps per processor, depot threshold, panels per battery, processing time — so it is the surface that would have paid for this.

## The nine sites

Each is a Tier 2 figure that must be supplied and is not:

| Site | Message |
|---|---|
| `Curated.validate` | curated constant `%q` is 0; it must be supplied |
| `ExtractorRate` | no hotspot category is configured for `%q` |
| `ExtractorRate` | `U_EXTRACTOR_S` states no extraction rate |
| `DepotCapacity` | `U_SILO_S` states no storage buffer |
| `BatteryCapacity` | `U_BATTERY_S` states no storage buffer |
| `farmRow` | crop `%q` yields 0 per plant |
| `powerFor` | `U_GENERATOR_S` states no output |
| `powerFor` | `U_SOLAR_S` states no output |
| `powerFor` | panels per battery is 0; it must be supplied |

## What deliberately did not change

`Part`, `ClassStrength`'s category lookup, and `CropFor` keep `ErrUnknownItem`. Those report a **looked-up name that does not exist**, which is a different failure from a name that exists with no usable value — and `rollup_test.go` already asserts that distinction, so it is deliberate rather than incidental.

`NewConstants` with no economy section, a hotspot class outside C/B/A/S, a non-finite strength, and the nil-argument guards all keep `ErrInvalidArtifact`: those are the artifact being structurally wrong, which is what that sentinel is for.

`producer.go`'s "site fill duration must be configured" also keeps `ErrInvalidArtifact`. It is caller configuration rather than a Tier 2 constant, and the domain has no better sentinel for it — noted rather than reclassified, since inventing one is a SPEC-0001 change.

## Three tests changed, not worked around

`TestCuratedConstantsMustBeSupplied`, `TestGeneratorWithNoStatedOutputIsRefused`, and my own `TestMissingCuratedConstantCrossesWithAStableCode` from #69 all asserted the old sentinel. They encoded the behaviour the spec says is wrong, so they are updated to match the spec — per the audit rule that specs are the source of truth and code deviation is presumed wrong.

The boundary test's comment previously explained *why* the code was `INVALID_ARTIFACT` rather than `MISSING_CONSTANT`. That explanation is now the history of the bug, so the comment says so.

## New regression guard

`TestAbsentTier2ConstantMatchesTheMissingConstantSentinel` covers all six kinds of absence in one table, and asserts three things per case: the error matches `ErrMissingConstant`, it does **not** also match `ErrInvalidArtifact` (which would erase the distinction), and it names the constant at fault as the scenario requires.

Verified it fails without the fix — reverting just the depot site turns that subtest red and nothing else.

## Also

`ErrMissingConstant`'s doc comment read *"Stage 1 never returns it; it is defined here so the sentinel set is complete and stable across stages."* Stages 2 and 3 have since landed and still did not raise it, so the comment described a state that was meant to be temporary. It now says what raises it and how it differs from its two neighbours. (This was the audit's one policy finding.)

## Verification

```
```

Found by `/sdd:audit`. Producer-stage provenance — the audit's other critical finding — follows in a separate PR.

Part of SPEC-0001

### DIFF
--- a/internal/bridge/stages_test.go
+++ b/internal/bridge/stages_test.go
@@ -363,10 +363,12 @@ func TestMixedGenerationSourcesCross(t *testing.T) {
 // SPEC-0002 REQ "Sentinel Error Preservation":
 // a missing curated constant crosses with a stable code, not UNCLASSIFIED.
 //
-// The code is INVALID_ARTIFACT rather than MISSING_CONSTANT because that is
-// the sentinel the engine raises — see the note in the PR. What this asserts
-// is the property the requirement states: a distinct, stable code, and a
-// message naming the constant so the caller can act on it.
+// MISSING_CONSTANT specifically. When this test was written the engine
+// raised ErrInvalidArtifact for an absent constant and the assertion said
+// so; SPEC-0001 REQ "Error Handling Standards" requires the
+// missing-constant sentinel, and the engine now raises it. The distinction
+// is the point: a view can tell "you did not supply this" from "our
+// artifact is broken".
 func TestMissingCuratedConstantCrossesWithAStableCode(t *testing.T) {
 	m := stagesModule(t)
 	incomplete := `{"biodomeCropSlots":"16","faunaYieldPerCycle":"12"}`
@@ -381,8 +383,8 @@ func TestMissingCuratedConstantCrossesWithAStableCode(t *testing.T) {
 	if env.Error.Code == bridge.CodeUnclassified {
 		t.Error("a missing constant crossed as UNCLASSIFIED")
 	}
-	if env.Error.Code != bridge.CodeInvalidArtifact {
-		t.Errorf("code = %q, want %q", env.Error.Code, bridge.CodeInvalidArtifact)
+	if env.Error.Code != bridge.CodeMissingConstant {
+		t.Errorf("code = %q, want %q", env.Error.Code, bridge.CodeMissingConstant)
 	}
 	if !strings.Contains(env.Error.Message, "fauna cycle seconds") {
 		t.Errorf("message %q does not name the missing constant", env.Error.Message)
--- a/internal/domain/errors.go
+++ b/internal/domain/errors.go
@@ -28,9 +28,20 @@ var (
 	// ancestor.
 	ErrCycleDetected = errors.New("cycle detected")
 
-	// ErrMissingConstant reports an absent Tier 2 economy constant. Stage 1
-	// never returns it; it is defined here so the sentinel set is complete
-	// and stable across stages.
+	// ErrMissingConstant reports an absent Tier 2 economy constant — a
+	// curated scalar the caller did not supply, or an economy value the
+	// artifact carries no usable figure for.
+	//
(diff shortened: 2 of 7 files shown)


jonstump:
## Review — one inconsistency found and fixed

**A hotspot class strength of zero was left as `ErrInvalidArtifact`** while three cases of exactly the same shape were reclassified in this PR: a part stating no rate, no storage buffer, no output. In every one the struct field is present and zero because the source did not state it, so absence and a stated zero are indistinguishable at that point.

The result was a caller getting `MISSING_CONSTANT` for a zero part rate and `INVALID_ARTIFACT` for a zero class strength, with no principled difference between them.

Fixed in `193e126`, plus a subtest using a hotspot that states `c`, `a` and `s` but not `b`. The non-finite check immediately above keeps `ErrInvalidArtifact` — that is data the artifact has and got wrong.

The rule the PR now applies uniformly: **absent-or-zero means missing; malformed means invalid.**

**Also audited the sites this PR deliberately left alone**, all correct:

- `Part`, `ClassStrength`'s category lookup, `CropFor` → `ErrUnknownItem`. A looked-up name that does not exist, and tested as such.
- Nil-argument guards, no economy section, a hotspot class outside C/B/A/S, a non-finite strength → `ErrInvalidArtifact`. Structurally wrong rather than absent.
- `site fill duration must be configured`, `base %q has no site configuration`, kitchen step quantity → caller configuration, not Tier 2. Noted in the PR body rather than reclassified, since the domain has no better sentinel and inventing one is a SPEC-0001 change.

PR #74, opened 2026-08-21
fix(domain): carry provenance through the producer stage

jonstump:
Second of two fixes from `/sdd:audit`. Independent of #73 — different requirement, no shared files beyond the package.

## The gap

SPEC-0001 REQ "Provenance Propagation" opens:

> Every computed figure MUST carry provenance indicating whether all inputs contributing to it were verified.

No producer-stage figure did. `FarmRow`, `ExtractorRow`, `RanchRow`, `KitchenStep`, `NoBuild` and `BaseBuild` had **no provenance field at all** — `grep -n Verified internal/domain/producer.go` returned nothing. Stage 1 computed the leaf's provenance into `LeafDemand.Verified`, and stage 2 read that grouping and dropped it.

The requirement's second scenario had nothing to evaluate either:

> **WHEN** a Tier 2 constant used in a producer calculation lacks a verified date
> **THEN** the resulting producer count is marked unverified

`Curated` carried no dates, so "lacks a verified date" could not be asked of it.

This lands hardest on SPEC-0007's base planner card, which is built almost entirely on curated constants — biodome capacity, fauna yield and cycle, steps per processor, depot threshold, panels per battery, processing time. Its provenance requirement is currently phrased as a *prohibition on implying a check was made*, because the positive form was unavailable. It can be tightened once this lands.

## Verified dates, not a boolean

`Curated` gains `VerifiedOn map[string]string`, keyed by constant name.

Absence rather than a boolean is deliberate. "We have not checked" is the honest default for a set that is curated *precisely because the game tables do not state these values* — ADR-0001 records biodome slots as community-wiki-sourced, and `Curated`'s own comments say the fauna constants "do not appear in the reality tables". An empty map taints every producer count, which is the correct reading of today's data rather than a limitation of the model.

## Selectivity is the point

Each row's provenance is its demand's provenance **and the constants its own arithmetic read** — not the whole set.

A blanket "any constant unverified" flag would satisfy the requirement's letter and be useless: it would mark a farm row unverified because nobody has confirmed the fauna cycle length. So:

| Constant | Taints |
|---|---|
| biodome crop slots | farm rows |
| depot threshold | extractor rows (every row — it is the comparison deciding whether a row reports *no* depots) |
| fauna yield per cycle, fauna cycle seconds | ranch rows |
| process seconds | kitchen steps |
| steps per processor | the base's processor count |

Four selectivity subtests hold that line, including two that assert a row is **not** tainted by a constant it never read.

The base's own flag is computed after its rows rather than alongside them, so a row type added later is included without a second place to remember.

## Constant names

The names were string literals inside `validate()`. They are now named constants, because a constant validated under one spelling and verified under another would silently never be verified — a bug with no symptom except everything reading unverified forever.

## Tests

- `TestUnverifiedConstantTaintsOnlyTheRowsThatReadIt` — four cases: no dates taints everything; all dates verifies everything; an unverified fauna constant leaves farm and extractor alone; an unverified dome constant leaves the extractor alone.
- `TestUnverifiedDemandTaintsItsProducerRow` — stage 1's mark now survives into stage 2, and does not bleed onto a sibling row whose demand was verified.
- `TestProducerProvenanceCrossesTheBoundary` — both halves cross, and the flag tracks the input rather than being hardcoded either way.

## Contract version 1.1.0 → 1.2.0

Every producer row and each base gained `verified`; the request gained `verifiedOn`. Additive, same reasoning as 1.1.0 — a consumer that branches on provenance needs to know it is there.

## Verification

```
```

Purity guards pass, including `TestAdapterHoldsNoDomainLogic` — the provenance decision is the domain's and the adapter only renders the flag.

Found by `/sdd:audit`.

Part of SPEC-0001

### DIFF
--- a/internal/bridge/bridge.go
+++ b/internal/bridge/bridge.go
@@ -38,7 +38,11 @@ import (
 // backward compatible in shape — but the result payload is part of what a
 // consumer parses, and leaving two different payloads claiming one version
 // is the drift the requirement exists to prevent.
-const ContractVersion = "1.1.0"
+//
+// 1.2.0 added provenance to every producer row and to each base, and the
+// verified-dates map to the request. Same reasoning: additive, but a
+// consumer that branches on provenance needs to know it is there.
+const ContractVersion = "1.2.0"
 
 // Quantity is an exact quantity on the wire.
 //
--- a/internal/bridge/stages.go
+++ b/internal/bridge/stages.go
@@ -54,6 +54,14 @@ type Curated struct {
 	// category each extracted resource sits on.
 	FaunaProducts    []string          `json:"faunaProducts,omitempty"`
 	ResourceHotspots map[string]string `json:"resourceHotspots,omitempty"`
+
+	// VerifiedOn records the date each constant was confirmed in game,
+	// keyed by the domain's constant names. A constant absent from this map
+	// has no verified date, and every figure derived from it comes back
+	// unverified.
+	//
+	// Governing: SPEC-0001 REQ "Provenance Propagation".
+	VerifiedOn map[string]string `json:"verifiedOn,omitempty"`
 }
 
 // Byproduct declares that one producer's output covers another item's
@@ -149,6 +157,7 @@ type FarmRow struct {
 	Biodomes      Quantity   `json:"biodomes"`
 	YieldPerPlant YieldRange `json:"yieldPerPlant"`
 	GrowthSeconds Quantity   `json:"growthSeconds"`
+	Verified      bool       `json:"verified"`
 }
 
 // ExtractorRow is one extracted resource at one base.
@@ -169,6 +178,7 @@ type ExtractorRow struct {
 	// configured duration, and usually less because the count rounded up.
 	RatePerSecond Quantity `json:"ratePerSecond"`
 	FillSeconds   Quantity `json:"fillSeconds"`
+	Verified      bool     `json:"verified"`
 }
 
 // RanchRow is one fauna product at one base.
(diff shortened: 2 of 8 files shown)


jonstump:
## Review — one gap found and fixed

Reviewing my own branch, so weighting toward finding problems.

**The producer stage was fixed and the power stage was left with the identical defect.** `PowerBudget.Verified` came solely from the caller's `Unverified` map, while `Batteries` is computed entirely from `PanelsPerBattery` — a curated constant. Demonstrated before fixing:

```
batteries=2  Verified=true  (panels per battery has no verified date)
```

A solar base reported two batteries and claimed the figure was verified, with no verified date anywhere in the set. That is the same requirement this PR exists to satisfy, missed one stage over.

Fixed in `4f7e2a1`, applied inside the solar branch so a base running no panels is not tainted by a constant it never reads — the selectivity the producer rows hold to. Three subtests, including that negative case.

**Also checked and found correct:**

- `NoBuild` rows carry demand provenance only — they read no constant, so nothing else to AND in.
- `KitchenStep` reads no demand (steps come from `ProducerInput.Kitchen`, not `LeafDemand`), so `ProcessSeconds` alone is right.
- `PelletFeeders` reads no constant.
- An empty `BaseBuild` reports `Verified: true` — vacuous, but a base with no assigned leaves never reaches `rollupBase`, so it is unreachable rather than wrong.

**Note for the merge:** this PR and #73 share `internal/domain/producer.go`, `internal/domain/rollup.go` and `internal/bridge/stages_test.go`, and now `power.go` as well. They are semantically independent but not merge-independent — whichever lands second needs a rebase. I said "any order" when opening them; that was wrong about file overlap.

Issue #62, opened 2026-08-20
Tests: accessibility baseline


jonstump:
Part of #57. Covers #60. Governed by [SPEC-0005](docs/openspec/specs/view-foundations/spec.md).

**Blocked by #60** — there is nothing to test until the shell and its primitives exist.

ADR-0004's fourth Confirmation criterion is that accessibility is "tested, not assumed", naming two behaviours specifically: focus returns to the invoking node on popover close, and every recompute produces an `aria-live` announcement. This story is that criterion.

## Test Requirements

- [ ] **Template render tests**: all four ARIA landmarks present exactly once; every icon-only control carries an `aria-label`
- [ ] **JS unit tests**: the focus-trap primitive returns focus to the invoking element on close by **every** route — Escape, backdrop click, and the close control — not only the one the implementation happened to be written against
- [ ] **JS unit tests**: the live-region primitive announces on recompute and not on first render, and the announcement names what changed
- [ ] **Keyboard integration tests**: tab order follows the visual layout; Enter and Space activate; Escape dismisses; arrow keys move within composite widgets
- [ ] **Automated audit**: a WCAG 2.1 AA pass over the shell with no violations

## Acceptance Criteria

- Every focus-return route is tested separately — a single "close the popover" test would pass while two of the three routes are broken
- The live-region test distinguishes announcing on recompute from announcing on render, which a naive implementation conflates
- The audit runs in CI and fails the build on a violation rather than reporting one
- Colour-only distinctions are checked: for each state distinguished by colour, a test asserts the second signal exists

## Notes

The focus-return criterion is worded around routes rather than around the behaviour because that is where this breaks in practice. An implementation that handles the close button and forgets the backdrop passes any test that closes the popover the way the author was thinking about.

### Branch

`feat/62-tests-accessibility-baseline`

### PR Convention

```
Closes #62
Part of #57

Implements SPEC-0005 REQ "View State Boundaries", Accessibility Requirements".
```

Issue #61, opened 2026-08-20
Tests: token and border discipline


jonstump:
Part of #57. Covers #58. Governed by [SPEC-0005](docs/openspec/specs/view-foundations/spec.md).

**Blocked by #58** — there is nothing to test until the stylesheet exists.

ADR-0004's Confirmation says token discipline is "verified by grep for `#` literals outside the token file". That is the right instinct and the wrong mechanism: a grep in a review checklist is an honour-system rule, and the border rules it sits beside are not greppable at all. This story makes both mechanical.

## Test Requirements

- [ ] **Stylesheet lint tests**: no colour literal outside the token file; every component colour resolves through a custom property; no `inset box-shadow` anywhere
- [ ] **Template render tests**: the three interaction states produce the specified output — hover as `filter: brightness(1.12)`, focus as an outboard 2px `--ok` outline, selection as an inboard ring drawn by an overlay element
- [ ] **JS unit tests**: control-scale helpers return 40px and 30px and never mix within a row; coarse-pointer targets resolve to at least 44px
- [ ] **Regression case**: a selected element containing positioned children keeps its ring visible over them — the defect the overlay element exists to prevent

## Acceptance Criteria

- Each test fails against a deliberately broken stylesheet, checked by breaking it rather than assumed
- The literal check runs in CI, not only locally
- The positioned-children case is a real rendered assertion, not a source-text check — the failure it guards is a paint-order defect and source text cannot see it

## Notes

The last criterion is the point of the story. `inset box-shadow` paints under positioned children, which is why the design specifies an overlay element instead. A test that greps for `inset` catches the known spelling of the mistake; a test that renders a selected element over positioned children catches the mistake.

### Branch

`feat/61-tests-token-and-border-discipline`

### PR Convention

```
Closes #61
Part of #57

Implements SPEC-0005 REQ "Token Discipline", REQ "Component Styling Discipline".
```

Issue #57, opened 2026-08-20
Implement View Foundations


jonstump:
Governed by [SPEC-0005](docs/openspec/specs/view-foundations/spec.md) ([design](docs/openspec/specs/view-foundations/design.md)), implementing [ADR-0004](docs/adrs/ADR-0004-react-view-layer.md).

The floor the view surfaces are built on: how tokens and borders are expressed, what the view is allowed to hold, how it talks to the Go core, and the accessibility behaviour every surface inherits.

ADR-0004 chose React on familiarity rather than merit-in-isolation and recorded that trade openly. What makes it cheap is ADR-0003's package boundary — a framework change costs the views and not the domain. **That holds only if domain logic stays out of the view**, which is what this epic makes checkable rather than merely intended.

Three of ADR-0004's four Confirmation criteria are requirements here — tokens as custom properties, border discipline, and the view recomputing nothing. The fourth, accessibility being tested rather than assumed, is the spec's Accessibility Requirements section.

## Scope

Foundations only. The tree canvas (React Flow, elkjs, the method popover) and the base planner cards are separate capabilities that will require SPEC-0005. Splitting them keeps each spec plannable and stops a change to one surface touching the other's requirements.

## Stories

| # | Story | Covers |
|---|-------|--------|
| #58 | Project scaffolding and the token stylesheet [foundation] | REQ Token Discipline, REQ Component Styling Discipline |
| #59 | Boundary client and module loading | REQ Boundary Client, REQ Module Loading, REQ The View Computes No Domain Values |
| #60 | App shell, view state, and the accessibility baseline | REQ View State Boundaries, Accessibility Requirements |
| #61 | Tests: token and border discipline (covers #58) | — |
| #62 | Tests: accessibility baseline (covers #60) | — |

## Dependencies

```
#58 (foundation)
  ├─ #59  boundary client      ─┐ parallel
  ├─ #60  shell + a11y         ─┘
  ├─ #61  tests for #58
  └─ (#60) ─ #62  tests for #60
```

Everything depends on #58. After it, #59 and #60 are independent of each other and of #61. #62 needs #60.

**No hotspot serialization applies.** The frontend is entirely greenfield — no file these stories touch exists yet, and the busiest file in the last 50 commits sits at 14%, well under the 50% threshold.

### Branch

`epic/57-implement-view-foundations`

jonstump:
[...]

PR #76, opened 2026-08-21
feat(web): project scaffolding and the token stylesheet

### LINKED ISSUE #58: Project scaffolding and the token stylesheet

**Foundation story** — every other story in this epic needs the project to exist. Merge before any of them begins.

Greenfield: no view code exists. This story creates the Vite + TypeScript project and recreates the design's tokens as CSS custom properties, which is the same piece of work — the token file *is* the scaffolding's stylesheet, and splitting them would leave a project with no way to express a colour.

The token source is `docs/design/theme/handoff.md`. Its validation tables (WCAG contrast, protanopia/deuteranopia simulation) are computed at render time from the token hexes, so the reference cannot drift from its own math — but a copy can drift from the reference. Recreate rather than reinterpret.

## Requirements

- [ ] **Token Discipline** (SPEC-0005) — All colour, spacing, type and control-size values MUST be declared as CSS custom properties in a single global stylesheet, recreated from the theme handoff. Component styles MUST reference those properties and MUST NOT contain hardcoded colour literals. A token the design reference does not state MUST NOT be invented in a component.
  - WHEN the stylesheet is searched for colour literals outside the token file THEN none is found, and every component colour resolves through a custom property
  - WHEN a component needs a colour or spacing step the token file does not define THEN the value is added to the token file with its design provenance, rather than written inline

- [ ] **Component Styling Discipline** (SPEC-0005) — Borders MUST carry identity only; a 3px frame denotes identity and nothing else may write to a border. Hover MUST be `filter: brightness(1.12)`; focus MUST be a 2px `--ok` outline, outboard; selection MUST be a 2px `--ok` ring, inboard, drawn by an overlay element. `inset box-shadow` MUST NOT be used for any of them. Controls MUST use one scale — 40px default, 30px small — and a row MUST NOT mix the two. Where the pointer is coarse, square targets MUST grow to at least 44px.
  - WHEN a selected element contains positioned children THEN the selection ring is drawn by an overlay element and remains visible over them
  - WHEN a row contains more than one control THEN every control in it is the same scale
  - WHEN the primary pointer is coarse THEN square targets are at least 44px

## Also in scope

- **Content Security Policy** (SPEC-0005 Security Requirements → Security Headers). The deployment MUST serve a CSP permitting WebAssembly compilation and nothing more, with no inline script and `connect-src` restricted to the origin. It lands here because it is build and deploy configuration.
- Lint, typecheck and formatting configuration for the frontend toolchain. SPEC-0005 is purely frontend, so `/sdd:plan`'s backend CI story does not apply — but the existing CI covers Go only, and the checks this project already holds itself to should extend to the view.

## Acceptance Criteria

- A grep for `#` colour literals outside the token file returns nothing
- The three interaction states render as specified, and a selected element with positioned children shows its ring over them
- No `inset box-shadow` appears anywhere in the stylesheet
- The CSP permits `wasm-unsafe-eval` and no broader `unsafe-eval`, forbids inline script, and restricts `connect-src` to the origin
- Frontend lint, typecheck and format checks run in CI alongside the existing Go jobs

### Branch

`feat/58-project-scaffolding-and-token-stylesheet`

### PR Convention

```
Closes #58
Part of #57

Implements SPEC-0005 REQ "Token Discipline", REQ "Component Styling Discipline".
```

jonstump:
Foundation story for epic #57. Greenfield — no view code existed.

Creates the Vite + React + TypeScript project under `web/` and recreates the design's tokens as CSS custom properties. These are one piece of work rather than two: the token file *is* the scaffolding's stylesheet, and a project with no way to express a colour is not scaffolding.

## The token file is a recreation, and I checked that it is

The issue says recreate rather than reinterpret, so I verified it mechanically rather than by eye. Extracted every token the handoff states and diffed against `src/styles/tokens.css`:

```
handoff states 25 colour tokens
MISSING from tokens.css: none
VALUE MISMATCH:          none
```

Then recomputed the contrast figures quoted in the comments straight from the hexes, since the handoff's own tables are computed live and a copied number is exactly the thing that drifts:

```
--text        on #282828: computed 10.75  quoted 10.75  ok
--text-muted  on #3c3836: computed  4.17  quoted  4.17  ok   <- drives the ≥17px rule
--accent      on #32302f: computed  5.20  quoted  5.20  ok
--danger      on #32302f: computed  3.82  quoted  3.82  ok
--ok          on #32302f: computed  6.24  quoted  6.24  ok
--base-6      on #32302f: computed  3.12  quoted  3.12  ok   <- the palette floor
--accent-on   on --accent: computed  6.49  quoted  6.49  ok
```

All twelve check exactly.

**One token is named here that the handoff does not name.** `--accent-on: #1d2021` — the dark text on orange buttons. The handoff states the value and its 6.49:1 measurement in prose (`dark text #1d2021 on orange buttons = 6.49:1`) without giving it a token name. REQ "Token Discipline" forbids inventing a *value* the design does not state; the value is stated, only the name is mine, and the alternative is every button repeating the literal. Flagging it because it is the one judgement call in the file.

## Discipline is mechanical, not held by review

The acceptance criteria are properties a reviewer would otherwise have to check by grepping. Each is now a job:

| Criterion | Enforced by |
|---|---|
| No `#` colour literals outside the token file | stylelint `color-no-hex` (scoped to exempt the token file) **and** `check-tokens.sh`, which also covers `.ts`/`.tsx` |
| No `inset box-shadow` anywhere | stylelint `declaration-property-value-disallowed-list` |
| CSP forbids inline script | a CI step asserting the *built HTML* has none |

`check-tokens.sh` exists because stylelint does not read `.tsx`, and a React `style` prop is just as capable of carrying a literal.

**All three were verified by injecting a violation and confirming each fires**, then restoring:

```
inject hex in base.css      → ✖ Disallowed hex color "#ff0000"  color-no-hex
                            → check:tokens: src/styles/base.css:255
inject inset box-shadow     → ✖ Disallowed value "inset 0 0 0 2px red"
inject hex in a .tsx style  → check:tokens: src/App.tsx:32
```

A guard that has never failed is a guard nobody knows works.

## Interaction states as primitives

Hover, focus and selection are defined once in `base.css`, not per component. Each was chosen to avoid a rendering defect rather than for appearance, and a defect avoided in one component and reintroduced in the next is not avoided.

Selection is an overlay pseudo-element with `z-index`, never an inset box-shadow. `App.tsx` gives every card an absolutely positioned badge stacked *above* the ring — the exact arrangement an inset shadow would vanish behind — so the criterion "a selected element with positioned children shows its ring over them" is something you can click rather than something the PR asserts.

## CSP

`default-src 'self'; script-src 'self' 'wasm-unsafe-eval'; style-src 'self'; connect-src 'self'; object-src 'none'; base-uri 'none'; frame-ancestors 'none'; form-action 'none'`

`'wasm-unsafe-eval'` permits `WebAssembly.compile` without permitting `eval()`, which is the distinction SPEC-0005 draws.

Lives in `vite.config.ts` (for `preview`, which serves the real build) and `public/_headers` (the artifact a static host reads). **Deliberately not applied to `vite dev`** — the dev server injects an inline HMR client, so a policy forbidding inline script would break the thing it protects and teach everyone to disable it. Dev is not a deployment.

Vite's module-preload polyfill is disabled because it is injected as an inline script. Verified the built output:

```
<script type="module" crossorigin src="/assets/index-BXC6af4E.js"></script>
```

Every script is `src`-referenced.

## Deliberately not done

**No webfont is loaded.** The token file names Chakra Petch, Space Grotesk and JetBrains Mono with system fallbacks, but shipping them means self-hosting under `font-src 'self'`, and that decision is not part of scaffolding. Recorded in `web/README.md`. The fallback stacks mean nothing looks broken meanwhile, but the typography is not yet the design's.

**No tests.** Issue #61 (token and border discipline) and #62 (accessibility baseline) own those, and writing them here would take their work. The lint rules above are configuration, not the tests those stories will write.

## CI

Four frontend jobs alongside the three Go ones, each independent for the same reason the Go jobs are — a formatting failure and a type error should surface separately. No path filter: gating on `web/` having changed saves one cheap install and buys a check that can silently stop running.

## Verification

```
npm run check:tokens   PASS      gofmt -l .        clean
npm run typecheck      PASS      go vet ./...      clean
npm run lint           PASS      go test ./...     all ok
npm run lint:css       PASS
npm run format:check   PASS
npm run build          PASS
```

Go side re-run because this PR edits `ci.yml`. No protected paths modified; no deferred design doc updates.

`package-lock.json` is 4,781 of the 5,760 added lines.

Closes #58
Part of #57

Implements SPEC-0005 REQ "Token Discipline", REQ "Component Styling Discipline".

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ permissions:
 
 env:
   GO_VERSION: "1.26.6"
+  NODE_VERSION: "24"
 
 # Each job is independent so a formatting failure and a test failure surface
 # separately rather than one masking the other.
@@ -67,3 +68,103 @@ jobs:
         # install must skip, not fail. See SPEC-0003 REQ "Real-Archive
         # Verification".
         run: go test -race ./...
+
+  # ---------------------------------------------------------------------
+  # Frontend (web/). SPEC-0005 is purely frontend, so the backend quality
+  # jobs above do not cover it — these are its equivalents. Each is
+  # independent for the same reason the Go jobs are: a formatting failure
+  # and a type error should surface separately rather than one masking the
+  # other.
+  #
+  # These deliberately carry no path filter. Gating them on web/ having
+  # changed would save one cheap install on Go-only changes, at the cost of
+  # a check that can silently stop running.
+  # ---------------------------------------------------------------------
+
+  web-format:
+    name: Frontend formatting
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: web
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          cache: npm
+          cache-dependency-path: web/package-lock.json
+      - run: npm ci
+      - name: prettier --check
+        run: npm run format:check
+
+  web-typecheck:
+    name: Frontend typecheck
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: web
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          cache: npm
+          cache-dependency-path: web/package-lock.json
+      - run: npm ci
+      - name: tsc
+        run: npm run typecheck
+
+  web-lint:
+    name: Frontend lint
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: web
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          cache: npm
+          cache-dependency-path: web/package-lock.json
+      - run: npm ci
+      - name: eslint
+        run: npm run lint
+      - name: stylelint
+        # Enforces SPEC-0005 REQ "Token Discipline" (no colour literal
+        # outside the token file) and REQ "Component Styling Discipline"
+        # (no inset box-shadow) mechanically rather than by review.
+        run: npm run lint:css
+      - name: colour literals outside the token file
+        # Covers .ts/.tsx, where an inline style attribute can carry a
+        # literal and stylelint does not look.
+        run: npm run check:tokens
+
+  web-build:
+    name: Frontend build
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: web
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          cache: npm
+          cache-dependency-path: web/package-lock.json
+      - run: npm ci
+      - name: vite build
+        run: npm run build
+      - name: no inline script in the built HTML
+        # SPEC-0005 Security Requirements → Security Headers forbids inline
+        # script. The CSP is only honest if the build output can live under
+        # it, so this asserts the output rather than the policy.
+        run: |
+          if grep -qE '<script[^>]*>[^<]+</script>' dist/index.html; then
+            echo "dist/index.html contains an inline script, which the CSP forbids:"
+            grep -nE '<script[^>]*>[^<]+</script>' dist/index.html
+            exit 1
+          fi
+          echo "No inline script in the built HTML."
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,8 @@ testdata/saves/
 
 # Agent worktrees created by /sdd:work — local scratch, never committed.
 .claude/worktrees/
+
+# Frontend build artifacts and dependencies (web/).
+web/node_modules/
+web/dist/
+web/node_modules/.tmp/
--- a/web/.prettierignore
+++ b/web/.prettierignore
@@ -1,0 +1,3 @@
+dist
+node_modules
+package-lock.json
--- a/web/.prettierrc.json
+++ b/web/.prettierrc.json
@@ -1,0 +1,6 @@
+{
+  "semi": true,
+  "singleQuote": false,
+  "trailingComma": "all",
+  "printWidth": 90
+}
--- a/web/.stylelintrc.json
+++ b/web/.stylelintrc.json
@@ -1,0 +1,20 @@
+{
+  "extends": "stylelint-config-standard",
+  "rules": {
+    "color-no-hex": true,
+    "color-named": "never",
+    "declaration-property-value-disallowed-list": {
+      "box-shadow": ["/inset/"]
+    },
+    "custom-property-pattern": null
+  },
+  "overrides": [
+    {
+      "files": ["src/styles/tokens.css"],
+      "rules": {
+        "color-no-hex": null,
+        "color-named": null
+      }
+    }
+  ]
+}
--- a/web/README.md
+++ b/web/README.md
@@ -1,0 +1,59 @@
+# web
+
+The React + TypeScript view layer.
+
+Governing: [ADR-0004](../docs/adrs/ADR-0004-react-view-layer.md),
+[SPEC-0005](../docs/openspec/specs/view-foundations/spec.md).
+
+## Commands
+
+| Command                | What it does                                             |
+| ---------------------- | -------------------------------------------------------- |
+| `npm run dev`          | Vite dev server                                          |
+| `npm run build`        | Typecheck, then production build to `dist/`              |
+| `npm run preview`      | Serve `dist/` with the deployment's CSP applied          |
+| `npm run typecheck`    | `tsc` with no emit                                       |
+| `npm run lint`         | ESLint over `.ts`/`.tsx`                                 |
+| `npm run lint:css`     | Stylelint over `src/**/*.css`                            |
+| `npm run check:tokens` | Colour literals outside the token file, including `.tsx` |
+| `npm run format:check` | Prettier check                                           |
+
+All of these run in CI as separate jobs.
+
+## The token file
+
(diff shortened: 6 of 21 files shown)


jonstump (commented):
## Review: story #58 acceptance criteria

I verified each criterion against the branch rather than against the PR description. Four of five hold as claimed. The fifth is satisfied by the primitive but contradicted by the surface built to demonstrate it.

### Verified

**Colour literals outside the token file.** `scripts/check-tokens.sh` has teeth — I reconstructed the tree and ran it. A planted `#ff0000` in `base.css` and a planted `rgba(…)` in `App.tsx` were both caught, exit 1. It covers `.ts`/`.tsx` where stylelint does not look, and `.stylelintrc.json` covers `.css` via `color-no-hex` + `color-named` with a `tokens.css` override. The two together close the gap each leaves alone.

**No `inset box-shadow`.** There is no `box-shadow` declaration anywhere in `src/` — only four mentions, all in comments explaining why it is absent. `declaration-property-value-disallowed-list` enforces it going forward.

**The CSP.** `script-src 'self' 'wasm-unsafe-eval'` — WASM compilation permitted, `eval()` not, no `unsafe-inline`. `connect-src 'self'`. Plus `object-src`/`base-uri`/`frame-ancestors`/`form-action` locked down beyond what the criterion asked for. Duplicated correctly in `vite.config.ts` for `preview` and `public/_headers` for a static host, with the dev-server exclusion explained rather than silently applied.

The part I did not expect and think is the strongest thing in the PR: `web-build` greps `dist/index.html` for an inline script and fails on one. That asserts the build output can actually live under the policy, rather than asserting the policy. `modulePreload: { polyfill: false }` is what makes it pass, and the comment says so.

**CI.** Four independent jobs, no path filter, with the reasoning for the absent filter written down. `NODE_VERSION` is set at `env`.

**Token provenance.** `tokens.css` claims every value is recreated from `docs/design/theme/handoff.md`. I checked: all eleven hexes I sampled are present in the handoff, the contrast figures (`10.75`, `5.20`, `6.24`, purple's `3.12` floor, the six base-accent triples) match verbatim, and the `40px = 10+10+18+2 / 30px = 6+6+16+2` control scale is quoted exactly from line 57. The `focus = 2px --ok outline outboard, selection = 2px --ok ring inboard` pairing on the same token is the handoff's, not a slip.

### Finding — the reference surface demonstrates the opposite of the criterion

The criterion is *"a selected element with positioned children shows its ring over them."*

The primitive is right. `base.css:167-175` draws the ring as a `::after` overlay at `z-index: 1`, which paints above a positioned child at `z-index: auto` — exactly the defect an inset box-shadow has.

But `reference.css:55` gives `.reference-badge` `z-index: 2`, above the ring, with this comment:

> Above the selection ring's overlay, so the ring visibly crosses in front of a child that is itself stacked.

That says the opposite of what it does. `.selectable` is `position: relative` with no `z-index`, so it creates no stacking context; the badge at `2` and the overlay at `1` are siblings in the same context and the higher value paints last. The badge is in front of the ring.

Nothing looks broken, because the badge is inset `--space-1` (4px) and the ring is a 2px border at `inset: 0` — they never touch. Which is the second half of the problem: with no geometric overlap and the stacking order inverted, the card demonstrates neither the defect nor its absence. `App.tsx:12-16` says the card exists *"so the selection-over-children case has something to show"*, and as written it shows nothing.

This matters more than a demo bug normally would because `reference.css:9-13` states the intent plainly — *"the first component in a project is the one every later one is copied from."* A stacking convention that is wrong and commented as correct is precisely what gets copied.

Suggested resolution, both one-liners: drop `z-index: 2` from `.reference-badge` so it stacks at `auto`, and move it to `inset-block-start: 0; inset-inline-end: var(--space-4)` so it crosses the ring's border box. Then the ring visibly passes in front of it and the comment becomes true. Story #61 can assert the resulting paint order.

### Minor, not blocking

- `check-tokens.sh` matches `#`-hex, `rgb(a)`, `hsl(a)` and `oklch(`, but not `lab(`, `lch(`, `color(` or `color-mix(`. Worth adding to the alternation before someone reaches for `color-mix()` — which the texture-token comment already anticipates wanting.
- The script scans `src` only, so `index.html` is outside it. A `<meta name="theme-color">` would slip past. There isn't one today.
- `style-src 'self'` with no `unsafe-inline` also blocks the `style` attribute (CSP3 falls `style-src-attr` back to `style-src`). That is consistent with this PR's no-inline-style stance, but it is a constraint on every later component, and it is not called out where the policy is explained. Worth a sentence in the `vite.config.ts` comment so the first person to hit it knows it was chosen.
- `web/node_modules/.tmp/` in `.gitignore` is already covered by `web/node_modules/`.
- `index.html` has no `<link rel="icon">`, so every load takes a `/favicon.ico` 404.

Holding the merge on the finding rather than merging and filing a follow-up, since the fix is two lines in the file everything later is copied from.

jonstump:
## Finding addressed — 828b16b

`.reference-badge` no longer sets `z-index`, and it moves to `inset-block-start: 0; inset-inline-end: var(--space-4)` so it crosses the ring's top border rather than sitting 4px inside it. With the badge stacking at `auto` and the ring's `::after` at `1` in the same stacking context, the ring now paints in front of it — which is the thing an inset box-shadow cannot do, and the reason SPEC-0005 REQ "Component Styling Discipline" forbids one.

The comment was rewritten to state why there is no `z-index` here, since the absence is the load-bearing part and an absence attracts a well-meaning addition.

All eight checks pass on the updated head, including `Frontend formatting` and `Frontend lint` — so prettier and stylelint both accept the new block.

The minor notes from the review are left as-is and are not blocking: the `check-tokens.sh` alternation missing `lab(`/`lch(`/`color(`/`color-mix(`, the script not scanning `index.html`, `style-src 'self'` also blocking the `style` attribute without that being called out where the policy is explained, the redundant `web/node_modules/.tmp/` ignore, and the missing favicon link.

PR #80, opened 2026-08-28
feat(web): app shell, view state, and the accessibility baseline

### LINKED ISSUE #60: App shell, view state, and the accessibility baseline

jonstump:
Closes #60
Part of #57

Implements SPEC-0005 REQ "View State Boundaries", Accessibility Requirements.

Based on `feat/59-boundary-client-and-module-loading` (#79), which is based on #78. Merge in that order and each retargets cleanly.

`src/App.tsx` stops being the token reference surface and becomes the app. `reference.css` is deleted; `shell.css` replaces it.

## The state library, settled

ADR-0004 chose React and left this open, recommending `useReducer` plus context or Zustand and reserving Redux Toolkit for a concrete need. This story was asked to settle it on the evidence of the first working slice.

**`useReducer` plus context.** Recorded in `web/README.md` and in `src/state/ViewStateProvider.tsx`.

The interesting state is not in the view. Every domain value lives in Go and arrives through one boundary call, so what React holds is a selection, some collapse flags, two form fields and two preferences — a handful of scalars with no cross-cutting async, no normalised entity graph and no derived-selector layer. That is the problem the alternatives exist to solve.

The one argument for Redux ADR-0004 flagged as worth keeping in view — time-travel devtools when debugging a boundary you cannot step through — does not apply. The boundary is a pure function of the plan: replaying an action sequence tells you nothing that re-issuing one `resolve` with the same plan does not, and the plan is in the URL hash. Reconsider if a surface grows a normalised cache; the bases map is the candidate, and the swap is contained behind two hooks.

- `docs/adrs/ADR-0004-react-view-layer.md`: record `useReducer` + context as the resolved state-library choice, superseding the open recommendation.

## Two rules kept by shape rather than by discipline

**View state cannot hold a plan.** `ViewState` has no field a graph could go in and none a total could go in. Caching a figure there means adding a field, which is a diff a reviewer sees — unlike an assignment into a `data` bag that was already present. Tested against the reducer rather than through the rendered app, because the claim is about every reachable state and not the ones a click sequence happens to visit.

**A cached result cannot be adjusted.** `ResultCache` holds exactly one entry and deep-freezes it. An input change leaves a stale result nowhere to be, rather than superseded-but-present — "which of these three graphs is current" is the question that produces a screen showing figures from two plans at once. And a component that tried to adjust a total rather than re-cross throws, at depth: the test mutates `stored.nodes[0].total`, which a shallow freeze would have allowed.

## The focus trap, and why it is built the way it is

Issue #62 words the focus-return criterion around **routes** — Escape, backdrop, close control — noting that "an implementation that handles the close button and forgets the backdrop passes any test that closes the popover the way the author was thinking about."

So the restore does not live in any handler. It lives in the effect's cleanup, which React runs when `open` goes false and on unmount, whatever caused either. All three routes just call `onClose`; none touches focus. A fourth route added next year inherits the behaviour by doing nothing special.

**That design was checked, not asserted.** I rewrote it as the plausible wrong version — restore inside the Escape handler, nowhere else — and ran the suite:

```
✓ focus returns to the invoking node when Escape closes the popover
✘ focus returns to the invoking node when the backdrop is clicked
✘ focus returns to the invoking node when the close control is used
```

Exactly the failure #62 predicts, and the reason the routes are separate tests rather than one "close the popover" test.

Three further mutations, all red as they should be:

| Mutation | Result |
|---|---|
| focus is never returned | all three route tests red |
| the live region announces on first render too | "announces on recompute and not on render" red |
| `<footer>` becomes a `<div>` | landmark test red |

## Live region: recompute, not render

`useAnnounceOnChange` tracks a result token, skips the first value it ever sees, and speaks only when the token changes. The token changes exactly when a crossing produces a new answer.

There are two companion tests, because the naive implementation passes the obvious one. `a preference change does not announce` opens and dismisses a popover after a resolve and asserts the region's text is byte-identical — an implementation announcing on render would fire there. And the region is asserted **empty** on load and after typing into the form, before any recompute: a user who hears "totals updated" on page load learns to ignore the region.

## Colour is never the only carrier

`StatusBadge` has no prop for "just the colour". A status is a member of a table where every member has a token class, a glyph and a word; a new status added without them does not render differently, it does not compile. The test enumerates `STATUSES` exported from the component rather than a list written in the test, so a status added without a second carrier fails rather than being missed.

## Audit

`@axe-core/playwright`, WCAG 2.1 A + AA tags, run twice: on the empty shell and on the shell with figures. The empty state and the populated state are different documents, and auditing only the first audits the screen nobody spends time on. Both clean.

## Two test-infrastructure fixes

**The stylesheet spec now discovers its files.** It listed `["base.css", "reference.css"]`. This story deletes one and adds another; a stale list would have survived that by silently checking one fewer file than exists — the direction that matters, because nothing fails. It now globs, with an assertion that the list is non-empty and contains `base.css`.

**Playwright no longer reuses an existing dev server.** `reuseExistingServer: !process.env.CI` is the obvious setting and is wrong in a repo worked in git worktrees: a dev server left running from a sibling branch answers on the same port with that branch's files, and vite serves `index.html` with a **200** for every path it does not recognise. The suite then runs against the wrong tree and fails in ways that read as application bugs. It cost two debugging detours in this session before I diagnosed it. A cold start is two seconds.

## CI

The `setup-go` change deferred out of #79 is now on that branch and inherited here, so `Frontend tests` and `Frontend build` should pass. SSH signing came back long enough to push it.

## Verified locally

`npm run lint`, `lint:css`, `typecheck`, `format:check`, `check:tokens`, `npm run build` (no inline script in `dist/index.html`), `npm test` (109 passed), `npm run test:mutation` (11/11 still red).

## Unblocks

#62 (accessibility tests) was blocked by this story and is now ready.


jonstump:
## Review — approved

Reviewed as a third party. 9/9 CI green, no conflict markers, mergeable clean. Unlike #78, this PR's body describes everything in its diff.

### Claims I checked rather than took on trust

**Focus restore lives in the effect cleanup, not a handler.** Confirmed — `useFocusTrap.ts:123` returns a cleanup that calls `target.focus()`, and no close route touches focus. The `target?.isConnected` guard is the right extra: restoring focus to a node that has since been removed would throw the user to `body` silently. This is the correct answer to #62's wording about routes, and the mutation you ran (restore inside the Escape handler) failing exactly two of three route tests is the evidence that matters.

**`ResultCache` deep-freezes.** Confirmed — recursive `deepFreeze` applied on store. A shallow freeze would have permitted `stored.nodes[0].total = ...`, which is precisely the adjustment SPEC-0005 REQ "View State Boundaries" forbids, so testing at depth is the right test.

**`ViewState` has nowhere a plan could go.** Confirmed, and worth noting how exactly it lands: the interface is `selection`, `collapsed`, `inputs`, `focusReturnTo`, `preferences` — which is SPEC-0005's own enumeration ("selection, section collapse, form inputs, focus, and view-local preferences") field for field, every one `readonly`. Caching a graph means adding a field, which is a diff a reviewer sees. That is the rule kept by shape rather than by vigilance, and it is the strongest thing in the PR.

**`STATUSES` is exported from the component and enumerated by the test**, so a status added without a glyph and a word fails rather than being missed.

### Deferred design doc update — correctly handled

The `### Deferred Design Doc Updates` section declares the ADR-0004 amendment (recording `useReducer` + context as the settled choice) rather than editing a protected path in a feature branch. That is the right handling. It needs a batch PR once epic #57 closes — #62 is still open, so not yet.

The reasoning for the choice is sound and, more usefully, states its own reversal condition: reconsider if a surface grows a normalised cache, with the bases map named as the candidate and the swap contained behind two hooks.

### One latent issue, not introduced here

`base.css` still has `.label { color: var(--text-muted) }` at `--text-label` (11.5px). `--text-muted` measures 4.72 on `--panel` and 5.30 on `--bg`, so **every use in this PR passes AA** — I checked each: the shell's header, figures section and popover all carry `panel`, and nothing carries `panel-raised`.

But it fails on the two lighter surfaces: 4.17 on `--panel-raised` (the token's stated rule, ≥17px or use `--text`) and **3.68 on `--input-bg`**, which the theme handoff never measures at all — it validates `--text-muted` against three surfaces and `--input-bg` is lighter than all of them.

This PR actually *fixes* the one live instance: `reference.css` had a `--input-bg` badge carrying `.label`, and deleting that file removes it. What remains is a primitive that is safe today and unsafe the first time a component puts it on a raised or input surface at label size. Not a change I would ask for here — flagging it for #62 or a follow-up, since the fix is one line in `base.css` plus a note that `--input-bg` is outside the design's validation.

### Also worth a follow-up

`.status-unverified` and `.status-pending` both render as `--text-muted`, so the two states are visually identical and distinguished only by glyph and word. That is compliant — colour is not the sole carrier and `StatusBadge` makes that structural — but "unverified" and "pending" mean quite different things to a planner, and a reader scanning colour sees one state. Worth asking the design for a second token rather than deciding it here.

Approving and merging.

PR #81, opened 2026-08-28
feat(web): app shell, view state, and the accessibility baseline

### LINKED ISSUE #60: App shell, view state, and the accessibility baseline

jonstump:
Closes #60
Part of #57

Implements SPEC-0005 REQ "View State Boundaries", Accessibility Requirements.

## Why this PR exists

**Recovery from my own mistake.** PR #80 carried this work and I merged it during `/sdd:review` without checking its base. It targeted `feat/59-boundary-client-and-module-loading`, not `main` — the stack's parent, which was never deleted after #79 merged, so GitHub never auto-retargeted it.

The result: #80 shows MERGED, the shell code is not on `main`, and #60 stayed open. Content intact, tracking wrong — the same stacked-PR hazard that lost SPEC-0007 in #72 and would have orphaned #59 in #78, which I flagged both times and then walked into.

This PR puts the same content on `main` and closes #60.

## What is actually in the diff

GitHub's three-dot comparison lists 60 files because the merge-base predates #78's squash. Most of those are content-identical to `main` already — I checked the blob SHAs for `boundary/client.ts`, `boundary/module.ts` and `boundary/contract.ts` and all three match `main` exactly, so they merge as no-ops.

The effective change is #80's work: `src/shell/`, `src/a11y/`, `src/state/`, `styles/shell.css`, the `App.tsx` rewrite, the `reference.css` deletion, and the shell test specs.

## Reviewed before merging

Full review on #80. Everything material was verified rather than taken on trust:

- **Focus restore lives in the effect cleanup**, not any close handler — `useFocusTrap.ts:123`, with an `isConnected` guard so a removed node does not throw focus to `body`. All three close routes inherit it by doing nothing special.
- **`ResultCache` deep-freezes** recursively, so `stored.nodes[0].total = ...` throws. A shallow freeze would have allowed exactly the adjustment SPEC-0005 forbids.
- **`ViewState` is SPEC-0005's own enumeration**, field for field — selection, collapsed, inputs, focusReturnTo, preferences, every one `readonly`. There is nowhere a graph or a total could go without adding a field, which is a diff a reviewer sees.
- **`STATUSES` is exported and enumerated by its test**, so a status added without a glyph and a word fails rather than being missed.
- **Contrast checked, not assumed.** Every `--text-muted` use in the shell sits on `panel` (4.72) or `body`/`--bg` (5.30) — both pass AA. Nothing carries `panel-raised`. This work also removes the one live failing instance by deleting `reference.css`, which had a `--input-bg` badge at label size (3.68).

CI was 9/9 green on #80 and this is the same tree.

Carried over from #80, still deferred:

- `docs/adrs/ADR-0004-react-view-layer.md`: record `useReducer` + context as the resolved state-library choice, superseding the open recommendation.

Due as a batch PR once epic #57 closes. #62 is still open.

## The lesson, since this is twice now

Merging a stacked PR requires checking `baseRefName`, not just `mergeable`. `MERGEABLE / CLEAN` is true of a PR that merges cleanly into the wrong branch. I checked the conflict gate and CI and skipped the one field that mattered.


jonstump:
Superseded by the clean branch — #81 came back CONFLICTING because `feat/59-…` diverged before #78's squash, so git sees the same content arriving through two histories. Replaced by a tree delta applied onto `main`.

Issue #87, opened 2026-08-28
Method and recipe selection


jonstump:
Part of #84. Governed by [SPEC-0006](docs/openspec/specs/tree-canvas/spec.md), implementing [ADR-0005](docs/adrs/ADR-0005-multiple-recipes-per-output.md).

**Blocked by #86** — the control opens from the card.

> **Scoped to Method Selection.** The Recipe Selection half moved to #132, which is blocked on design rather than on code. `docs/design/tree-canvas/handoff.md` contains no mention of "recipe" (verified), and SPEC-0006 requires an implementation carry the design's answer rather than stretch the segmented control past the two options it was drawn for. design.md's Migration Plan already sequences them apart: step 4 (method) is buildable, step 5 (recipe) "is blocked on design rather than on code".

The control that lets a player change how a node is made. This is the requirement ADR-0005 was written to force: a canvas that shows one route where the domain reports 26 is exactly the failure the ADR exists to prevent.

**The design has not drawn this control.** The prototype's segmented `craft | refine` control was authored 2026-08-17, a day before ADR-0005 was accepted, and addresses method only. The spec is explicit that an implementation MUST carry the design's answer for the recipe control rather than stretching the segmented control past the size it was drawn for. Expect a design round before or alongside the build.

## Requirements

- [ ] **Method Selection** (SPEC-0006) — Clicking or pressing Enter on a node MUST open a control offering that node's methods. The options offered MUST be the node's `legalMethods` from the payload; the canvas MUST NOT compute which methods are legal. A method that is not available MUST be rendered and inert, with the reason stated, rather than hidden. The control MUST state the consequence of a change before it is made, in the domain's own terms.
  - WHEN the method control opens for a node THEN the options shown are that node's `legalMethods`, and the canvas consulted no other source to determine them
  - WHEN a node cannot be refined THEN the refine option is rendered, is not activatable, and states why

## Acceptance Criteria

- Every option offered comes from the payload — no legality is computed in the view, checkable by grep as much as by test
- An unavailable method is present, inert, and says why. A hidden option cannot be told apart from one that does not exist
- Both controls are keyboard-operable end to end, inheriting the focus trap from #82 rather than reimplementing it

## Notes

The 26 and 61 figures are measured from the shipped refiner tables, not estimates. They moved to #132 along with the requirement they constrain.

The consequence-before-change requirement is not a confirmation dialog: it is that the control says what will happen in the domain's terms, so the player is choosing rather than discovering.

### Branch

[...]

PR #93, opened 2026-08-28
test(web): complete the accessibility baseline

### LINKED ISSUE #62: Tests: accessibility baseline

jonstump:
Closes #62
Part of #57

Implements SPEC-0005 Accessibility Requirements.

ADR-0004's fourth Confirmation criterion is that accessibility is "tested, not assumed". #60 shipped the primitives with tests for the parts load-bearing to building them — four landmarks, the axe audit, all three focus-return routes, the live region's recompute-not-render distinction. Those are on main and are **not repeated here**. This is what that left.

10 new tests, and the mutation check grows from 11 to 21.

## Every case here passes vacuously if written carelessly

**Icon-only labelling needed a bespoke check, because axe does not catch it.** A button containing only `✕` has an accessible name by the specification's own definition — it is named `✕` — and axe's `button-name` rule is satisfied. A screen reader reads "multiplication X" or skips it. The rule enforced is therefore stricter than the standard: a name containing no letter and no digit is not a name.

There's a test asserting axe still misses this, so if a future axe version starts flagging glyph-named controls, that fails and the bespoke check gets reconsidered — rather than carried forever because nobody rechecked the assumption.

**Tab order is asserted for the shell only, and says why.** SPEC-0006's design settles the *opposite* rule for the tree canvas: node tab order is the payload's topological order, not the layout's, because build order beats reading order for a dependency graph. It explicitly calls that "a deliberate divergence ... worth stating so it is not later 'fixed'". A blanket "tab order follows visual layout" test would be exactly that later fix, so this one is scoped and the exception is written down.

**Arrow keys have no composite widget to drive yet.** A test pressing arrow keys would assert nothing. The honest form enumerates the roles that carry the obligation — `tablist`, `menu`, `listbox`, `tree`, `grid`, `radiogroup`, `toolbar`, `combobox` — and fails the moment one appears, so the requirement is enforced when it becomes real. #87's method and recipe controls are where that's expected.

**Space as well as Enter.** Testing one and assuming the other is the entire failure mode: a `role="button"` div responds to Enter through the browser default and to Space not at all.

**Colour-carried states are now exhaustive.** All five statuses render through the real component via a fixture, one badge per member of its own exported table. #60 checked one badge and inferred four.

## The mutation check found two real defects in my own tests

#82 ran the accessibility mutations by hand and didn't commit them — which proves an afternoon and nothing after. They're committed now, and they earned it on the first run:

**Relabelling `unverified` to "Pending" left "each status is distinguishable" green.** The test compared `innerText`, which includes the glyph — so two badges that *announce identically* looked distinct because one showed `?` and the other `…`. It now compares accessible text with `aria-hidden` content removed, which is the population the requirement is actually about.

**One mutation anchor didn't match** — the check reporting its own blind spot rather than quietly passing.

The Escape-only focus-trap mutation is the one worth reading in the output. It is not a strawman; it's the natural way to write the restore, and it leaves the Escape test green while the backdrop and close-control tests go red. That is precisely why #62 words the criterion around routes.

```
ok    focus is never returned on close — tests/shell/a11y-primitives.spec.ts went red
ok    focus is restored only by the Escape handler — tests/shell/a11y-primitives.spec.ts went red
ok    the focus trap stops containing Tab — tests/shell/a11y-primitives.spec.ts went red
ok    the live region announces on first render too — tests/shell/a11y-primitives.spec.ts went red
ok    the live region announces on every render — tests/shell/a11y-primitives.spec.ts went red
ok    a landmark goes missing — tests/shell/shell.spec.ts went red
ok    the navigation landmark loses its name — tests/shell/shell.spec.ts went red
ok    a status loses the word beside its colour — tests/shell/a11y-baseline.spec.ts went red
ok    two statuses become indistinguishable without colour — tests/shell/a11y-baseline.spec.ts went red
ok    a control is named only by a glyph — tests/shell/a11y-baseline.spec.ts went red
```

All 21 red, including the 11 stylesheet mutations from #61.

## One threshold I got wrong

The exhaustive status test failed on its first run against a threshold I'd set at three letters. The `ok` badge reads **"OK"** — two characters, and a perfectly good word. The threshold was wrong, not the component. It's calibrated to the real table now with the reason recorded in the test, rather than the table trimmed to fit the threshold.

## Acceptance criteria

| Criterion | Where |
|---|---|
| Every focus-return route tested separately | #82, plus two mutations here proving the separation matters |
| Live-region test distinguishes recompute from render | #82, plus two mutations here |
| The audit runs in CI and fails the build | `Frontend tests` job, already green on main |
| Colour-only distinctions checked for **each** state | new — exhaustive via the fixture, plus a distinguishability test |
| All four landmarks present exactly once | #82, plus a mutation here |
| Icon-only controls carry `aria-label` | new — bespoke check, because axe doesn't catch it |
| Tab order, Enter/Space, Escape, arrow keys | new |

## Verified locally

`npm run lint`, `lint:css`, `typecheck`, `format:check`, `check:tokens`, `npm test` (119 passed), `npm run test:mutation` (21/21 red).

### DIFF
--- a/web/scripts/mutation-check.sh
+++ b/web/scripts/mutation-check.sh
@@ -1,8 +1,18 @@
 #!/usr/bin/env bash
-# Governing: SPEC-0005 REQ "Token Discipline", REQ "Component Styling Discipline"
+# Governing: SPEC-0005 REQ "Token Discipline", REQ "Component Styling
+# Discipline", Accessibility Requirements
 #
 # "Each test fails against a deliberately broken stylesheet, checked by
 #  breaking it rather than assumed." (issue #61)
+#
+# "Every focus-return route is tested separately — a single 'close the
+#  popover' test would pass while two of the three routes are broken."
+#  (issue #62)
+#
+# #61 covered the stylesheet. #62 extends the same treatment to the
+# accessibility primitives, because the same argument applies and more
+# sharply: a focus trap that has stopped restoring focus looks identical on
+# screen to one that works.
 #
 # A passing suite proves the stylesheet is not broken in the ways the suite
 # knows how to look for. It does not prove the suite can see anything at all —
@@ -20,8 +30,15 @@ cd "$(dirname "$0")/.."
 
 BASE="src/styles/base.css"
 TOKENS="src/styles/tokens.css"
+TRAP="src/a11y/useFocusTrap.ts"
+LIVE="src/a11y/useLiveRegion.ts"
+SHELL="src/shell/AppShell.tsx"
+BADGE="src/shell/StatusBadge.tsx"
 
-restore() { git checkout -- "$BASE" "$TOKENS" 2>/dev/null || true; }
+MUTABLE="$BASE $TOKENS $TRAP $LIVE $SHELL $BADGE"
+
+# shellcheck disable=SC2086
+restore() { git checkout -- $MUTABLE 2>/dev/null || true; }
 trap restore EXIT
 
 # mutate <file> <python-expression-safe-old> <new>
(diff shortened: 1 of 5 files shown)


jonstump:
## Review — approved

Reviewed as a third party. Base is `main` (checked first this time), no conflict markers, 9/9 CI, +599/-2 across five test-only files.

### Claims I verified rather than took on trust

**The axe blind-spot test is the best thing here.** A button containing only `✕` genuinely satisfies `button-name` — it *is* named, by the specification's own definition — so the bespoke check is stricter than the standard, and the reasoning for that is right. What makes it good is `axe does not catch the glyph case, which is why the check above exists`: it fails *when axe improves*, with a message saying the bespoke check may now be redundant. That is the opposite of the usual outcome, where a redundant check is carried forever because nobody rechecked the assumption.

The planted-control test discriminates in both directions — two glyph-only controls flagged, one `aria-label`led and one glyph-plus-word not — so a checker returning everything or nothing fails. That is the failure mode a plant test usually misses.

**The `aria-hidden` fix is correct and the defect was real.** `accessible-name.ts` clones the element, removes `[aria-hidden="true"]` subtrees, then requires `\p{L}` or `\p{N}`. Comparing `innerText` (as the pre-fix version did) meant two badges that announce identically looked distinct because their glyphs differed — the population the requirement is about is what a screen reader reads, not what is painted. Finding that via a mutation rather than by inspection is the point of having the mutations.

**The composite-widget test is the honest form.** It enumerates the nine roles carrying an arrow-key obligation and fails when one appears, after opening the popover so the dialog's contents are in scope. A test pressing arrow keys against a shell with no composite widget would assert nothing and pass forever. Naming SPEC-0006's method and recipe controls as where this becomes real is the right pointer.

**The tab-order scoping is right, and for the stated reason.** SPEC-0006 settles the opposite rule for the tree canvas — payload order, not visual order — and explicitly calls it "a deliberate divergence ... worth stating so it is not later 'fixed'". A blanket layout-order test would have been exactly that later fix. Scoping it to the shell and writing down why is correct.

**The threshold calibration went the right way.** Two characters, not three, because `ok` renders "OK" — and the note records that the first draft failed on it and the threshold moved rather than the table. That is the direction that matters: a threshold trimmed to fit is a test that stopped measuring.

### One accuracy fix, applied

The body read `Implements SPEC-0005 REQ "View State Boundaries", Accessibility Requirements` — carried over from #82's line. This PR touches no view-state file; the diff is five accessibility test files. Corrected to `Implements SPEC-0005 Accessibility Requirements`, because `/sdd:check` traces implementation back through exactly these references and a wrong one is noise in the tool meant to catch drift.

### Note for whoever picks up SPEC-0006

`ICON_ONLY_CONTROL_AUDIT` and the `\p{L}\p{N}` name rule are reusable, and the tree canvas will need them — its method badges are glyph-plus-label by design, and the recipe control does not exist yet. Worth importing rather than rewriting.

Approving and merging. This closes the last story in epic #57.

Issue #94, opened 2026-08-28
Implement Base Planner Card


jonstump:
Governed by [SPEC-0007](docs/openspec/specs/base-planner-card/spec.md) ([design](docs/openspec/specs/base-planner-card/design.md)).

Where the plan stops being a graph and becomes instructions: this many plants in this many biodomes, this many extractors at this class, this much power short and this many generators to fix it.

Inherits SPEC-0005 wholesale — tokens, component styling, the no-arithmetic rule, the boundary client, view-state boundaries, the accessibility baseline. All of it is on main, and the accessibility work completed in #93.

## Stories

| Issue | Story | Requirements | Depends on |
|---|---|---|---|
| #95 | Boundary client: rollup and power `foundation` | — (the client surface every story below reads through) | — |
| #96 | Card composition, identity, and producer sections | Card Composition From the Build Payload; Base Identity and Selection; Producer Sections; Byproducts Are Shown, Not Omitted | #95 |
| #97 | Site and power configuration | Site Configuration; Power Configuration Supports Mixed Sources | #96 |
| #98 | Power position, deficit, and the build rollup footer | Power Position; Deficit Is an Action, Including When It Cannot Be Sized; Build Rollup Footer | #96 |
| #99 | Duration, provenance, and absent data | Duration Display; Provenance on Displayed Figures; Absent Data Is Absent | #96 |
| #100 | Tests: composition, identity, and producer sections | covers #96 | #96 |
| #101 | Tests: power, deficit, provenance, and absence | covers #97, #98, #99 | #98, #99 |

All twelve of the spec's requirements appear in exactly one story.

## The upstream half is done; the client half is not

design.md records that when the spec was drafted `Module.Rollup` and `Module.Power` were reserved stubs, and that **#64 wired both** — so stage 2 and stage 3 payloads cross today. Verified against `main`: `internal/bridge/module.go` implements both in full and `cmd/planner` registers them on the `nmsPlanner` namespace.

What is missing is on the view side. `web/src/boundary/client.ts` exposes **only `resolve`**. That is what #95 builds, and it is why nothing else in this epic can start before it.

## A consolidates work that was scoped to SPEC-0006

#88 (leaf assignment, SPEC-0006) was planned carrying its own `rollup` client extension, with a note that SPEC-0007 would extend the same file for `power` and that whoever planned this should consider pulling both together.

[...]

Issue #88, opened 2026-08-28
Leaf assignment to bases


jonstump:
Part of #84. Governed by [SPEC-0006](docs/openspec/specs/tree-canvas/spec.md).

**Blocked by #85** and **#95** (the boundary client this story reads through).

> **The URL-hash round trip moved to #135.** `Plan` is `{target, quantity, methods, recipes}` and carries no assignments; `RollupRequest` carries them as a sibling of `plan`; SPEC-0002 has no requirement encoding them; and `encodePlanToHash` serialises `Plan` only. This story builds to the narrower reading of "carried as SPEC-0002 encodes it" — assignments travel to the domain through the documented boundary contract, which `RollupRequest.assignments` satisfies. What a shared link should carry beyond the plan is #135, and it applies equally to SPEC-0007's site configuration.

Assigning a leaf resource to a base, which is what turns a dependency tree into a build plan.

## The boundary work moved to #95

The spec says assignment "reaches the domain through stage 2, whose boundary entry point is reserved and not yet wired", and forbids working around that by reading the domain's rollup types directly.

That is half true, and worth stating precisely: **the Go side is complete.** `internal/bridge/module.go` implements `Rollup` and `Power` in full, `cmd/planner` registers both, and `EntryPoints` lists them. What was missing is on the view side — `web/src/boundary/client.ts` exposed only `resolve`.

This story originally carried that client extension, with a note that SPEC-0007 would extend the same file for `power` and that whoever planned SPEC-0007 should consider pulling both together.

**That happened.** [#95](https://github.com/jonstump/nms-base-planner/issues/95) is the consolidated boundary-client story and carries `rollup` and `power` with their payload decoders. Two stories independently adding methods to one file is the drift a foundation story exists to prevent.

**This story now depends on #95** and consumes `client.rollup()` rather than building it.

## Requirements

- [ ] **Leaf Assignment to Bases** (SPEC-0006) — The canvas MUST allow a leaf node to be assigned to a base, and the assignment MUST be operable without a pointing device. An assignment MUST be plan state carried as SPEC-0002 encodes it, not view-local state. Reassigning a leaf MUST cause the affected figures to be recomputed through the boundary; the canvas MUST NOT adjust any base's totals itself.
  - WHEN a player navigates to a leaf node and opens its control using only the keyboard THEN the base assignment can be changed and committed without a pointing device
  - WHEN a leaf is moved from one base to another THEN the new figures come from a boundary call, and no base total was adjusted in the view

[...]

PR #116, opened 2026-08-28
feat(web): boundary client rollup and power

### LINKED ISSUE #95: Boundary client: rollup and power

The client surface every other story in this epic reads through, and the reason none of them can start first.

## What exists and what does not

The Go half is complete. `internal/bridge/module.go` implements `Rollup` and `Power` in full, `cmd/planner` registers both on the `nmsPlanner` namespace, and `bridge.EntryPoints` lists them — design.md records this as #64's work, and it holds on `main` today.

`web/src/boundary/client.ts` exposes **only `resolve`**. That is the entire gap.

Extends `web/src/boundary/` — `client.ts` gains two methods, and the payload decoders join `graph.ts` as siblings. The pattern is established: `selectGraph` validates every field, keeps every quantity as an exact string, and rejects the whole payload rather than dropping a field, because a build missing one figure reads as a build that costs less than it does.

## This consolidates work scoped to SPEC-0006

#88 (leaf assignment, SPEC-0006) was planned carrying its own `rollup` extension, with a note that SPEC-0007 would want `power` from the same file and that whoever planned this epic should consider pulling both together. This is that consolidation, and #88 now depends on this story.

Two stories independently adding methods to one client is the drift a foundation story exists to prevent.

## Requirements

This story carries no SPEC-0007 requirement of its own — it builds the surface the requirements are read through. The contract it must honour is SPEC-0002's, which is already implemented and fixed on the Go side.

- [ ] **`rollup` client method** — Crosses a plan plus curated constants and returns the stage 2 build payload. MUST decode through the same envelope path as `resolve`, MUST branch on stable error codes and never on message text, and MUST reject a payload whose shape it does not recognise rather than returning a partial build.
- [ ] **`power` client method** — Crosses a base's power configuration and returns the stage 3 payload. Same envelope path, same code-only branching, same all-or-nothing decode.
- [ ] **Typed decoders** — `BaseBuild` and `PowerBudget` payloads decoded field by field, with every quantity kept as the exact string the boundary sent. `fixUnsized` MUST be decoded as its own field rather than inferred from a zero generator count.
- [ ] **The readiness contract is inherited, not reimplemented** — Both methods MUST go through the existing `#call` path, so a call issued before the module is ready waits on the same load `resolve` does and retries once, rather than each method growing its own readiness handling.

## Acceptance Criteria

- Both methods return the same `Outcome` discriminated union `resolve` does, so a component handles all three identically
- `tests/boundary/discipline.spec.ts` still passes unchanged — no new boundary source branches on message text, converts a quantity to a number, or navigates
- A payload missing a field the decoder requires yields a failure, not a build with a hole in it
- `fixUnsized` survives the decode as a distinct signal. A test asserts that a deficit with zero additional generators and `fixUnsized: true` is distinguishable from one with zero and `fixUnsized: false` — the state design.md warns an implementer would otherwise conclude was nothing to show
- Every quantity round-trips as a string. The exactness discipline the graph decoder already enforces applies identically here
- Tests live in this story rather than a companion: this is library code with no rendered surface, matching the reasoning #59 used

### Branch

`feat/95-boundary-client-rollup-and-power`

### PR Convention

```
Closes #95
Part of #94

Implements SPEC-0002 REQ "Result Envelope", REQ "Exact Quantity Encoding" on the view side, unblocking SPEC-0007.
```

jonstump:
Closes #95
Part of #94

Implements SPEC-0002 REQ "Result Envelope", REQ "Exact Quantity Encoding" on the view side, unblocking SPEC-0007.

The client surface every SPEC-0007 story reads through, and the reason none of them could start first.

## The gap was one-sided

The Go half was already complete — `internal/bridge/module.go` implements `Rollup` and `Power` in full, `cmd/planner` registers both, and `EntryPoints` lists them. `web/src/boundary/client.ts` exposed **only `resolve`**. That was the entire gap, and it blocked an entire epic.

Both methods go through the existing `#call` path rather than growing their own readiness handling, so a rollup issued before the module has loaded waits on the same load a `resolve` does and retries once. That behaviour is inherited, not repeated.

## The decode primitives moved rather than being copied

`decode.ts` now holds `object`, `text`, `textList`, `quantity`, `optionalQuantity`, `flag` and `list`. Three decoders with three private copies of `optionalQuantity` is three places for the omitempty rule to drift, and the rule is identical everywhere: absent is fine, present-and-wrong is not.

`graph.ts` imports them, and its two hand-rolled `children`/`nodes` loops collapse into the shared `list` helper — which is where the all-or-nothing rule lives now, in one place instead of three.

## Two decisions worth reading

**`fixUnsized` is decoded, never inferred.** design.md warns that an implementer working from the prototype alone "would meet a budget in deficit with `AdditionalGenerators` of zero and reasonably conclude there was nothing to show." A test drives three payloads that differ only in that boolean and asserts the two zero-generator cases are told apart by the flag alone — a decoder that derived it from `additionalGenerators === "0"` would collapse them.

**Booleans decode strictly, not truthily.** `verified` and `fixUnsized` both carry meaning in their false case, so an absent flag is a payload this view does not understand rather than a definite "no". Coercing it would report an answer the payload never gave, and SPEC-0007 requires the card to mark unverified figures.

## Tested twice, deliberately

`tests/boundary/stages.spec.ts` drives the decoders with hand-built payloads. That proves they decode what those tests *say* the wire looks like.

`tests/boundary/stages-integration.spec.ts` runs both stages against the real module and the real artifact. That proves what the wire **actually is**:

- `rollup` crosses and returns a build, with unassigned leaves reported rather than lost
- `power` returns generation, draw and balance as exact quantity strings — no subtraction anywhere in the view
- **a deficit with no generator class configured crosses as `fixUnsized: true` from the real engine**, with `additionalGenerators: "0"`
- a missing curated constant is refused rather than defaulted, per SPEC-0001 design.md's "injected, never hardcoded"
- all three stages return the same `Outcome` shape at runtime, not just in the type declaration

Without the second file, the first is a test of my own assumptions about the payload.

## Consolidates work #88 was carrying

#88 (leaf assignment, SPEC-0006) was planned with its own `rollup` extension and a note that SPEC-0007 would want `power` from the same file. It now depends on this instead — two stories independently adding methods to one client is the drift a foundation story exists to prevent. **This unblocks #88 as well as the whole of #94.**

## One small thing

The rest-destructuring omit idiom (`const { x: _dropped, ...rest }`) trips `@typescript-eslint/no-unused-vars`. I used an explicit `without()` helper in the tests rather than loosening the rule project-wide for one idiom.

## Verified locally

`npm run lint`, `lint:css`, `typecheck`, `format:check`, `check:tokens`, `npm run build`, `npm test` (138 passed — 73 boundary, up from 54), `npm run test:mutation` (21/21 still red).

## Note on the push

Pushed over HTTPS: the 1Password SSH agent has stopped signing again. No workflow files are touched here, so the token's missing `workflow` scope did not bite.

### DIFF
--- a/web/src/boundary/build.ts
+++ b/web/src/boundary/build.ts
@@ -1,0 +1,319 @@
+/*
+ * The stage 2 build payload: one base's construction instructions.
+ *
+ * Governing: ADR-0004 (React view layer), SPEC-0002 REQ "Exact Quantity
+ * Encoding", SPEC-0005 REQ "The View Computes No Domain Values",
+ * SPEC-0007 REQ "Card Composition From the Build Payload"
+ *
+ * Every count here is the domain's — plants, biodomes, extractors, depots,
+ * fauna, nutrient processors, pellet feeders. SPEC-0007 forbids the view
+ * computing any of them from a quantity and a rate even when both are in
+ * the payload, so they arrive as exact strings and stay that way.
+ *
+ * Two fields exist because omitting them would lose a planning result:
+ * `noBuild` carries demands a byproduct already covers, and `unassigned`
+ * carries leaves the plan places nowhere. An absent row is indistinguishable
+ * from an overlooked requirement.
+ */
+
+import { flag, list, object, quantity, text, type Raw } from "./decode";
+import type { Quantity } from "./quantity";
+
+export interface YieldRange {
+  readonly min: Quantity;
+  readonly max: Quantity;
+}
+
+export interface FarmRow {
+  readonly itemId: string;
+  readonly name: string;
+  readonly required: Quantity;
+  readonly plants: Quantity;
+  readonly biodomes: Quantity;
+  /** Both bounds. The domain sizes plants on the pessimistic one. */
+  readonly yieldPerPlant: YieldRange;
+  readonly growthSeconds: Quantity;
+  readonly verified: boolean;
+}
+
+export interface ExtractorRow {
+  readonly itemId: string;
+  readonly name: string;
+  readonly class: string;
+  readonly required: Quantity;
+  readonly extractorCount: Quantity;
+  readonly depots: Quantity;
+  readonly ratePerSecond: Quantity;
+  /** What the built extractors actually take — at most the configured duration. */
+  readonly fillSeconds: Quantity;
+  readonly verified: boolean;
+}
+
+export interface RanchRow {
+  readonly itemId: string;
+  readonly name: string;
+  readonly required: Quantity;
+  readonly fauna: Quantity;
+  readonly cycleSeconds: Quantity;
+  readonly verified: boolean;
+}
+
+export interface KitchenInput {
+  readonly itemId: string;
+  readonly perOutput: Quantity;
+}
+
+export interface KitchenStep {
+  readonly itemId: string;
+  readonly name: string;
+  readonly recipe: string;
+  readonly required: Quantity;
+  readonly processSeconds: Quantity;
+  readonly final: boolean;
+  readonly verified: boolean;
+  readonly inputs: readonly KitchenInput[];
+}
+
+/** A demand a byproduct at the same base already covers. */
+export interface NoBuildRow {
+  readonly itemId: string;
+  readonly name: string;
+  readonly from: string;
+  readonly required: Quantity;
+  readonly verified: boolean;
+}
+
+export interface Site {
+  readonly extractorClass: string;
+  readonly fillSeconds: Quantity;
+}
+
+export interface BaseBuild {
+  readonly base: string;
+  readonly site: Site;
+  readonly farms: readonly FarmRow[];
+  readonly extractors: readonly ExtractorRow[];
+  readonly ranches: readonly RanchRow[];
+  readonly kitchen: readonly KitchenStep[];
+  /** Per base, not per row. Summing a per-row figure would overcount. */
+  readonly nutrientProcessors: Quantity;
+  readonly pelletFeeders: Quantity;
+  readonly noBuild: readonly NoBuildRow[];
+  /** The base-level answer: every row here, and the constant that sized its processors. */
+  readonly verified: boolean;
+}
+
+export interface Demand {
+  readonly itemId: string;
+  readonly name: string;
+  readonly total: Quantity;
+  readonly verified: boolean;
+}
+
+export interface Build {
+  readonly bases: readonly BaseBuild[];
+  /** Leaves the plan places nowhere. Reported rather than lost silently. */
+  readonly unassigned: readonly Demand[];
+}
+
+function decodeYieldRange(value: unknown): YieldRange | null {
+  const raw = object(value);
+  if (!raw) return null;
+  const min = quantity(raw["min"]);
+  const max = quantity(raw["max"]);
+  return min === null || max === null ? null : { min, max };
+}
+
+function decodeFarm(value: unknown): FarmRow | null {
+  const raw = object(value);
+  if (!raw) return null;
+
+  const itemId = text(raw["itemId"]);
+  const name = text(raw["name"]);
+  const required = quantity(raw["required"]);
+  const plants = quantity(raw["plants"]);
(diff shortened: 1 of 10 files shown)


jonstump:
## Review — one gap in the foundation surface, merging anyway

Reviewed as a third party. Base is `main`, no conflict markers, 9/9 CI, +1126/-51.

### The gap: the client cannot send `verifiedOn`

`Curated` in `requests.ts` carries all seven scalars plus `faunaProducts` and `resourceHotspots` — but not `verifiedOn`, which the Go wire type has as `json:"verifiedOn,omitempty"`.

The consequence is end to end. `Constants.ConstantVerified` reads `_, ok := c.curated.VerifiedOn[name]`, so an absent map means *every* curated constant is unverified, which means every producer row and every solar power budget returns `verified: false`. This client decodes that flag carefully — seven row types declare it, `flag()` refuses to coerce it — and then can never influence it.

So the provenance mechanism is **observable but not controllable from the view.** That is the same shape as the tree canvas's `unverified` badge being unreachable because the normalizer never emits `verified: false`: a mechanism built, wired, tested, and unable to fire.

It is not wrong *today* — no constant has a verified date, so permanently-false is the honest answer, and SPEC-0007's amended REQ "Provenance on Displayed Figures" already anticipates the marker being on everything. What it means is that the day someone confirms a constant in game and records the date, there is no way to send it without reopening this file.

**Which is the one thing this PR argues against.** Its own case for consolidating #88's rollup extension is that "two stories independently adding methods to one client is the drift a foundation story exists to prevent." An omitted request field is that same later touch, one field instead of one method.

**I am merging rather than blocking**, and filing the gap as a follow-up. It is additive, non-breaking, three lines, and holding an epic-unblocking foundation PR over one absent optional field is disproportionate. But it should be closed before a SPEC-0007 card story reads provenance and finds it inert.

### Everything else verified, not taken on trust

**Wire names match, including the two awkward ones.** `#69` renamed `ExtractorRow.Extractors` to `extractorCount` (because `extractors` is already the array holding the row) and the power request's `generation` to `sources` (because the response's `generation` is a quantity). Both renames are honored — `build.ts` reads `raw["extractorCount"]` while typing the array as `extractors`, and `requests.ts` uses `sources`. Those are exactly the fields most likely to be got wrong by anyone working from the Go struct names rather than the JSON tags.

**Contract version is `1.2.0`**, matching `internal/bridge` after #74. A stale pin would have refused every payload.

**`fixUnsized` is decoded, never inferred** — its own field with a comment explaining the distinction from a zero `additionalGenerators`, and the integration test drives three payloads differing only in that boolean. A decoder deriving it from `additionalGenerators === "0"` would collapse the two zero-generator cases, which is precisely the state SPEC-0007 REQ "Deficit Is an Action, Including When It Cannot Be Sized" exists for.

**Booleans decode strictly.** `flag()` is `typeof value === "boolean" ? value : null` — no coercion. Correct, and for the stated reason: `verified` and `fixUnsized` both carry meaning in their false case, so an absent flag is a payload the view does not understand rather than a definite no.

**The two-file test split is the right instinct.** Hand-built payloads prove the decoders match what the test author believes the wire looks like; the integration file against the real module and real artifact proves what it is. The body says it plainly — "without the second file, the first is a test of my own assumptions about the payload" — and the deficit-with-no-class case crossing as `fixUnsized: true` from the real engine is the assertion that earns it.

### Small note

Using an explicit `without()` helper in tests rather than loosening `no-unused-vars` project-wide for the rest-destructuring omit idiom is the right trade. A lint rule relaxed for one idiom stays relaxed for everything.

jonstump:
Correction to my review above: I said this client **cannot send** `verifiedOn`. That is wrong.

The encoder is `JSON.stringify` and `callPower` takes `unknown`, so a caller could always put the field on the wire whether or not `Curated` declared it. I found out writing the fix in #118 — my first round-trip test passed before the change.

The gap is real but narrower than I described: type-level, not runtime. Nobody writing against `Curated` could see the field existed, and a properly-typed call site would have been rejected for using it. Everything else in the review stands.

Issue #112, opened 2026-08-28
Store discipline: no network, nothing marked for sync


jonstump:
Part of #110. Governed by [SPEC-0009](docs/openspec/specs/durable-store/spec.md).

**Blocked by #111** — there is nothing to check until the store exists.

Two requirements that are both absences, and absences cannot be covered by exercising paths. A behavioural test shows that the paths someone thought of behave; the failure mode is always the path nobody thought of.

Extends `web/tests/helpers/source-checks.ts`, which already does exactly this for the boundary client — message-text branching, numeric conversion, navigation — each checker run against the real sources and against a snippet containing the mistake it catches.

## Requirements

- [ ] **Stage 1 Reaches No Network** (SPEC-0009) — Nothing in this capability MUST issue a network request. The absence MUST be checkable mechanically rather than by review, and the check MUST assert on this capability's own call paths rather than on the absence of network code in the application bundle.
  - WHEN the store is opened, read from, and written to THEN no network call of any kind is issued
  - WHEN the application also contains code that makes network requests for unrelated reasons THEN the assertion still holds and still fails if a request is added to a store path

- [ ] **Nothing Is Marked for Synchronization** (SPEC-0009) — No record written in stage 1 MUST be marked shared, synced, or pending upload. Where the schema carries fields serving stages 2 and 3 they MUST be present and unset rather than absent.
  - WHEN any place is written THEN it carries no flag marking it shared, synced, or pending upload

## Acceptance Criteria

- **The network check asserts on store call paths, not on the bundle.** The bundle will contain network code by design once ADR-0008 stage 2 ships, and a check phrased against the bundle would then be weakened to accommodate it — reporting a guarantee it has stopped checking. This is the same constraint SPEC-0008 places on ADR-0002's parse-path test, for the same reason
- Each checker is run against a snippet containing the mistake it catches, so a checker that has quietly stopped matching fails here rather than passing forever. `source-checks.ts` already establishes the pattern and its negative controls
- The sync check asserts on **written records**, not on source. A field set by a code path the checker did not read is still a field that was set
- The checks run in CI and are added to `scripts/mutation-check.sh`, which stands at 21 mutations covering the stylesheet and the accessibility primitives
- A mutation that adds a `fetch` to a store path turns the suite red. A mutation that pre-marks a record as synced turns it red

## Notes

[...]

Issue #110, opened 2026-08-28
Implement Durable Store


jonstump:
Governed by [SPEC-0009](docs/openspec/specs/durable-store/spec.md) ([design](docs/openspec/specs/durable-store/design.md)), realizing [ADR-0008](docs/adrs/ADR-0008-durable-user-data-store.md) **stage 1**.

Where durable, player-authored data lives on the device: ticked construction items, stocked quantities, notes, tags, player-assigned names, and the view preferences that currently forget themselves on reload.

Stage 1 only — local IndexedDB, no account, no server, no network. ADR-0008's stages 2 and 3 are ADR-0009 and ADR-0010's, and nothing in this epic implements them.

## Stories

| Issue | Story | Requirements | Depends on |
|---|---|---|---|
| #111 | The store: schema, versioning, and operations `foundation` | A Workspace Owns Places; A Place Is One Record Type, Whatever Its Kind; Versioned, and Fails Legibly; Error Handling Standards; Storage Operation Standards | — |
| #112 | Store discipline: no network, nothing marked for sync | Stage 1 Reaches No Network; Nothing Is Marked for Synchronization | #111 |
| #113 | Preferences persist, and an empty store is a designed state | View Preferences Survive a Reload; An Empty Store Is a Designed State | #111 |
| #114 | Data custody: deletion, eviction honesty, and images | Deletion Is a First-Class Operation; Storage Is Evictable and the Application Must Not Imply Otherwise; Screenshots Are Local-Only | #111 |
| #115 | Tests: preferences, empty state, deletion, and custody | covers #113 and #114 | #113, #114 |

All twelve of the spec's requirements appear in exactly one story.

## Greenfield, verified

`grep -rn "localStorage\|sessionStorage\|indexedDB\|IDBDatabase" web/src web/tests` returns nothing. The design README's storage prohibition is intact in code, not merely on paper, and ADR-0008 lifts it for durable data while keeping it for plan state. **#111 is the first code in this project that legitimately holds durable state**, and the rules `ViewState` and `ResultCache` were built to enforce must not be loosened by its arrival.

## What this unblocks

[...]

Issue #96, opened 2026-08-28
Card composition, identity, and producer sections


jonstump:
Part of #94. Governed by [SPEC-0007](docs/openspec/specs/base-planner-card/spec.md).

**Blocked by #95** — the card has no payload to compose from until `rollup` crosses.

The card's skeleton: what it is, what it says it is, and the construction rows that are the point of it.

## Requirements

- [ ] **Card Composition From the Build Payload** (SPEC-0007) — Each card MUST be composed from the stage 2 and stage 3 payloads for its base, obtained through the SPEC-0005 boundary client. The card MUST NOT aggregate, total, or reconcile figures across bases, and MUST NOT derive one base's figures from another's. Where a figure is a count the domain reports — plants, biodomes, extractors, depots, fauna, nutrient processors, pellet feeders, additional generators, batteries — the card MUST render the domain's value and MUST NOT compute a count from a quantity and a rate. Base-level counts MUST NOT be re-derived per row.
  - WHEN a card renders a farm row showing plants and biodomes THEN both numbers came from the payload, and neither was computed in the view from a quantity and a yield
  - WHEN a base carries three kitchen steps THEN the card shows the base's reported nutrient processor count once, and does not sum a per-step figure

- [ ] **Base Identity and Selection** (SPEC-0007) — A card MUST identify its base by name as the primary identifier, with the base's colour as reinforcement rather than the sole carrier. The card's frame is reserved for base identity; selection MUST be rendered as an inboard ring using an overlay element. A card MUST be selectable using a control with the correct semantics for a selection, operable by keyboard. A generic element given a tab index MUST NOT be used.
  - WHEN a card is selected THEN its base-colour frame is unchanged and the selection is shown by an inboard ring
  - WHEN base colours are not perceivable THEN each card remains identifiable by its name

- [ ] **Producer Sections** (SPEC-0007) — A card MUST present its construction rows grouped by producer type, covering every type the payload reports: farm, extractor, ranch and kitchen. Each row MUST state what is built and how much is required, and MUST NOT present only the required quantity. Where the payload reports a range rather than a single value, the card MUST NOT present the optimistic bound as the planning figure.
  - WHEN a base's payload carries farm, extractor, ranch and kitchen rows THEN the card presents all four groups
  - WHEN a farm row requires a quantity of a crop THEN the card shows the plants and biodomes to build, not the required quantity alone

[...]

PR #119, opened 2026-08-29
feat(web): the durable store

### LINKED ISSUE #111: The store: schema, versioning, and operations

The store itself. Every other story in this epic reads and writes through it.

Greenfield — `grep -rn "localStorage\|sessionStorage\|indexedDB\|IDBDatabase" web/src web/tests` returns nothing. **This is the first code in the project that legitimately holds durable state**, and the rules `ViewState` and `ResultCache` were built to enforce must not be loosened by its arrival. `ViewState` has no field a graph could go in; `ResultCache` holds one deep-frozen entry. Neither changes here.

New under `web/src/store/`. It sits beside `web/src/boundary/`, not inside it: the boundary is the route to domain values and this is the route to the player's own data, and merging them would put a note inside the thing that decodes an envelope.

## Requirements

- [ ] **A Workspace Owns Places** (SPEC-0009) — The store MUST hold exactly one workspace per device, carrying a `schemaVersion`, an `ownerId`, and a collection of place records. `ownerId` MUST be nullable and MUST be null in stage 1, and MUST be present in the schema from the first version rather than added later.
  - WHEN the store is opened on a device that has never used it THEN a workspace exists with a null `ownerId`, the current `schemaVersion`, and no places
  - WHEN a workspace written in stage 1 is inspected THEN `ownerId` is present and null, rather than the field being absent from the record

- [ ] **A Place Is One Record Type, Whatever Its Kind** (SPEC-0009) — A place MUST be represented by a single record type covering bases, freighters and settlements. Each place MUST carry a stable `id` generated at creation, independent of any save file and independent of any account, plus `updatedAt` and a monotonically increasing `revision`.
  - WHEN a freighter and a base are both stored THEN both are place records distinguished by a kind field, and no second record type exists
  - WHEN a place is created by save import and the same place is later re-imported from a changed save THEN the place's `id` is unchanged, because it was generated at creation rather than derived from save contents
  - WHEN a place is modified twice THEN its `revision` is strictly greater after the second write than after the first, and `updatedAt` reflects the later write

- [ ] **Versioned, and Fails Legibly** (SPEC-0009) — The store MUST carry a `schemaVersion` at the workspace and at each place. An unrecognized version MUST load **nothing** and MUST report it. The store MUST NOT apply a partial load, MUST NOT drop records it cannot read while keeping ones it can, and MUST NOT migrate silently.
  - WHEN the store contains a workspace whose `schemaVersion` is higher than the running build understands THEN no place is loaded, and a diagnostic states both versions
  - WHEN one place record carries an unreadable `schemaVersion` and the rest do not THEN the load fails as a whole rather than returning the readable subset

- [ ] **Error Handling Standards** (SPEC-0009) — Errors MUST be wrapped with context naming the workspace or place being read or written. Sentinel errors MUST be defined for unsupported version, quota exceeded, storage unavailable, and record not found, and MUST be selectable by identity rather than by message text. Silent error swallowing MUST NOT occur — a quota failure in particular MUST NOT be discarded. Structured logging MUST be used.
  - WHEN a write fails because the storage quota is exceeded THEN the failure is returned to the caller and surfaced, and the application does not indicate that the change was stored
  - WHEN the application distinguishes an unsupported version from unavailable storage THEN it selects on the sentinel error, and no source or test matches on message text

- [ ] **Storage Operation Standards** (SPEC-0009) — Transactions MUST be used for multi-step mutations requiring atomicity. Connection lifecycle MUST be explicitly managed, with the database opened once and version-change events handled rather than ignored. The parameterized-query requirement does not apply and is recorded as considered.
  - WHEN a write that touches a place and the workspace fails partway THEN neither change is applied, and the store is left as it was
  - WHEN another tab holds the database open at an older version and an upgrade is attempted THEN the version-change event is handled and the condition is surfaced, rather than the load blocking indefinitely

## Acceptance Criteria

- **`ownerId`, `updatedAt` and `revision` are written even though nothing reads them.** Asserted, not left to judgement. ADR-0008's migration works only because they exist from version 1, and a reviewer who has not read that ADR will reasonably ask why they are there
- A `schemaVersion` the build does not understand loads **zero** places and names both versions. Tested with a workspace-level mismatch and a single-place mismatch, because the second is the one that tempts a partial load
- A quota failure is returned, not swallowed. This is the one storage failure that presents as success — a write that did not happen, reported as one that did — so it gets its own test rather than relying on the general no-swallowing rule
- A version-change event from another tab is handled. Untested, it presents as the application hanging on load, which points nowhere near the cause
- A multi-step write that fails partway leaves the store as it was
- Error selection is by sentinel identity. `tests/boundary/discipline.spec.ts` already forbids message-text branching in the boundary; the same discipline applies here and #96 makes it mechanical
- The per-place size bound from SPEC-0009 § Request Body Size Limits is **enforced and its value recorded with the measurement it came from.** The spec requires it be measured rather than guessed; ADR-0008 measured the text case at 596 KB for 200 places and excluded images, so the bound must say whether images are in scope
- Tests are in this story rather than a companion: library code with no rendered surface, matching #59 and #95

### Branch

`feat/111-store-schema-versioning-and-operations`

### PR Convention

```
Closes #111
Part of #110

Implements SPEC-0009 REQ "A Workspace Owns Places", REQ "A Place Is One Record Type, Whatever Its Kind", REQ "Versioned, and Fails Legibly", REQ "Error Handling Standards", REQ "Storage Operation Standards".
```

jonstump:
Closes #111
Part of #110

Implements SPEC-0009 REQ "A Workspace Owns Places", REQ "A Place Is One Record Type, Whatever Its Kind", REQ "Versioned, and Fails Legibly", REQ "Error Handling Standards", REQ "Storage Operation Standards".

SPEC-0009 stage 1: local IndexedDB, no account, no server, no network. **The first code in this project that legitimately holds durable state.**

`web/src/store/` — 20 new tests, all in a real browser because IndexedDB exists nowhere else and a fake would be a test of the fake.

## The rules that kept state out are not loosened

`ViewState` still has no field a graph could go in. `ResultCache` still holds one deep-frozen entry. Nothing here stores a plan, a resolved graph, or a derived quantity — and the message-text discipline check now scans `src/store` alongside `src/boundary`, so the store is held to the same rule by the same checker rather than a second copy of it.

IndexedDB rather than localStorage, and the distinction matters: the design README's prohibition names localStorage *specifically*, so satisfying its letter while overturning its spirit would have been the worse outcome. localStorage is synchronous on the thread the WASM module runs on, string-only, and capped near 5 MB.

## Three fields nothing reads, asserted rather than assumed

`ownerId`, `updatedAt`, `revision` — written from version 1, unset or trivially set. A test asserts each, because a reviewer who has not read ADR-0008 will reasonably ask why. Sign-in attaches an owner to a workspace that *already carries the field*; a field added in version 2 cannot do that for data written under version 1, and the player who ticked fifty items before signing up is who that migration is for.

## The size bound is measured, and recorded beside itself

64 KiB, from real measurement against the 108-entry parts catalog with natural-language notes rather than compressible filler:

| | ticks | stocked | tags | bytes |
|---|---:|---:|---:|---:|
| typical | 40 | 20 | 4 | 1,677 |
| every part ticked | 108 | 20 | 4 | 2,966 |
| heavy | 108 | 200 | 20 | 6,156 |

A test asserts the bound keeps **an order of magnitude** over the heaviest measured place, so changing the number without revisiting the measurements fails. There's a companion test that a place at the heavy size is comfortably accepted — a bound that refused real data would pass the refusal test and make the store unusable. Images are explicitly out of scope; ADR-0013 owns them, and this is a text bound a store holding blobs must revisit rather than inherit.

## Two failures named individually

**An oversized write is refused before it is attempted.** SPEC-0009 requires this because a quota failure is shared with everything else the origin stores, and discovering the limit that way is worse than being told.

**The `versionchange` handler closes this connection** so another tab's upgrade proceeds. Unhandled, that leaves the other tab's open pending forever and presents as the application hanging on load — a symptom pointing nowhere near its cause, which is exactly why the spec names it. Tested by opening a raw connection at a higher version and asserting it upgrades rather than blocking.

## The tests found a real bug on their first run

`load()` treated a missing workspace record as "a device that has never used the store" and returned an empty workspace — **silently dropping every place present**.

That is worse than the partial load SPEC-0009 forbids: it is a total one, reported as success. A missing workspace record with places present is now a malformed store, and the test that caught it is kept as its own case rather than folded away.

## And the linter surfaced a design smell worth fixing

`readPlace` returned `PlaceRecord | StoreFailure`. Both carry a `kind` field — a place's is its type, a failure's is the literal `"failed"` — so narrowing on it happened to work and read as though it were designed. Either type changing would have broken it silently. It returns a `StoreResult` now, which is what made the assertion the linter flagged unnecessary in the first place.

## What is honestly not tested

**Real quota exhaustion.** It is tested through the pure `classify()` function — `QuotaExceededError` → `QUOTA_EXCEEDED`, `InvalidStateError`/`SecurityError` → `STORAGE_UNAVAILABLE`, and an unrecognised name → `UNCLASSIFIED`. Filling a disk to observe the real condition is not a test anyone runs twice, and the question worth answering is whether the error reaches the caller as the right sentinel. The end-to-end path is covered by the size bound, which SPEC-0009 requires precisely so a quota failure is not how a player finds out.

## Note on #112

#111's acceptance criteria said "#96 makes it mechanical" about the message-text check. That was a wrong issue number in the planning — #96 is a SPEC-0007 card story. The mechanical check belongs to **#112** (store discipline), and since extending the existing checker to `src/store` was three lines, I did it here rather than leave the criterion unmet. #112 still owns the network and sync-flag checks, which are the substantial half.

## Verified locally

`npm run lint`, `lint:css`, `typecheck`, `format:check`, `check:tokens`, `npm run build`, `npm test` (162 passed, up from 142), `npm run test:mutation` (21/21 still red).

Pushed over HTTPS — the SSH agent is still refusing. No workflow files touched.

### DIFF
--- a/web/src/store/durable-store.ts
+++ b/web/src/store/durable-store.ts
@@ -1,0 +1,360 @@
+/*
+ * The local store.
+ *
+ * Governing: ADR-0008 (durable user data lives in a local-first store),
+ * SPEC-0009 REQ "A Workspace Owns Places", REQ "A Place Is One Record Type,
+ * Whatever Its Kind", REQ "Versioned, and Fails Legibly", REQ "Error
+ * Handling Standards", REQ "Storage Operation Standards", REQ "Stage 1
+ * Reaches No Network"
+ *
+ * IndexedDB, not localStorage. The design README's prohibition names
+ * localStorage specifically and ADR-0008 lifts it for durable data only —
+ * plan state keeps the rule. localStorage would have been the letter of it
+ * and not the spirit: synchronous on the thread the WASM module runs on,
+ * string-only, and capped around 5 MB.
+ *
+ * This is the first code in the project that legitimately holds durable
+ * state, and it must not become the back door the other rules were keeping
+ * shut. `ViewState` still has no field a graph could go in and `ResultCache`
+ * still holds one deep-frozen entry. Nothing here stores a plan, a resolved
+ * graph, or any derived quantity.
+ */
+
+import { classify, failure, ok, type StoreResult } from "./errors";
+import { MAX_PLACE_BYTES, serializedBytes } from "./limits";
+import {
+  emptyWorkspace,
+  isPlaceKind,
+  SCHEMA_VERSION,
+  type PlaceRecord,
+  type Workspace,
+  type WorkspaceRecord,
+} from "./schema";
+
+const DATABASE = "nms-planner";
+
+/*
+ * The IndexedDB version, which is not the schema version.
+ *
+ * IndexedDB's version governs the object stores; `schemaVersion` on each
+ * record governs the shape of what is in them. They move independently: a
+ * record shape can change without adding a store, and the version failure
+ * SPEC-0009 requires is about the records, not the stores.
+ */
+const DB_VERSION = 1;
+
+const WORKSPACE_STORE = "workspace";
+const PLACES_STORE = "places";
+/** The workspace is a singleton; this is its key. */
+const WORKSPACE_KEY = "self";
+
+function request<T>(source: IDBRequest<T>): Promise<T> {
+  return new Promise((resolve, reject) => {
+    source.addEventListener("success", () => {
+      resolve(source.result);
+    });
+    source.addEventListener("error", () => {
+      reject(source.error ?? new Error("request failed"));
+    });
+  });
+}
+
+function settled(transaction: IDBTransaction): Promise<void> {
+  return new Promise((resolve, reject) => {
+    transaction.addEventListener("complete", () => {
+      resolve();
+    });
+    transaction.addEventListener("error", () => {
+      reject(transaction.error ?? new Error("transaction failed"));
+    });
+    transaction.addEventListener("abort", () => {
+      reject(transaction.error ?? new DOMException("aborted", "AbortError"));
+    });
+  });
+}
+
+/**
+ * Validate a stored place against the shape this build reads.
+ *
+ * Returns a StoreResult rather than `PlaceRecord | StoreFailure`. Both of
+ * those carry a `kind` field — a place's is its type, a failure's is the
+ * literal "failed" — so narrowing on it happened to work and read as though
+ * it were designed. One of them changing would have broken it silently.
+ */
+function readPlace(value: unknown): StoreResult<PlaceRecord> {
+  if (typeof value !== "object" || value === null) {
+    return failure("MALFORMED_RECORD", "a stored place is not an object");
+  }
+  const raw = value as Record<string, unknown>;
+
+  const version = raw["schemaVersion"];
+  if (typeof version !== "number") {
+    return failure("MALFORMED_RECORD", "a stored place carries no schemaVersion");
+  }
+  if (version > SCHEMA_VERSION) {
+    return failure(
+      "UNSUPPORTED_VERSION",
+      `a stored place is schema version ${String(version)}, and this build reads ${String(SCHEMA_VERSION)}`,
+    );
+  }
+
+  if (typeof raw["id"] !== "string" || !isPlaceKind(raw["kind"])) {
+    return failure("MALFORMED_RECORD", "a stored place has no usable id or kind");
+  }
+  if (typeof raw["updatedAt"] !== "string" || typeof raw["revision"] !== "number") {
+    return failure(
+      "MALFORMED_RECORD",
+      `place ${raw["id"]} is missing updatedAt or revision`,
+    );
+  }
+
+  return ok(raw as unknown as PlaceRecord);
+}
+
+export interface StoreOptions {
+  /** Injected so tests can drive time. Defaults to the wall clock. */
+  readonly now?: () => string;
+  /** Injected so a test can use its own database rather than the app's. */
+  readonly databaseName?: string;
+}
+
+export class DurableStore {
+  readonly #databaseName: string;
+  readonly #now: () => string;
+  #db: IDBDatabase | null = null;
+
+  constructor(options: StoreOptions = {}) {
+    this.#databaseName = options.databaseName ?? DATABASE;
+    this.#now = options.now ?? (() => new Date().toISOString());
+  }
+
+  /**
+   * Open the database, creating its stores on first use.
+   *
+   * Idempotent. The `versionchange` handler is the load-bearing part: when
+   * another tab opens the database at a higher version, this connection is
+   * told and closes. Without it that tab's upgrade blocks indefinitely and
+   * presents as the application hanging on load — a symptom pointing
+   * nowhere near its cause, which is why SPEC-0009 names it.
+   */
+  async open(): Promise<StoreResult<void>> {
+    if (this.#db) return ok(undefined);
+
+    if (typeof indexedDB === "undefined") {
+      return failure("STORAGE_UNAVAILABLE", "this browser exposes no IndexedDB");
+    }
+
+    try {
+      const opening = indexedDB.open(this.#databaseName, DB_VERSION);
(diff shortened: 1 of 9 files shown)


jonstump:
## Review — approved

Reviewed as a third party. Base is `main`, no conflict markers, 9/9 CI, +1283/-0.

### Claims I checked rather than took on trust

**The size bound's margin is genuinely load-bearing.** 64 KiB against the 6,156-byte heavy case is 10.65×, and the test asserts `> 10`. That is a 6.5% margin — tight enough that raising the bound or the measurements without revisiting the other actually fails, which is the stated intent rather than an accident.

`limits.ts` records the derivation as SPEC-0009 requires — the three measured shapes, the 108-entry catalog they came from, natural-language notes rather than compressible filler, and the earlier pass where `"x" * 400` reported figures four times too small. Images are scoped out explicitly with ADR-0008's numbers (1.5–3 MB per capture against 596 KB for a 200-place text workspace), so a store that later holds blobs has to revisit this rather than inherit it. That is the honest form of "measured, not guessed": the measurements are constants, but their provenance and methodology travel with them so a future change can argue with them.

**Both version-mismatch tests are non-vacuous.** The workspace case plants a place *first*, so "loads nothing" means something. The single-place case — the one #111 calls out as tempting a partial load — plants one good and one future place and asserts the whole load fails, with the assertion message `"a readable subset was returned"`. Its comment names the reason precisely: a workspace missing four of eleven bases is indistinguishable to the player from a complete one, and they plan against what is there.

**The discipline check really does cover the store.** `discipline.spec.ts` gained a `STORE` path constant and three tests, including `expect(sources.length, "the store sources were not found").toBeGreaterThanOrEqual(4)` — the guard that stops the check silently scanning nothing, which is the failure #78's stylesheet spec had to fix after a hardcoded file list nearly went stale.

There is also a test the body does not claim: `no store source turns a stored quantity into a number`, extending SPEC-0005's no-float rule to stored values. Worth having and worth mentioning, since nobody reading the PR description would know it is there.

### The bug the tests found is the most valuable thing here

`load()` treating a missing workspace record as "never used the store" and returning an empty workspace — silently dropping every planted place — is worse than the partial load SPEC-0009 forbids, and the PR says so in exactly those terms: a total loss reported as success. Keeping it as its own named test rather than folding it into a general malformed-store case is right; that specific shape is the one that will be re-introduced by someone reasoning about first-run behaviour.

### The disclosure about what is not tested is the right call

Real quota exhaustion is covered through the pure `classify()` function rather than by filling a disk. That is the correct trade and it is stated rather than glossed: the question worth answering is whether `QuotaExceededError` reaches the caller as `QUOTA_EXCEEDED` rather than `UNCLASSIFIED`, and the end-to-end path is covered by the size bound, which SPEC-0009 requires precisely so a quota failure is not how a player finds out.

### On the #96 / #112 correction

The correction is right — #96 is "Card composition, identity, and producer sections" under epic #94 (SPEC-0007), so #111's criterion citing it was a planning error. Extending the existing checker rather than leaving a criterion unmet on a wrong reference is the correct handling, and leaving #112 the substantial half (network and sync-flag checks) keeps that story worth doing.

### Nothing requested

The rules that kept state out are visibly not loosened: `ViewState` unchanged, `ResultCache` unchanged, and nothing here stores a plan, a resolved graph, or a derived quantity. The IndexedDB-over-localStorage reasoning — that the design README's prohibition names localStorage specifically, so satisfying its letter while overturning its spirit would be the worse outcome — is the right way to read a prohibition.

Approving and merging.

Issue #114, opened 2026-08-28
Data custody: deletion, eviction honesty, and images


jonstump:
Part of #110. Governed by [SPEC-0009](docs/openspec/specs/durable-store/spec.md).

**Blocked by #111.**

What the player is told about their data, and what they can do about it. Three requirements that are all about custody rather than storage.

## Why deletion is owed now

ADR-0002 banked "needs no upload endpoint, retention policy, or deletion story" as a benefit of holding nothing. This epic spends the third of those the moment the first byte is written: once data is held, the player is owed a way to remove it that is not browser developer tools.

Retention genuinely is not owed yet — nothing retains anything, because nothing leaves the device. That is ADR-0011's, at stage 2.

## Requirements

- [ ] **Deletion Is a First-Class Operation** (SPEC-0009) — The application MUST offer the player a way to delete everything this store holds, without requiring browser developer tools or a manual site-data clear. Deletion MUST remove the workspace and every place in it, MUST be confirmed before it runs, and MUST leave the application in the designed empty state rather than in an error state.
  - WHEN the player chooses to delete stored data and confirms THEN the workspace and every place are removed, and the application presents the empty state
  - WHEN the deletion control is activated THEN the action is confirmed before anything is removed

- [ ] **Storage Is Evictable and the Application Must Not Imply Otherwise** (SPEC-0009) — The application MUST NOT present locally stored data as guaranteed to persist. Where it indicates that data is stored, the indication MUST be accurate about its scope — on this device — and MUST NOT use language implying a backup or a guarantee.
  - WHEN the application indicates that a change has been stored THEN the indication does not claim the data is backed up or synchronized

- [ ] **Screenshots Are Local-Only** (SPEC-0009) — Where the store holds an image, it MUST be held locally and MUST NOT be shareable or synchronizable. Until ADR-0013 exists, the application MUST NOT offer a control to share or upload an image.
  - WHEN a place holds an image THEN the application offers no control whose effect would be to share or upload it

## Acceptance Criteria

- Deletion is reachable from the application. A player who wants their data gone must not need developer tools
- Deletion is confirmed first, and the confirmation is a dialog using **the shell's focus trap** from #82 rather than its own — the trap restores focus on every close route because the restore lives in the effect cleanup, and a bespoke dialog would lose that
- After deletion the application is in the designed empty state, not an error state. Deleting your data is a thing you chose, not a fault
[...]

PR #120, opened 2026-08-29
test(web): store discipline — no network, nothing marked for sync

### LINKED ISSUE #112: Store discipline: no network, nothing marked for sync

jonstump:
Closes #112
Part of #110

Implements SPEC-0009 REQ "Stage 1 Reaches No Network", REQ "Nothing Is Marked for Synchronization".

## Two absences, two layers

Both requirements are claims that something never happens, and exercising a path only shows that the paths someone thought of behave. So each is checked twice, and the layering is the point rather than belt-and-braces.

**The source scan** (`networkCalls`, `syncMarkers` in `tests/helpers/source-checks.ts`) is the weaker half. It exists because it names the offending line.

**The runtime check** (`tests/store/discipline.spec.ts`) is what carries the requirement. A request can be issued through a reference no regex can see; a field can be written under a computed key no source scan reads.

## Why the network check is phrased against call paths

The acceptance criterion is explicit that a check phrased against the bundle would have to be weakened once ADR-0008 stage 2 ships network code by design — and would then go on reporting a guarantee it had stopped making.

So each recorded attempt carries the stack it was made from, and the question asked is *did this come from a store path*, not *did this happen*. Three tests:

1. A quiet page runs every store call path and records **no attempt of any kind**. This one needs no stack and is where the teeth are.
2. A page with unrelated traffic records that traffic and still attributes nothing to a store path — the requirement's second scenario.
3. The source scan.

A limit stated in the code rather than papered over: a continuation resumed after `await` depends on the engine's async stack tagging. Test 1 is why the suite does not rest on attribution alone.

## The sync check reads disk, not source

Records are read back through a raw IndexedDB connection rather than `DurableStore.load`, because the criterion is about what ends up written. A field the store's own reader drops on the way out is still a field on disk.

`updatedAt` and `revision` are deliberately **not** treated as sync markers. ADR-0012 reserves both and SPEC-0009 requires they be "present and unset rather than absent"; a checker that read a reserved-and-empty field as a marked one would be arguing with the schema. A separate test asserts `ownerId` is present and null.

## Mutations

`scripts/mutation-check.sh` goes 21 → 24, all verified red:

| Mutation | Catches |
|---|---|
| the store fetches something on write | source scan + runtime |
| **the store fetches through a name the source scan cannot see** | runtime only |
| a written place is pre-marked as synced | source scan + runtime |

The middle one earns the runtime layer its place. Verified by hand before adding it: with `["fet" + "ch"]` applied, *"no store source reaches for a network primitive"* **passes** and *"nothing goes out"* **fails**. Delete the runtime check and that mutation goes green.

Both store mutations are type errors `tsc` would reject. That is not a gap — vite strips types without checking them, so the mutated store runs, and a rule only the compiler enforces stops being enforced the moment someone reaches for `as unknown as`.

## Gates

lint, lint:css, typecheck, format:check, check:tokens, build all clean. **171 tests pass** (up from 162). **24/24 mutations red.**

No protected paths touched.

### DIFF
--- a/web/scripts/mutation-check.sh
+++ b/web/scripts/mutation-check.sh
@@ -34,8 +34,9 @@ TRAP="src/a11y/useFocusTrap.ts"
 LIVE="src/a11y/useLiveRegion.ts"
 SHELL="src/shell/AppShell.tsx"
 BADGE="src/shell/StatusBadge.tsx"
+STORE="src/store/durable-store.ts"
 
-MUTABLE="$BASE $TOKENS $TRAP $LIVE $SHELL $BADGE"
+MUTABLE="$BASE $TOKENS $TRAP $LIVE $SHELL $BADGE $STORE"
 
 # shellcheck disable=SC2086
 restore() { git checkout -- $MUTABLE 2>/dev/null || true; }
@@ -218,6 +219,54 @@ check "a control is named only by a glyph" \
   '        <button type="submit" className="control control-primary interactive">
           →
         </button>'
+
+# ----------------------------------------------------------------------
+# The durable store's two absences.
+#
+# SPEC-0009 REQ "Stage 1 Reaches No Network" and REQ "Nothing Is Marked for
+# Synchronization" are both claims that something never happens, and a suite
+# that only exercises the store cannot distinguish "no request was issued"
+# from "no test looked". These two are the ones issue #112 names.
+#
+# Both mutations are type errors the compiler would reject. That is not a
+# gap: vite strips types without checking them, so the mutated store runs,
+# and a rule that only `tsc` enforces is a rule that stops being enforced
+# the moment someone reaches for `as unknown as`.
+# ----------------------------------------------------------------------
+
+check "the store fetches something on write" \
+  tests/store/discipline.spec.ts "$STORE" \
+  '      const now = this.#now();
+      const transaction = db.transaction([WORKSPACE_STORE, PLACES_STORE], "readwrite");' \
+  '      const now = this.#now();
+      await fetch("/api/places", { method: "POST", body: place.id });
+      const transaction = db.transaction([WORKSPACE_STORE, PLACES_STORE], "readwrite");'
+
+# The one that earns the runtime layer its place. Both mutations above trip
+# the source scan *and* the runtime check, so neither shows the runtime check
+# is doing anything. This one is invisible to any regex — verified: with it
+# applied, "no store source reaches for a network primitive" passes and
+# "nothing goes out" fails. Delete the runtime check and this goes green.
+check "the store fetches through a name the source scan cannot see" \
+  tests/store/discipline.spec.ts "$STORE" \
+  '      const now = this.#now();
+      const transaction = db.transaction([WORKSPACE_STORE, PLACES_STORE], "readwrite");' \
+  '      const now = this.#now();
+      const send = (globalThis as Record<string, unknown>)["fet" + "ch"] as (
+        url: string,
+      ) => Promise<unknown>;
+      await send("/api/places");
+      const transaction = db.transaction([WORKSPACE_STORE, PLACES_STORE], "readwrite");'
+
+check "a written place is pre-marked as synced" \
+  tests/store/discipline.spec.ts "$STORE" \
+  "      const record: PlaceRecord = {
+        ...place,
+        schemaVersion: SCHEMA_VERSION," \
+  "      const record: PlaceRecord = {
+        ...place,
+        synced: false,
+        schemaVersion: SCHEMA_VERSION,"
 
 echo
 if [ "$failures" -gt 0 ]; then
--- a/web/tests/fixtures/store-discipline-harness.ts
+++ b/web/tests/fixtures/store-discipline-harness.ts
@@ -1,0 +1,229 @@
+/*
+ * Fixture for the store's two discipline requirements.
+ *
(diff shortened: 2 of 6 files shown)


jonstump:
## Review — approved, one citation to correct

Base `main`, no conflict markers, 9/9 CI, +685/-1. Tier 0 with #121 — no file overlap, so either order.

### The middle mutation is what makes this PR worth its size

I verified it exists and does what the body says:

```
check "the store fetches through a name the source scan cannot see" \
  const send = (globalThis as Record<string, unknown>)["fet" + "ch"] as (
```

A computed key no regex can see. That is the mutation that earns the runtime layer its place, and running it by hand before adding it — confirming the source scan goes green while `"nothing goes out"` goes red — is the right way to establish that two checks are two checks rather than one written twice.

The three-test structure is also right, and test 1 is correctly identified as where the teeth are: a quiet page recording **no attempt of any kind** needs no stack attribution and therefore does not inherit attribution's limits. Stating that limit in the code — that a continuation resumed after `await` depends on the engine's async stack tagging — rather than leaving it for someone to discover is the difference between a known bound and a surprise.

Phrasing against call paths rather than the bundle is the same reasoning SPEC-0008 REQ "The Save Never Leaves the Device" gives, and for the same reason: ADR-0008 stage 2 ships network code by design, and a bundle-phrased check would be weakened then and go on reporting a guarantee it had stopped making.

### Reading records back through a raw connection is the right call

"A field the store's own reader drops on the way out is still a field on disk" — exactly. A sync check that went through `DurableStore.load` would be testing the reader, not the writer.

### The correction: ADR-0012 does not exist

Two code comments cite it:

- `tests/helpers/source-checks.ts:135` — "ADR-0012 reserves both"
- and a second at the reserved-fields test — "ADR-0012 has `updatedAt` and `revision` written"

`docs/adrs/` holds ADR-0001 through ADR-0008. ADR-0012 is a **planned** ADR that ADR-0008 forward-references in its table at line 244: `| ADR-0012 | Multi-device sync and conflict resolution — schema room is reserved here (updatedAt, revision) |`.

So the reservation is ADR-0008's, and the normative requirement is SPEC-0009 line 49: *"`updatedAt` and `revision` MUST be written from the first version even though stage 1 has nothing to reconcile."* The second comment is the one that reads oddly — ADR-0012 does not "have" those fields written, because ADR-0012 has nothing.

**The substance is right** and the decision it justifies is right: treating a reserved-and-empty field as a sync marker would flag every record #111 writes, which would make the check useless rather than strict. Only the citation is off, and it matters because `/sdd:audit` traces governing comments and a reference to a nonexistent artifact is noise in the tool meant to catch drift — the same class as #93's inherited REQ reference.

Merging rather than blocking; two comment lines, and the reasoning they support is sound. Worth correcting whenever `src/store` is next touched, or in #114.

### On the type-error note

"Both store mutations are type errors `tsc` would reject. That is not a gap — vite strips types without checking them, so the mutated store runs, and a rule only the compiler enforces stops being enforced the moment someone reaches for `as unknown as`." That is exactly right, and worth having written down: it is the argument for why a runtime check earns its keep in a strictly-typed codebase.

PR #121, opened 2026-08-29
feat(web): preferences persist, and an empty store is a designed state

### LINKED ISSUE #113: Preferences persist, and an empty store is a designed state

**Blocked by #111.**

The first consumer, and the contract every later consumer inherits.

Extends `web/src/state/view-state.ts` and `ViewStateProvider.tsx`. `ViewState.preferences` holds exactly two values today — `groupSeparator` and `showUnverified` — and they last one page load. SPEC-0005 permits the view to hold them; nothing lets them survive a reload, and a preference that forgets itself is not a preference.

## The line this story must not cross

Persisting preferences MUST NOT move them out of view state. They remain interface state the view owns; the store is where the view's own copy is written and read.

This matters because a persistence layer is the obvious place for the plan to end up, and `ViewState` having no field a graph could go in is the thing that has kept it out. A store wired in as a general state container is the back door those rules were keeping shut.

## Requirements

- [ ] **View Preferences Survive a Reload** (SPEC-0009) — View-local preferences that SPEC-0005 REQ "View State Boundaries" permits the view to hold MUST be persisted through this store and MUST be restored on load. Persisting them MUST NOT move them out of view state — the plan, the resolved graph and every derived quantity MUST remain outside both.
  - WHEN a player changes a view preference and reloads THEN the preference is as they left it
  - WHEN the view's state is inspected after a plan is resolved and preferences are restored THEN it holds selection, collapse, inputs, focus and preferences, and no plan, graph or derived quantity

- [ ] **An Empty Store Is a Designed State** (SPEC-0009) — An empty store MUST be a state the application presents deliberately, not an absence it renders as zeroes. A consumer MUST distinguish "nothing stored" from "stored as zero", and MUST NOT render a figure the player never entered.
  - WHEN the store is opened on a device with no prior data THEN the application presents an intentional empty state, and reports no failure
  - WHEN a place has no stocked quantity recorded for an item THEN the consumer renders the quantity as absent rather than as `0`

## Acceptance Criteria

- A preference set, then reloaded, is as the player left it — driven through a real reload, not by calling the restore function
- **`ViewState` gains no field a plan, graph or derived quantity could occupy.** The shape check `tests/shell/view-state.spec.ts` already runs asserts the key set; it must still pass, and persistence must not be the reason it changes
- Restoring a preference does not widen view state. The existing "no reachable state holds a derived figure" test covers this and must stay green
- A device with no prior data reaches the designed empty state and **reports no failure**. Empty is ordinary — cleared storage, a fresh device, a private window all produce it
- `0` is never rendered for a value the player never entered. Asserted on a place with no stocked quantity, which is the case where absent and zero are easiest to conflate
- The empty-state contract is expressed where later consumers inherit it, rather than reimplemented per surface

### Branch

`feat/113-preferences-persist-and-empty-state`

### PR Convention

```
Closes #113
Part of #110

Implements SPEC-0009 REQ "View Preferences Survive a Reload", REQ "An Empty Store Is a Designed State".
```

jonstump:
Closes #113
Part of #110

Implements SPEC-0009 REQ "View Preferences Survive a Reload", REQ "An Empty Store Is a Designed State".

## The line this story was told not to cross

Persisting preferences must not move them out of view state. `useStoredData` only ever dispatches `setPreference` — the action that already existed — and holds its own bookkeeping in local component state, where it cannot be mistaken for shared view state.

`ViewState` gains no field. The shape check in `tests/shell/view-state.spec.ts` still asserts the same five keys, and it is untouched. That rule is what has kept graphs out of view state; a store wired in as a general state container is the back door it was keeping shut.

Deliberately **not** a bulk `replacePreferences` action — a bulk setter is the natural place for a later caller to pass something that is not a preference.

## Absent is not zero

`src/store/absence.ts` expresses the contract once, because SPEC-0009 requires it be expressed "where later consumers inherit it, rather than reimplemented per surface" — and the per-surface version is always written the same wrong way:

```ts
place.stocked?.[itemId] ?? 0   // reads as a careful default; renders a claim
```

`Stored<T>` is a discriminated union, so `?? 0` does not typecheck against it. Both directions are tested: a place with no stocked entry renders an em dash, and a place with a stored `"0"` renders `0` — a surface that dashed everything would pass the first test and be useless.

The dash is paired with an `aria-label`, because "—" announces as nothing in most screen readers and a row reading as a name followed by silence is indistinguishable from a broken one.

## Three states, not two

| State | Meaning | Reported as |
|---|---|---|
| `loading` | the read has not finished | pending |
| `ready` + `empty` | read, and there was nothing | **not a failure** |
| `unavailable` | IndexedDB blocked or absent | a warning |

Collapsing loading into empty flashes "nothing saved" at a player who has data, which reads as loss rather than latency. Collapsing empty into unavailable is the one SPEC-0009 names directly.

## A defect the tests found

The first draft issued `putPreferences` and dropped the promise. A reload arriving before the IndexedDB transaction committed lost the change — caught by the test that toggles two preferences and reloads at once, which is also exactly what a player does when they set something and leave.

Writes are now chained and their settlement is observable via `data-saving`. That attribute carries **no** user-facing claim; SPEC-0009 REQ "Storage Is Evictable" governs what may be *said* about stored data, and that indication belongs to #114.

## One rename

The preferences checkbox was first labelled "Show unverified figures", which made `getByLabel("Figures")` in `shell.spec.ts` ambiguous. Renamed to "Show unverified" rather than loosening that locator — two controls whose accessible names share a word is a real ambiguity for anyone navigating by label, not only for the test that caught it.

## Verification

**175 tests pass** (up from 162). lint, lint:css, typecheck, format:check, check:tokens, build all clean. **21/21 existing mutations still red.**

Both persistence tests were verified to fail with the write removed — including after the `data-saving` wait was added, so the wait does not mask the defect. The absent-vs-zero tests were verified to fail with `?? "0"` substituted.

**`scripts/mutation-check.sh` is deliberately untouched** to avoid conflicting with #120, which extends the same file. The two mutations this story earns — "preferences are never written" and "absent defaults to zero" — are verified by hand above and owed to that script once #120 lands.

No protected paths touched.

### DIFF
--- a/web/src/shell/AppShell.tsx
+++ b/web/src/shell/AppShell.tsx
@@ -12,7 +12,11 @@ import { Popover } from "../a11y/Popover";
 import { ViewStateProvider } from "../state/ViewStateProvider";
 import { useViewDispatch, useViewState } from "../state/useViewState";
 import { usePlanResolution, type Resolution } from "../state/usePlanResolution";
+import { useStoredData } from "../state/useStoredData";
+import { DurableStore } from "../store";
 import { StatusBadge } from "./StatusBadge";
+import { StoredPlaces } from "./StoredPlaces";
+import { ViewPreferences } from "./ViewPreferences";
 
 /*
  * The shell every surface mounts into.
@@ -190,9 +194,16 @@ function Figures({ resolution }: { resolution: Resolution }): ReactNode {
   );
 }
 
-function Chrome({ client }: { client: BoundaryClient }): ReactNode {
+function Chrome({
+  client,
+  store,
+}: {
+  readonly client: BoundaryClient;
+  readonly store: DurableStore;
+}): ReactNode {
   const state = useViewState();
   const { resolution, resultToken, recompute } = usePlanResolution(client, state);
+  const stored = useStoredData(store);
 
   /*
    * The announcement is keyed to the result token, which changes exactly when
@@ -232,6 +243,26 @@ function Chrome({ client }: { client: BoundaryClient }): ReactNode {
         <section className="panel" aria-label="Figures">
           <Figures resolution={resolution} />
         </section>
+
+        <section className="panel" aria-label="Saved places">
+          <StoredPlaces data={stored} target={state.inputs.target} />
+        </section>
+
+        {/*
+          `data-saving` reports whether a preference write is still in
+          flight. It carries no user-facing claim — SPEC-0009 REQ "Storage
+          Is Evictable" governs what may be *said* about stored data, and
+          that indication is #114's to build. This is the hook for it, and
+          in the meantime it is what lets a test wait for a write to settle
+          rather than sleeping and hoping.
+        */}
+        <section
+          className="panel"
+          aria-label="Preferences"
+          data-saving={String(stored.saving)}
+        >
+          <ViewPreferences />
+        </section>
       </main>
 
       <footer className="shell-footer">
@@ -243,11 +274,26 @@ function Chrome({ client }: { client: BoundaryClient }): ReactNode {
   );
 }
 
-export function AppShell({ client }: { client: BoundaryClient }): ReactNode {
+/**
+ * `store` is injectable so a test can use its own database.
+ *
+ * Constructed at module scope rather than defaulted in the parameter list,
+ * which would build a new store on every render and reopen the connection
+ * each time.
+ */
+const APP_STORE = new DurableStore();
+
+export function AppShell({
+  client,
+  store = APP_STORE,
+}: {
+  readonly client: BoundaryClient;
+  readonly store?: DurableStore;
+}): ReactNode {
   return (
     <ViewStateProvider>
       <LiveRegionProvider>
-        <Chrome client={client} />
+        <Chrome client={client} store={store} />
       </LiveRegionProvider>
     </ViewStateProvider>
   );
--- a/web/src/shell/AppShell.tsx
+++ b/web/src/shell/AppShell.tsx
@@ -252,9 +252,15 @@ function Chrome({
           `data-saving` reports whether a preference write is still in
           flight. It carries no user-facing claim — SPEC-0009 REQ "Storage
           Is Evictable" governs what may be *said* about stored data, and
-          that indication is #114's to build. This is the hook for it, and
-          in the meantime it is what lets a test wait for a write to settle
-          rather than sleeping and hoping.
+          that indication belongs to the data-custody story rather than
+          here.
+
+          Issue numbers are spelled without the leading hash in this file.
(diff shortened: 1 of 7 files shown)


jonstump:
## Review — approved

Base `main`, no conflict markers, 9/9 CI, +861/-3. Tier 0 with #120.

### The line was not crossed, and I checked rather than took the word for it

`tests/shell/view-state.spec.ts` is **not in this diff**. The five-key shape assertion that has kept graphs out of view state since #82 is untouched, and `useStoredData` only dispatches `setPreference` — the action that already existed.

Declining a bulk `replacePreferences` action is the load-bearing decision here, and the stated reason is the right one: a bulk setter is where a later caller passes something that is not a preference. That is how a store wired for one narrow purpose becomes a general state container, which is the back door the shape rule exists to keep shut.

### `Stored<T>` makes the wrong thing not compile

`?? 0` reads as a careful default and renders a claim — a player who has never looked at their copper and one who looked and found none are different facts, and `place.stocked?.[itemId] ?? 0` collapses them. Expressing it as a discriminated union means the collapse is a type error rather than a code review finding, which is the same move `flag()` makes in the boundary and `ViewState` makes for plans.

**Both directions are tested**, which is what makes it meaningful: a place with a stored `"0"` renders `0`, and a place with no entry renders an em dash. A surface that dashed everything would pass the absence test and be useless — the same failure shape #82's single-status-badge inference had.

Pairing the dash with an `aria-label` because "—" announces as nothing is the detail that would have been missed. A row reading as a name followed by silence is indistinguishable from a broken one.

### Three states rather than two

Loading / ready-and-empty / unavailable, with the collapses named individually. "Collapsing loading into empty flashes 'nothing saved' at a player who has data, which reads as loss rather than latency" is the one a reasonable implementation gets wrong, and it is not the one SPEC-0009 names — the spec names empty-into-unavailable. Covering both is more than the requirement asks and the right more.

### Two disclosures worth the credit

**The dropped promise.** `putPreferences` issued and not awaited, so a reload arriving before the transaction committed lost the change — found by a test that toggles two preferences and reloads at once, which the body notes "is also exactly what a player does when they set something and leave." That is the test being shaped by the failure rather than by the code.

**The flaky wait.** The file records that the first version "passed locally on timing luck and failed in CI, twice, which is exactly the failure mode a flag-based wait has." Recording a fixed flake in the file rather than quietly fixing it is what stops the next person reintroducing it.

### Scope held

`data-saving` carrying no user-facing claim, with the eviction-honesty indication deferred to #114 as SPEC-0009 REQ "Storage Is Evictable" governs it — correct, and the kind of boundary that is easy to blur when the attribute is already there.

Approving and merging.

Issue #115, opened 2026-08-28
Tests: preferences, empty state, deletion, and custody


jonstump:
Part of #110. Covers #113 and #114. Governed by [SPEC-0009](docs/openspec/specs/durable-store/spec.md).

**Blocked by #113 and #114.**

## Test Requirements

- [ ] **A preference survives a real reload**: set, reload the page, assert. Driven through an actual navigation rather than by calling the restore function — a restore that works when invoked directly and never runs on load passes the second and fails the player
- [ ] **Persistence does not widen view state**: `ViewState`'s key set is unchanged, and the existing "no reachable state holds a derived figure" assertion still holds. The store must not become the place the plan ends up
- [ ] **The empty state is reached and is not a failure**: a context with no prior data renders the designed empty state and reports no error
- [ ] **Absent is not zero**: a place with no stocked quantity for an item renders the quantity as absent. The case where the two are easiest to conflate
- [ ] **Deletion removes everything and lands in the empty state**: after confirming, the workspace and every place are gone and the application is in the designed empty state, not an error state
- [ ] **Deletion is confirmed first**: activating the control removes nothing until the confirmation is accepted
- [ ] **The confirmation dialog uses the shell's focus trap**: focus returns to the invoking control on close by **every** route — Escape, backdrop, and the close control, each tested separately
- [ ] **No indication claims a backup**: the rendered text for a stored change is checked against language implying backup, sync, or a guarantee
- [ ] **No share or upload control for an image**: asserted over the rendered control set, since the requirement is an absence
- [ ] **WCAG 2.1 AA audit** over the deletion flow, including the confirmation dialog open

## Acceptance Criteria

- Each test fails against a deliberately broken implementation, checked by breaking it. `scripts/mutation-check.sh` stands at 21 mutations and this story extends it
- **The focus-return routes are separate tests.** A single "close the dialog" test passes while two of the three routes are broken — this was demonstrated on #82, where rewriting the trap as the Escape-only version left the Escape test green and turned the other two red
- The reload test navigates. A test that calls the restore function directly cannot tell a working restore from one that is never wired to load
- The backup-language check asserts on rendered text rather than on source, because the claim a player reads is what the requirement is about
- The image and share-control checks are absences, asserted over the rendered control set rather than by exercising a flow

### Branch

`feat/115-tests-preferences-empty-state-and-custody`

### PR Convention

```
Closes #115
Part of #110

[...]

PR #122, opened 2026-08-29
feat(web): data custody — deletion, eviction honesty, and images

### LINKED ISSUE #114: Data custody: deletion, eviction honesty, and images

jonstump:
Closes #114
Part of #110

Implements SPEC-0009 REQ "Deletion Is a First-Class Operation", REQ "Storage Is Evictable and the Application Must Not Imply Otherwise", REQ "Screenshots Are Local-Only".

> **Stacked on #121.** Base is `feat/113-preferences-persist-and-empty-state`, not `main`. The deletion criterion requires landing in *the designed empty state*, which is #113's deliverable, and both stories touch `AppShell.tsx`. Branching from main would have meant reimplementing that state and a guaranteed conflict. **Merge #121 first**, then this retargets to main cleanly.

## Deletion

Reachable from the page, confirmed before it runs, and it lands in the designed empty state rather than an error state — deleting your data is a thing you chose, not a fault.

The confirmation is the shell's `Popover`, so the focus trap restores focus in the **effect cleanup**. That is why Escape, the backdrop and the close control cannot disagree. A bespoke dialog would have restored inside its Escape handler and stranded the other two, which is the failure the accessibility story was worded around. Both routes are tested separately, not as one "close the dialog" case.

## Eviction honesty

A labelling requirement, because at stage 1 there is no technical mitigation. Browsers evict origin storage under pressure, private windows discard it on close, and until an account exists there is no recovery path. Being accurate about the scope **is** the mitigation.

The check runs over the **rendered page's visible text**, not over source. A component can hold "backed up" in a comment explaining why it must not say that — this repo's own checker file is the extreme case — and scanning source would make the explanation the offence.

Two calibrations, both with negative controls:

- **"Saved" is allowed.** It is the ordinary word for what happened; banning it pushes the copy toward circumlocution, which reads as evasive rather than honest. What is banned is the claim the data is somewhere *else* as well, or that it cannot be lost.
- **Export and download are not sharing.** Both hand the player their own file; neither transmits it. What is refused is uploading and publishing, which ADR-0013 has not been written to authorise.

A companion test asserts the page is *accurate about the scope* rather than merely silent — a page that said nothing would pass the claim check and leave the player unaware their data is one cache clear from gone.

## A test that had no teeth

"Deletion does not immediately recreate the workspace" **passed with the guard removed**. The test never changed a preference away from its default, so the post-deletion reset was a no-op and the write effect had nothing to fire on. It changes one first now, and goes red without the guard.

That guard matters: preferences live on the workspace record, so the reset following deletion can write a fresh workspace back one tick later — silently, after the player has been told their data is gone.

## Verification

**189 tests pass.** lint, lint:css, typecheck, format:check, check:tokens, build all clean.

Four probes, each confirmed red before being accepted:

| Probe | Test that must fail |
|---|---|
| delete without confirming | deletion is confirmed before anything is removed |
| copy claims backup and sync | nothing on the page claims the data is backed up |
| a "Share screenshot" control appears | no control anywhere would share or upload |
| the post-deletion guard removed | deletion does not immediately recreate the workspace |

`scripts/mutation-check.sh` is deliberately untouched, to avoid conflicting with #120 which extends the same file. These four are owed to it once #120 lands.

No protected paths touched.

### DIFF
--- a/web/src/shell/AppShell.tsx
+++ b/web/src/shell/AppShell.tsx
@@ -15,6 +15,7 @@ import { usePlanResolution, type Resolution } from "../state/usePlanResolution";
 import { useStoredData } from "../state/useStoredData";
 import { DurableStore } from "../store";
 import { StatusBadge } from "./StatusBadge";
+import { DataCustody } from "./DataCustody";
 import { StoredPlaces } from "./StoredPlaces";
 import { ViewPreferences } from "./ViewPreferences";
 
@@ -246,6 +247,10 @@ function Chrome({
 
         <section className="panel" aria-label="Saved places">
           <StoredPlaces data={stored} target={state.inputs.target} />
+        </section>
+
+        <section className="panel" aria-label="Your data">
+          <DataCustody data={stored} />
         </section>
 
         {/*
--- a/web/src/shell/DataCustody.tsx
+++ b/web/src/shell/DataCustody.tsx
@@ -1,0 +1,93 @@
+import { useCallback, useId, useState, type ReactNode } from "react";
+
+import { Popover } from "../a11y/Popover";
+import { useLiveRegion } from "../a11y/useLiveRegion";
+import type { StoredData } from "../state/useStoredData";
+
+/*
+ * What the player is told about their data, and what they can do about it.
+ *
+ * Governing: ADR-0008 (durable user data, local-first), SPEC-0009 REQ
+ * "Deletion Is a First-Class Operation", REQ "Storage Is Evictable and the
+ * Application Must Not Imply Otherwise", REQ "Screenshots Are Local-Only"
+ *
+ * ADR-0002 banked "needs no upload endpoint, retention policy, or deletion
+ * story" as a benefit of holding nothing. This spends the third of those
+ * the moment the first byte is written: once data is held, the player is
+ * owed a way to remove it that is not browser developer tools.
+ *
+ * The wording below is load-bearing and is checked mechanically by
+ * tests/helpers/claim-checks.ts. Browsers evict origin storage under
+ * pressure, private windows discard it on close, and until an account
+ * exists there is no recovery path — so "saved" on its own is a stronger
+ * claim than the storage makes, and "backed up" or "synced" would be false.
+ * There is no technical mitigation at stage 1; being accurate about the
+ * scope IS the mitigation, which is why this is a labelling requirement.
+ *
+ * No share or upload control appears here, and that is a requirement rather
+ * than an omission. ADR-0008 excluded blobs on measured grounds — one
+ * capture is 1.5-3 MB against 596 KB for a 200-place text workspace — and
+ * deferred them to ADR-0013. Until that decision exists there is nowhere
+ * for an image to go, so there is no control to send it there.
+ */
+export function DataCustody({ data }: { readonly data: StoredData }): ReactNode {
+  const [confirming, setConfirming] = useState(false);
+  const [deleting, setDeleting] = useState(false);
+  const headingId = useId();
+  const { announce } = useLiveRegion();
+
+  const close = useCallback(() => {
+    setConfirming(false);
+  }, []);
+
+  const confirm = useCallback(() => {
+    setDeleting(true);
+    void data.deleteEverything().finally(() => {
+      setDeleting(false);
+      setConfirming(false);
+      announce("Stored data deleted. Nothing is saved on this device.");
+    });
+  }, [data, announce]);
+
+  return (
+    <>
+      <p className="label">
+        Kept on this device, in this browser. Clearing site data removes it, and a browser
+        short of space can remove it on its own.
+      </p>
+
+      <button
+        type="button"
+        className="control control-sm interactive"
+        onClick={() => {
+          setConfirming(true);
+        }}
+      >
+        Delete stored data
+      </button>
+
+      {/*
+        The shell's Popover, not a bespoke dialog. Its focus trap restores
+        focus in the effect cleanup, so Escape, the backdrop and the close
+        control all converge on the same line — a dialog written here would
+        have to reimplement that, and would get one of the three routes
+        wrong. SPEC-0009's criterion names #82's trap for exactly this.
+      */}
+      <Popover open={confirming} onClose={close} labelledBy={headingId}>
+        <h3 id={headingId}>Delete stored data?</h3>
+        <p className="label">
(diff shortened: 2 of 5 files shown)


jonstump:
## Review — approved, rebased to resolve the stack

Arrived **CONFLICTING / DIRTY**. Rebased before reviewing; details below.

### The rebase

The PR body predicted this correctly — *"Merge #121 first, then this retargets to main cleanly."* Half of that happened: GitHub auto-retargeted the base to `main` when #121's branch was deleted on merge, but retargeting changes the base without rebasing the commits. So the branch still carried three:

```
d2eda5b  data custody — deletion, eviction honesty, and images   ← the only new one
85c4fd4  a store-content wait, and an issue number as a colour   ← already on main
588d239  preferences persist, and an empty store is a designed state  ← already on main (#121)
```

I checked both earlier commits against `main` rather than assuming: `git diff <commit> origin/main -- <their files>` is empty for each, so #121 delivered both. A plain `git rebase origin/main` conflicted on `AppShell.tsx` and `stored-preferences.spec.ts` because it was replaying work that was already there.

`git rebase --onto origin/main 85c4fd4` drops the duplicates and replays only `d2eda5b`. The diff went from +1482/-3 to **+625/-4**, five files, and the state is now MERGEABLE / CLEAN.

Safe to do because the worktree was clean and exactly at the remote — no in-flight work to clobber. I verified that first.

**Verified the rebased tree rather than trusting a clean rebase**, since this session has already seen one that reported clean and broke six tests:

```
typecheck PASS   lint PASS   lint:css PASS   format:check PASS
check:tokens PASS   build PASS   198 tests passed   all mutations red
```

198 rather than the body's 189, because the branch now sits on a main that includes #120 and #123.

### The toothless test is the best thing here

*"Deletion does not immediately recreate the workspace"* passed with the guard removed. The comment says exactly why and does not soften it:

> Without it the post-deletion reset is a no-op, the write effect sees no difference, and the test passes against an implementation that has no protection at all — which is exactly what it did until removing the guard failed to turn it red.

The test now unchecks a preference and waits for it to persist *before* deleting, so the reset has something to write back. Finding this required running the probe rather than reasoning about the test, and the failure it protects against is the worst kind: a fresh workspace written one tick after the player was told their data was gone.

### The claim check operates on rendered text, and the reason is right

> A component can hold the string "backed up" in a comment explaining why it must not say that — this file's own regex is the extreme case — and scanning source would make the explanation the offence.

Verified: pure functions over text, run against the rendered page's visible content. A source scan here would punish the documentation, and the file that defines the banned phrases would be its own first violation.

Both calibrations are right and both carry negative controls. "Saved" allowed because banning it pushes copy toward circumlocution, which reads as evasive rather than honest — that is a real observation about how honesty-by-labelling actually fails. Export and download not counted as sharing because both hand the player their own file.

**And the companion test is what makes the pair work.** `the page is accurate about the scope, rather than merely silent` asserts the custody region contains *"this device"* and *"remove"* — so a page that said nothing would fail. Without it, the strongest claim-check in the world passes against a blank page.

### One piece of debt now due

> `scripts/mutation-check.sh` is deliberately untouched, to avoid conflicting with #120 which extends the same file. These four are owed to it once #120 lands.

Correct call at the time, and **#120 has since landed** (mutation count 21 → 24). The four probes listed in the body — delete without confirming, copy claims backup, a share control appears, the post-deletion guard removed — were each verified red by hand, which proves an afternoon. Committing them makes it permanent, and the last one especially: it is the probe that already caught a toothless test once.

Not blocking. Worth doing in #115, which is the remaining story on this epic and is a tests story.

Approving and merging.

Issue #84, opened 2026-08-28
Implement Tree Canvas


jonstump:
Governed by [SPEC-0006](docs/openspec/specs/tree-canvas/spec.md) ([design](docs/openspec/specs/tree-canvas/design.md)).

The dependency-tree surface: the player picks a target and quantity and sees the full crafting and refining tree as a left-to-right flowchart, changes the method or recipe per node, and assigns leaf resources to bases.

This is the first surface built on SPEC-0005. Everything SPEC-0005 requires of a view surface is **inherited and not restated** — tokens, component styling, the no-arithmetic rule, the single boundary client, view-state boundaries, and the accessibility baseline. Those primitives are on `main` as of #82.

## Stories

| Issue | Story | Requirements | Depends on |
|---|---|---|---|
| #85 | Canvas rendering, layout, and edges `foundation` | Graph Rendering From the Boundary Payload; Layout Geometry Is Not a Domain Value; Edge Rendering | — |
| #86 | Node card: identity, yield, and provenance | Node Card; Yield and Application Display; Provenance Display | #85 |
| #87 | Method and recipe selection | Method Selection; Recipe Selection | #86 |
| #88 | Leaf assignment to bases | Leaf Assignment to Bases | #85 |
| #89 | Tests: rendering, ordering, and layout independence | covers #85 | #85 |
| #90 | Tests: node card, controls, and provenance | covers #86 and #87 | #87 |

All nine of the spec's requirements appear in exactly one story.

## What enters the project here

design.md names two substantial third-party dependencies arriving with this spec: **React Flow** for the graph surface and **elkjs** for layout. Both sit directly under a spec that forbids the view computing domain values, which is why the layout carve-out is stated normatively rather than left to review-time argument — the layout engine reads nodes and edges, never a total, and "changing quantity must not move the graph" is the observable form of that rule.

## The one thing that is not ready

#88 (leaf assignment) needs stage 2. The Go side is complete — `internal/bridge/module.go` implements `Rollup` and `Power` in full — but `web/src/boundary/client.ts` exposes only `resolve`. #88 extends the client with `rollup`; it is scoped there rather than into a separate foundation story because no other SPEC-0006 story needs it.

**SPEC-0007 will extend the same file for `power`.** Whoever plans SPEC-0007 should expect that, and may prefer to pull both into one boundary-client story at that point.

## No security checklist

[...]

PR #124, opened 2026-08-29
feat(web): canvas rendering, layout, and edges

### LINKED ISSUE #85: Canvas rendering, layout, and edges

The render pipeline: one boundary payload in, positioned nodes and drawn edges out. Everything else in this epic mounts into what this story builds, which is why it is the foundation.

React Flow and elkjs enter the project here — design.md names both, and both sit directly under a spec that forbids the view computing domain values. The carve-out is normative: the layout engine reads which nodes exist and which edges connect them, and never a quantity. Passing a total to it would be the violation, and "changing quantity must not move the graph" is what makes the rule observable rather than a matter of reading layout code.

Extends `web/src/boundary/` (the client is on main and already returns `legalMethods`, `legalRecipes`, `yield`, `applications`, `terminal`, `verified` and per-edge `perUnit`/`yield` on every node). Extends `web/src/shell/AppShell.tsx`, which currently renders a flat figure list where the canvas belongs.

## Requirements

- [ ] **Graph Rendering From the Boundary Payload** (SPEC-0006) — The canvas MUST render nodes and edges from a single `resolve` payload obtained through the SPEC-0005 boundary client. It MUST NOT assemble the graph from repeated boundary calls, and MUST NOT read the Tier 1 artifact. It MUST take the payload's node order as given and MUST NOT sort, rank, or otherwise reorder nodes. Each node's edges MUST be drawn from that node's own `children`, and the canvas MUST NOT infer an edge the payload does not contain.
  - WHEN the canvas renders a resolved plan of 36 nodes THEN exactly one `resolve` crossing produced it, and every node and edge was read from that single returned value
  - WHEN the canvas holds a resolved payload THEN the node sequence it iterates matches the payload's sequence exactly, and no comparison function is applied to it

- [ ] **Layout Geometry Is Not a Domain Value** (SPEC-0006) — Node positions and edge paths MUST be computed by the layout engine from graph structure alone. Position MUST NOT be derived from any quantity, total, yield, application count, or producer figure. A layout that scaled a node by its total, or ordered a column by quantity, is prohibited.
  - WHEN the layout engine is invoked for a resolved graph THEN its input is nodes and edges, and no node total, yield or application count is passed to it
  - WHEN the target quantity changes and every total scales THEN node positions are unchanged, because no position was derived from a total

- [ ] **Edge Rendering** (SPEC-0006) — An edge MUST carry the per-unit quantity relating its two nodes, taken from the payload. The method of the node an edge feeds MUST be readable from the edge itself as well as from the node's badge. Edge styling MUST be decorative reinforcement only: every fact an edge's appearance conveys MUST also be available as text on the node it connects to.
  - WHEN a node's method is refine THEN the edges feeding it are visually distinguished from those feeding a crafted node
  - WHEN edge styling is disregarded entirely THEN every fact it conveyed remains available as text on the connected nodes

## Acceptance Criteria

- One `resolve` call produces the whole tree, verified by counting crossings rather than by inspection
- The rendered node sequence is the payload's sequence, and no sort or comparator exists in the canvas source
- The layout engine's call site receives nodes and edges only — checkable mechanically, and the reason the carve-out was written as a line rather than a principle
- Changing target quantity leaves every node position byte-identical
- Tab order follows the payload's order, not the layout's, per design.md — a deliberate divergence from "tab order follows visual layout" and worth not later "fixing"
- Every edge shows its per-unit quantity, and disregarding edge styling entirely loses no fact

## Notes

design.md sequences this first for a reason: "Static rendering: payload → layout → node cards and edges, with no interaction. This exercises the ordering, the token discipline, and the no-arithmetic rule before any control exists."

Edge label legibility is flagged rather than settled. The handoff records labels as legible at the prototype's 34 nodes and flags 60+ as needing hover-only; the real Stasis Device target already resolves to 36. The spec requires the per-unit quantity be present without fixing how it is revealed, so hover-only remains available without a spec change.

### Branch

`feat/85-canvas-rendering-layout-and-edges`

### PR Convention

```
Closes #85
Part of #84

Implements SPEC-0006 REQ "Graph Rendering From the Boundary Payload", REQ "Layout Geometry Is Not a Domain Value", REQ "Edge Rendering".
```

jonstump:
Closes #85
Part of #84

Implements SPEC-0006 REQ "Graph Rendering From the Boundary Payload", REQ "Layout Geometry Is Not a Domain Value", REQ "Edge Rendering".

React Flow and elkjs enter the project here, as design.md sequences them.

## The layout carve-out

SPEC-0006 carves layout out of SPEC-0005's no-arithmetic rule and draws the line at what the engine may read: structure yes, quantities no. design.md is explicit that the line exists so the argument doesn't have to be had at review time on every surface — which only works if something watches it.

It's kept by the types rather than by discipline. `LayoutNode` has one string field and it's an id; there's no field a `Quantity` could occupy without adding one, which is a diff a reviewer sees — the same argument `ViewState` is built on.

`toLayoutInput` is a named function rather than an inline `.map` at the call site, so the line is something a test can hold. Checked three ways:

1. **Exact key set** of the object that reaches elk — `{id, width, height}` per node, `{id, sources, targets}` per edge. The engine is a third party and takes whatever it's given; the types keep a quantity out of `LayoutNode`, this keeps one out of what actually crosses.
2. **Deep scan** of the layout input against a fixture carrying distinctive quantities. This is what would catch a field added later — a new key holding a total satisfies `toEqual` only if someone updates the expectation, and they would.
3. **Byte-identical positions** across quantity 1 → 7 → 9999 on the real 36-node tree. Compared as strings, not approximately: a layout deriving x from a total moves sub-pixel for a small change and passes a tolerance.

Test 3 has a companion asserting the totals *did* change — without it, a silently failed recompute makes every position identical for the wrong reason.

## Ordering

The node sequence is the payload's, and there is no comparator in `src/canvas`. Both are tested, and both are needed: **a payload that arrived already sorted would satisfy every ordering assertion against a canvas that sorted it again.** So one test compares rendered order to the captured payload, and another scans the source for `.sort(`, `.reverse(`, `localeCompare`.

One `resolve` crossing renders all 36 nodes, counted at `globalThis.nmsPlanner` rather than at the client — a wrapper on the client would miss a second client, and constructing one is the shape the requirement forbids.

Tab order follows the payload, asserted by walking focus rather than inferred from DOM order, since design.md calls this a deliberate divergence from "tab order follows visual layout" and worth not later "fixing".

## Two measured findings, both fixed

**Bundle.** Statically imported, these two dependencies took the initial bundle from **222 kB to 1,864 kB raw — 69 kB to 575 kB gzipped**, paid on first paint by a player who has not resolved anything. That directly undercuts SPEC-0005's reason for lazy-loading the WASM binary, so it gets the same answer:

| chunk | raw | gzip | fetched |
|---|---|---|---|
| initial | 224 kB | 70 kB | always (main: 222 / 69) |
| TreeCanvas | 183 kB | 60 kB | on first resolve |
| elk.bundled | 1441 kB | 439 kB | on first layout |

elkjs splits again inside `layout.ts` rather than riding along in the canvas chunk — it is 1.6 MB on its own, and a canvas rendering a cached layout shouldn't fetch the engine that produced it.

**Contrast.** React Flow's attribution link ships at a computed ratio of **1.12** against this surface — an outright WCAG AA failure, caught by the axe audit already running on the shell. It's **restyled, not hidden.** React Flow is MIT and `hideAttribution` exists, but removing the credit is what xyflow asks people to buy a subscription for, and doing it to fix a contrast score would be a licensing decision wearing an accessibility finding as cover. The link stays and is now legible.

This is the one place the stylesheet names a third-party class; React Flow exposes no prop for that element, so the alternative was shipping the failure.

## Scope note: the canvas sits beside the flat figure list

SPEC-0006 puts the canvas where the flat list is, and that's where it will end up. But `.figure-row` is currently what the SPEC-0005 accessibility suite drives — focus trap, selection ring, and the live region's invoker all run through it — and the canvas card has no control to open until #87 lands. Removing the list now would leave those requirements untested in between. **The list should retire with #87**, not before.

## Worth flagging

`scripts/check-tokens.sh` scans comments as well as declarations, so a colour literal quoted in an explanatory comment fails the gate — while `tests/helpers/css-checks.ts` strips comments and has an explicit test that *"a literal inside a comment is not a finding"*. Two enforcers of the same rule disagree. It cost me two rewrites here (a `#nnn` issue reference in one PR, React Flow's grey in this one). Not fixed here — it's a shared script and out of scope — but worth an issue.

## Verification

**220 tests pass** (up from 202), 18 new. lint, lint:css, typecheck, format:check, check:tokens, build all clean. **30/30 mutations red**, six of them new:

| Mutation | Test that goes red |
|---|---|
| the canvas sorts the nodes it was given | rendering |
| a total reaches the layout engine | layout |
| node width is derived from the total | layout |
| an edge stops showing its per-unit quantity | edges |
| edge styling stops distinguishing a refine step | edges |
| the node card stops naming its method | edges |

The last pair are the two halves of *"disregarding edge styling loses no fact"* — one breaks the styling, one breaks the text it falls back to.

No protected paths touched.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -8,6 +8,8 @@
       "name": "nms-base-planner-web",
       "version": "0.0.0",
       "dependencies": {
+        "@xyflow/react": "^12.11.5",
+        "elkjs": "^0.12.0",
         "react": "^19.2.0",
         "react-dom": "^19.2.0"
       },
@@ -1801,6 +1803,55 @@
         "@babel/types": "^7.28.2"
       }
     },
+    "node_modules/@types/d3-color": {
+      "version": "3.1.3",
+      "resolved": "https://registry.npmjs.org/@types/d3-color/-/d3-color-3.1.3.tgz",
+      "integrity": "sha512-iO90scth9WAbmgv7ogoq57O9YpKmFBbmoEoCHDB2xMBY0+/KVrqAaCDyCE16dUspeOvIxFFRI+0sEtqDqy2b4A==",
+      "license": "MIT"
+    },
+    "node_modules/@types/d3-drag": {
+      "version": "3.0.7",
+      "resolved": "https://registry.npmjs.org/@types/d3-drag/-/d3-drag-3.0.7.tgz",
+      "integrity": "sha512-HE3jVKlzU9AaMazNufooRJ5ZpWmLIoc90A37WU2JMmeq28w1FQqCZswHZ3xR+SuxYftzHq6WU6KJHvqxKzTxxQ==",
+      "license": "MIT",
+      "dependencies": {
+        "@types/d3-selection": "*"
+      }
+    },
+    "node_modules/@types/d3-interpolate": {
+      "version": "3.0.4",
+      "resolved": "https://registry.npmjs.org/@types/d3-interpolate/-/d3-interpolate-3.0.4.tgz",
+      "integrity": "sha512-mgLPETlrpVV1YRJIglr4Ez47g7Yxjl1lj7YKsiMCb27VJH9W8NVM6Bb9d8kkpG/uAQS5AmbA48q2IAolKKo1MA==",
+      "license": "MIT",
+      "dependencies": {
+        "@types/d3-color": "*"
+      }
+    },
+    "node_modules/@types/d3-selection": {
+      "version": "3.0.11",
+      "resolved": "https://registry.npmjs.org/@types/d3-selection/-/d3-selection-3.0.11.tgz",
+      "integrity": "sha512-bhAXu23DJWsrI45xafYpkQ4NtcKMwWnAC/vKrd2l+nxMFuvOT3XMYTIj2opv8vq8AO5Yh7Qac/nSeP/3zjTK0w==",
+      "license": "MIT"
+    },
+    "node_modules/@types/d3-transition": {
+      "version": "3.0.9",
+      "resolved": "https://registry.npmjs.org/@types/d3-transition/-/d3-transition-3.0.9.tgz",
+      "integrity": "sha512-uZS5shfxzO3rGlu0cC3bjmMFKsXv+SmZZcgp0KD22ts4uGXp5EVYGzu/0YdwZeKmddhcAccYtREJKkPfXkZuCg==",
+      "license": "MIT",
+      "dependencies": {
+        "@types/d3-selection": "*"
+      }
+    },
+    "node_modules/@types/d3-zoom": {
+      "version": "3.0.8",
+      "resolved": "https://registry.npmjs.org/@types/d3-zoom/-/d3-zoom-3.0.8.tgz",
+      "integrity": "sha512-iqMC4/YlFCSlO8+2Ii1GGGliCAY4XdeG748w5vQUbevlbDu0zSjH/+jojorQVBK/se0j6DUFNPBGSqD3YWYnDw==",
+      "license": "MIT",
+      "dependencies": {
+        "@types/d3-interpolate": "*",
+        "@types/d3-selection": "*"
+      }
+    },
     "node_modules/@types/estree": {
       "version": "1.0.9",
       "resolved": "https://registry.npmjs.org/@types/estree/-/estree-1.0.9.tgz",
@@ -1839,7 +1890,7 @@
       "version": "19.2.18",
       "resolved": "https://registry.npmjs.org/@types/react/-/react-19.2.18.tgz",
       "integrity": "sha512-AnzbBERsrLKtk2XSfTbYRLjQPdy116Sty4q+T+Bp3IC4l6jNBvreVPAHmpq9qhXQM7CXZPjLVmGMw9sy+hxQ3w==",
-      "dev": true,
+      "devOptional": true,
       "license": "MIT",
       "dependencies": {
         "csstype": "^3.2.2"
@@ -1849,7 +1900,7 @@
       "version": "19.2.4",
       "resolved": "https://registry.npmjs.org/@types/react-dom/-/react-dom-19.2.4.tgz",
       "integrity": "sha512-Bsc+QHgp+P/F02XDzNCY9jnZNCUuLki36KT7VKrTXXLdHf+vHMNZnW1rVu5DNW/rCK+fya3DATySbLM4yhtKUw==",
-      "dev": true,
+      "devOptional": true,
       "license": "MIT",
       "peerDependencies": {
         "@types/react": "^19.2.0"
@@ -2171,6 +2222,48 @@
         "vite": "^4.2.0 || ^5.0.0 || ^6.0.0 || ^7.0.0 || ^8.0.0"
       }
     },
+    "node_modules/@xyflow/react": {
+      "version": "12.11.5",
+      "resolved": "https://registry.npmjs.org/@xyflow/react/-/react-12.11.5.tgz",
+      "integrity": "sha512-QqoryGkqEhWBuQN9bZWRKhwr3Uoj9lCGj/tg0NnIWHHEOXV+5c8cYsc7Q9TST63V92lHqcNV9P5cjvJ9ZAmblQ==",
+      "license": "MIT",
+      "dependencies": {
+        "@xyflow/system": "0.0.81",
+        "classcat": "^5.0.3",
+        "zustand": "^4.4.0"
+      },
+      "peerDependencies": {
+        "@types/react": ">=17",
+        "@types/react-dom": ">=17",
+        "react": ">=17",
+        "react-dom": ">=17"
+      },
+      "peerDependenciesMeta": {
+        "@types/react": {
+          "optional": true
+        },
+        "@types/react-dom": {
+          "optional": true
+        }
+      }
+    },
+    "node_modules/@xyflow/system": {
+      "version": "0.0.81",
+      "resolved": "https://registry.npmjs.org/@xyflow/system/-/system-0.0.81.tgz",
+      "integrity": "sha512-hfbafW4i7uLq7ILok8QWFFm4KMFw22lbZNJHKfHOMSOOoCk5e5m8yfr84UV9NaJajmogWaLVnp2XFU9JQejlqg==",
+      "license": "MIT",
+      "dependencies": {
+        "@types/d3-drag": "^3.0.7",
+        "@types/d3-interpolate": "^3.0.4",
+        "@types/d3-selection": "^3.0.10",
+        "@types/d3-transition": "^3.0.8",
+        "@types/d3-zoom": "^3.0.8",
+        "d3-drag": "^3.0.0",
+        "d3-interpolate": "^3.0.1",
+        "d3-selection": "^3.0.0",
+        "d3-zoom": "^3.0.0"
+      }
+    },
     "node_modules/acorn": {
       "version": "8.18.0",
       "resolved": "https://registry.npmjs.org/acorn/-/acorn-8.18.0.tgz",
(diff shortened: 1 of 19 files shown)


jonstump:
## Review — changes requested

Conflict-marker gate: clean. CI: 9/9 green. Governing spec loaded (SPEC-0006 REQ "Graph Rendering From the Boundary Payload", REQ "Layout Geometry Is Not a Domain Value", REQ "Edge Rendering", Accessibility Requirements) and issue #85's acceptance criteria checked against the diff. qmd retrieval over `nms-base-planner-adrs` surfaces ADR-0004 and ADR-0003; ADR-0004 is already cited in every new file, ADR-0003 governs the domain core and is not this PR's. No missing references.

The three governing requirements are met, and the way the layout carve-out is held — a named `toLayoutInput`, a type with no field a `Quantity` could occupy, and the byte-identical-positions test with its did-the-totals-actually-change companion — is the right construction. The reasoning about why test 3 needs that companion is exactly right.

Three findings, all verified by running them rather than by reading.

---

### 1. The axe audit does not cover the canvas, so the contrast fix is unguarded

The PR body says React Flow's attribution failure was **"caught by the axe audit already running on the shell."** It was not, and it cannot be.

`the shell passes a WCAG 2.1 AA audit with figures on screen` calls `resolveAPlan`, which waits for `getByRole("heading", { level: 3 })` — the flat figure list. The canvas is behind `lazy()` *and* waits on an elk layout, so it has not mounted when axe analyses.

Probe: disabled the two `.react-flow__attribution` rules in `canvas.css` and re-ran the audit suite.

```
✓ the shell passes a WCAG 2.1 AA audit before any figures exist
✓ the shell passes a WCAG 2.1 AA audit with figures on screen
2 passed
```

Green with the fix removed. A second probe, identical except that it also waits for the canvas to paint:

```
attribution present when the axe test analyses = 0
attribution present after the canvas paints     = 1
violations = ["color-contrast: a"]
```

So the finding is real, the fix is correct, and nothing holds it. The larger consequence is the one to fix: **the entire new surface is unaudited** — node cards, edge labels, the `.numeral` totals, the method badges. This PR adds a surface and the WCAG suite still only sees the one it replaces.

The fix is one wait. `resolveAPlan` should wait for the canvas's first `.node-card` the way `tests/canvas/rendering.spec.ts` already does, so the populated-state audit covers the populated state.

This is the same shape as the deletion test in #122 — a guard that passes because the thing it guards is not on screen when it looks.

### 2. The canvas ignores the group-digits preference

`TreeCanvas.tsx:80` hardcodes the separator:

```ts
display: formatQuantity(node.total, { groupSeparator: "," }),
```

Every other figure on the surface threads the preference — `AppShell.tsx:194` for the list rows, `:76` and `:234` for the live region. And because of this PR's own scope decision, the canvas sits **beside** the flat list, so both are on screen at once.

Probe, at `ULTRAPROD2` × 500, unchecking "Group digits":

```
list grouped before=20  after=0     sample: "30000"
card grouped before=20  after=20    sample: "30,000"
```

The same twenty figures, on one screen, at one moment, formatted two ways — after the player asked for one of them. That preference is the one #121 made persistent and #122 wrote a test around; SPEC-0009 REQ "View Preferences Survive a Reload" is satisfied by the store, and then undone at the point of display.

`useViewState()` is a hook and `Figures` already calls it (`AppShell.tsx:154`), so the fix is the same two lines here.

### 3. An elk failure renders every node piled at the origin

`layoutGraph` returns an empty `Map` when the engine throws or fails to load. `TreeCanvas` guards on `!placements`, which an empty Map passes, then falls back per node to `{ x: 0, y: 0 }`.

Probe, aborting the elk chunk request:

```
6 nodes, distinct positions = 1 of 6
```

Every node stacked at the origin — which is precisely what the `!placements` comment says the pending state exists to prevent:

> Pending, never zero, and never a graph drawn at the origin. Every node placed at (0, 0) would render as one pile that resolves into a tree a frame later, which reads as a rendering fault.

And `layout.ts` says the empty return is safe because **"the caller reports that as its own state"** — the caller does not. The two comments describe a contract neither side implements.

It is not hypothetical for this application: the engine is a lazily-fetched 1.6 MB chunk in an offline-capable, local-first app. Losing the network between first paint and first resolve, or a stale chunk reference after a deploy, produces exactly this.

Distinguishing "not yet" from "failed" is enough — `null` for pending, a failure state for failure, and `StatusBadge` already has the vocabulary. The lazy-load decision is right; it just needs the failure path the comment already promises.

---

### Not blocking

`TreeEdgeData` declares `sourceName` and `targetName`. Neither is supplied by `TreeCanvas`, and neither is read anywhere — `grep` finds only the two declarations. `data as unknown as TreeEdgeData` is what lets it compile: the double cast means the interface is documentation of a shape that does not exist rather than a check on the one that does. Drop both fields, or supply them.

### Agreed and worth its own issue

The `check-tokens.sh` / `css-checks.ts` disagreement about comments is real and has now cost two rewrites across two PRs. Correctly left out of scope. Happy to open it.

jonstump:
## All three fixed — `f6dc096`

Each fix is paired with a test that goes red without it, and a mutation in `scripts/mutation-check.sh` so it stays that way. **33 mutations, all red. 224 tests pass** (was 220).

### 1. The audit now covers the canvas

The wait went into `resolveAPlan` rather than into the audit, so every test built on the populated state gets the whole of it:

```ts
await expect(
  page.getByRole("region", { name: "Dependency tree" }).locator(".node-card").first(),
).toBeVisible({ timeout: 30_000 });
```

Mutation: *"React Flow's attribution goes back to shipping unreadable"* — unmatches the `.react-flow__attribution` selector; `tests/shell/shell.spec.ts` goes red. That is the assertion that was silently green before.

### 2. The separator is the player's

`useViewState()`, the same hook `Figures` uses. New test asserts both lists start grouped, both stop when "Group digits" is unchecked, and — so it cannot pass by the canvas showing something else — that every value in the list also appears on a card.

Mutation: *"the canvas hardcodes its digit separator again"*.

### 3. A failed layout is now a state, not an empty map

`layoutGraph` returns `null` on failure. A short layout counts as failed too: a placement map missing some nodes would put the rest on the fallback coordinate, which is the same pile in miniature.

`tests/canvas/degraded.spec.ts` aborts the elk chunk and asserts the pile does *not* appear — `.react-flow__node` count is 0, not 36 at one coordinate — plus that the figures are untouched, that the message is announced (`role="status"`), and a **negative control** with the engine available: every node at a distinct position. Without that last one the whole file passes against a canvas that never draws.

The message renders inside the labelled region rather than replacing it. Pending is transient; this is not, and a region that disappears leaves someone navigating by region with no tree and no account of why. It is body text, not a `StatusBadge` — every word in that vocabulary ("Deficit", "Unassigned", "Unverified") is a fact about a plan, and an engine that would not load is not one.

### One thing I changed that you did not ask about

The three states are a discriminated union carrying the `CanvasModel` they were computed from, and the current state is derived at render:

```ts
const layoutOf: LayoutState =
  layout.status === "laying-out" || layout.model === model ? layout : LAYING_OUT;
```

The obvious version resets to pending inside the effect, which `react-hooks/set-state-in-effect` rejects — and the rule is right here. Without the model check, a payload arriving after a settled layout would place its nodes at the previous graph's coordinates for a frame, which is the same pile for every id the two do not share. Deriving it is both the lint-clean shape and the correct one.

### Still open

`TreeEdgeData.sourceName` and `targetName` are still declared, still unsupplied, still unread — left alone as it was outside the three findings. One line each to delete.

jonstump:
## Both follow-ups done — `922b0c5`

**`TreeEdgeData` drops `sourceName` and `targetName`.** Neither was supplied nor read; the `as unknown as` cast is what let them compile.

**The comment disagreement is settled, in favour of the two that strip.** stylelint ignores comments like any CSS parser, `css-checks.ts` strips them and has a test saying so — `check-tokens.sh` was the odd one out, which is what made the rule a matter of which enforcer you hit first.

Block comments cover CSS and both TS forms. Line comments only where `//` opens the line: a `//` mid-line is as likely to be inside a URL, and blanking the rest of that line would hide a literal sitting after it.

**That change is why the second half was necessary.** `css-checks.ts` already had this in its header:

> the shell script has never been observed to reject anything, so nothing distinguishes "the stylesheet is clean" from "the pattern no longer matches"

Comment-stripping widens exactly that failure — over-strip and the gate reports a clean tree forever, silently. So it now proves it can see before it reports on anything: twelve fixtures through the same function on every invocation — five literals it must catch, four comments it must not, a token reference, and two guards against over-stripping (a literal on a line holding a URL, and one after a comment closes). A zero-files-scanned run fails too.

Verified by hand rather than by reading it:

```
a literal in a declaration   → exit 1, reported at canvas.css:163
the same literal in a comment → exit 0
stylelint on the same comment → pass
```

`css-checks.ts`'s header updated, since its claim about the script is no longer true.

**224 tests pass. 33 mutations, all red. All gates clean.**

Issue #99, opened 2026-08-28
Duration, provenance, and absent data


jonstump:
Part of #94. Governed by [SPEC-0007](docs/openspec/specs/base-planner-card/spec.md).

**Blocked by #96.**

Three rules about how figures are presented and what is not presented at all. They are grouped because each applies across the whole card rather than to one section, and implementing them per-section is how one section ends up disagreeing with another.

## The marker will be on everything

REQ "Provenance on Displayed Figures" says it outright, and it is the hardest constraint in this epic:

> Several of this card's figures rest on constants that are planner policy or community-sourced rather than read from the game tables — biodome capacity, fauna yield and cycle length, steps per processor, the depot threshold, panels per battery, processing time — and none of them carries a verified date today. A card rendered against the current constant set marks **every producer row**.

The design tuned this treatment against two subtle chips in a prototype. A card that is legible with two markers and unreadable with thirty has not met the requirement.

## Requirements

- [ ] **Duration Display** (SPEC-0007) — Where the card shows a duration — fill time, growth time, collection cycle, processing step, or readiness estimate — it MUST present it as an estimate rather than a precise figure. The domain's rate constants do not state their time unit; the artifact does not record it, and the engine's arithmetic is consistent under either reading while its absolute durations are not. The card MUST NOT convert a duration into a different unit by arithmetic of its own.
  - WHEN a card shows an extractor fill time THEN it is presented as an estimate, not as an exact figure
  - WHEN a duration crosses the boundary in one unit and is displayed in another THEN the converted value came from the domain, not from arithmetic in the card

- [ ] **Provenance on Displayed Figures** (SPEC-0007) — Where a figure is reported as not verified, the card MUST mark it using the same treatment as elsewhere in the view, without styling it as an error. Every producer row and each base carries its own provenance and the card MUST surface both; it MUST NOT substitute one for the other. The card MUST NOT compute provenance — re-deriving a row's flag in the view would be the arithmetic SPEC-0005 forbids applied to a boolean.
  - WHEN the payload reports a base as verified and one of its rows as not THEN the card marks that row, and does not mark the base's own figures as suspect
  - WHEN the payload reports a base's power budget as not verified THEN the card marks it, and does not style it as an error condition
  - WHEN no curated constant carries a verified date, so every producer row is unverified THEN the card remains readable, and the marker does not depend on being rare to stay unobtrusive

[...]

PR #125, opened 2026-08-29
feat(web): card composition, identity, and producer sections

### LINKED ISSUE #96: Card composition, identity, and producer sections

jonstump:
Closes #96
Part of #94

Implements SPEC-0007 REQ "Card Composition From the Build Payload", REQ "Base Identity and Selection", REQ "Producer Sections", REQ "Byproducts Are Shown, Not Omitted".

## What this adds

`src/card/BasePlannerCard.tsx` renders one base's construction instructions from the stage 2 payload #95 made reachable, plus `src/card/sections.ts`, `src/styles/card.css`, a fixture, and two test files. 835 lines, no existing file modified.

## The four requirements, and the case that catches each

**Every count is the domain's.** A farm row carries both `required` and `yieldPerPlant`, so `required / yieldPerPlant.min` looks like the plant count and is not — SPEC-0007 requires the domain's `plants` *"even where both are present in the payload"*, because the engine sizes on the pessimistic bound. `tests/card/discipline.spec.ts` asserts that absence mechanically, the way the acceptance criteria asked for. It **reuses `numericConversions` from `tests/helpers/source-checks.ts`** rather than writing a second checker: a rule enforced by two regexes drifts the moment one is improved. It carries a negative control, so a checker that stops matching fails there rather than passing forever.

**Base-level counts stay base-level.** `nutrientProcessors` and `pelletFeeders` render once, on the section that owns them, never inside a row. The fixture gives one base **three kitchen steps** because that is the only case that catches a per-row sum — with one step, a summed figure and a base-level figure agree.

**A group absent from the payload is absent, not empty.** The fixture carries a second base with extractors only. An implementation rendering every group unconditionally passes the all-four test and fails this one, which is the entire reason that base exists.

**Byproducts are shown.** `noBuild` rows render with demand and what covers them, marked "Nothing to build" — a word, not a colour, since the distinction has to survive not seeing it.

## Selection and identity

The selection control is a real `<button aria-pressed>`. A `div` with `tabIndex` fails the criterion even when it behaves correctly. The 3px identity frame and the inboard overlay ring are SPEC-0005's and already on main — this uses them and rebuilds neither, and a test asserts the frame survives selection.

Base identity metadata still has no source, which SPEC-0007's overview names. The card takes an identity slot as a prop rather than inventing one; absent a slot it says so with a dashed frame **and** a badge, since colour may not carry a distinction alone.

## Scope note: sibling PR #124

#124 (issue #85, tree canvas) has open claims on `App.tsx`, `AppShell.tsx`, `package.json` and `mutation-check.sh`. Rather than contend for them, the card mounts through its own fixture the way `StatusBadge` does, and touches none of the four. Wiring it into the shell belongs with the panel that arranges cards, which SPEC-0007 explicitly puts out of scope and which has no spec yet. This PR and #124 should merge in either order without a rebase.

Provenance display is left to #99, which owns it.

## Validation

Run against `e04a04a`:

- `npm test` — **214 passed**, including the 13 new card tests and the `card.css` token/inset-box-shadow checks that `stylesheet.spec.ts` picks up by directory scan without being told the file exists

One caveat on the local run: this container ships Chromium build 1194 and `@playwright/test` 1.57 expects 1234, so the browser path was bridged in the container to let the repo's own config run unmodified. No repo file encodes that — CI installs its own browser normally.

---

### DIFF
--- a/web/src/card/BasePlannerCard.tsx
+++ b/web/src/card/BasePlannerCard.tsx
@@ -1,0 +1,242 @@
+import type { ReactNode } from "react";
+
+import {
+  formatQuantity,
+  type BaseBuild,
+  type ExtractorRow,
+  type FarmRow,
+  type KitchenStep,
+  type NoBuildRow,
+  type Quantity,
+  type RanchRow,
+} from "../boundary";
+import { StatusBadge } from "../shell/StatusBadge";
+
+import { PRODUCER_HEADING, presentKinds, type ProducerKind } from "./sections";
+
+/*
+ * One base's construction instructions.
+ *
+ * Governing: ADR-0004 (React view layer), SPEC-0007 REQ "Card Composition
+ * From the Build Payload", REQ "Base Identity and Selection", REQ "Producer
+ * Sections", REQ "Byproducts Are Shown, Not Omitted", SPEC-0005 REQ "The
+ * View Computes No Domain Values", REQ "Component Styling Discipline"
+ *
+ * Every number rendered here arrives as a Quantity and leaves through
+ * formatQuantity, which only groups digits. There is no division, no
+ * rounding and no Number() in this file, and tests/card/discipline.spec.ts
+ * asserts that mechanically rather than trusting this comment — SPEC-0007
+ * forbids computing a count from a quantity and a rate *even where both are
+ * in the payload*, which is precisely the tempting case: a farm row carries
+ * both `required` and `yieldPerPlant`, and `plants` is already the answer.
+ *
+ * `nutrientProcessors` and `pelletFeeders` are rendered once, on the section
+ * that owns them, and never inside a row. They are base-level figures: a
+ * base with three kitchen steps has one processor count, and summing a
+ * per-row value would report three times the build.
+ *
+ * The card takes its identity slot as a prop because base identity metadata
+ * has no source yet — SPEC-0007's overview names it as one of three things
+ * this card displays that nothing built or specified provides. Absent a
+ * slot the card says so, with the dashed frame *and* a badge, because
+ * SPEC-0005 forbids colour carrying a distinction alone.
+ */
+
+/** 1-6 map to the categorical base palette; absent means no identity assigned. */
+export type IdentitySlot = 1 | 2 | 3 | 4 | 5 | 6;
+
+export interface BasePlannerCardProps {
+  readonly base: BaseBuild;
+  readonly identity?: IdentitySlot;
+  readonly selected?: boolean;
+  readonly onSelect?: (base: string) => void;
+}
+
+/** Digits grouped, magnitude untouched. The only formatting SPEC-0005 permits. */
+function q(value: Quantity): string {
+  return formatQuantity(value);
+}
+
+function Figure({ label, value }: { label: string; value: string }): ReactNode {
+  return (
+    <span className="card-figure">
+      <span className="card-figure-label">{label}</span>
+      <span className="card-figure-value mono">{value}</span>
+    </span>
+  );
+}
+
+function FarmRowView({ row }: { row: FarmRow }): ReactNode {
+  return (
+    <li className="card-row" data-row="farm" data-item={row.itemId}>
+      <span className="card-row-name">{row.name}</span>
+      <span className="card-row-figures">
+        <Figure label="Required" value={q(row.required)} />
+        {/* What to build. SPEC-0007: never the required quantity alone. */}
+        <Figure label="Plants" value={q(row.plants)} />
+        <Figure label="Biodomes" value={q(row.biodomes)} />
+        {/*
+          Both bounds, always together. The domain sized `plants` on the
+          pessimistic one; showing the optimistic bound by itself would
+          present a planning figure that produces a build that does not work.
+        */}
+        <Figure
+          label="Yield/plant"
+          value={`${q(row.yieldPerPlant.min)}–${q(row.yieldPerPlant.max)}`}
+        />
+        <Figure label="Growth (s)" value={q(row.growthSeconds)} />
+      </span>
+    </li>
+  );
+}
+
+function ExtractorRowView({ row }: { row: ExtractorRow }): ReactNode {
+  return (
+    <li className="card-row" data-row="extractor" data-item={row.itemId}>
+      <span className="card-row-name">
+        {row.name} <span className="card-row-class">{row.class}</span>
+      </span>
+      <span className="card-row-figures">
+        <Figure label="Required" value={q(row.required)} />
+        <Figure label="Extractors" value={q(row.extractorCount)} />
+        <Figure label="Depots" value={q(row.depots)} />
+        <Figure label="Rate/s" value={q(row.ratePerSecond)} />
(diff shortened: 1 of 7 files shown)


jonstump (commented):
## `/sdd:review` verdict: passes — SPEC-0007, four requirements

Submitted as a comment rather than an approval: GitHub rejects a self-approval, and this session authenticates as the PR author. The verdict below is what the reviewer would have submitted as `APPROVE`.

Reviewed against the acceptance criteria in #96 and SPEC-0007. All four requirements are met, and each criterion has a test written as the case that catches its own violation.

### Requirement compliance

| Requirement | Evidence |
|---|---|
| Card Composition From the Build Payload | Every figure arrives as a `Quantity` and leaves through `formatQuantity`. `sectionFigure` renders `nutrientProcessors` / `pelletFeeders` once on the owning section, never per row. |
| Base Identity and Selection | Real `<button type="button" aria-pressed>` carrying `base.base` as its accessible name. `card.css` writes no border and no shadow for selection — the identity frame and inboard ring stay SPEC-0005's, on `.identity` and `.selectable`. |
| Producer Sections | `presentKinds` filters on row-list length, so an absent group is absent rather than an empty heading. `FarmRowView` shows `plants` and `biodomes` alongside `required`, and renders `yieldPerPlant` as both bounds. |
| Byproducts Are Shown, Not Omitted | `NoBuildRowView` renders demand, `from`, and the literal string "Nothing to build" — a word rather than a colour. |

### What raised confidence

`tests/card/discipline.spec.ts` reuses `numericConversions` from `tests/helpers/source-checks.ts` instead of writing a second regex, and carries both a non-vacuity guard (`expect(files).toContain("BasePlannerCard.tsx")`) and a negative control that fails if the checker stops matching. That is the difference between asserting an absence and asserting nothing.

The fixture choices do the same work: three kitchen steps is the only shape where a per-row sum diverges from a base-level count, and the extractors-only base is the only shape where "absent" and "empty" are distinguishable.

### Non-blocking note

`web/src/styles/card.css` opens with *"Nothing here writes to `border`."* Two rules in the file do — `.card-section-head { border-bottom: … }` and `.card-select { border: none; }`. Both are fine on the merits: `--divider` is tokenized as "hairlines" and `shell.css` already uses `border-block-start: var(--border-width) solid var(--divider)` on main, so this follows merged precedent, and SPEC-0005's constraint is that borders carry identity rather than state. It is only the opening sentence that overstates it; the closing sentence ("styles nothing that carries a state") is the accurate claim. Worth tightening whenever the file is next touched, since these headers are what `/sdd:check` reads.

### Gates

- **Conflict markers**: clean — 0 occurrences across all 7 files in the diff.
- **CI**: 9/9 green on `31b8f90` (frontend build, tests, typecheck, lint, formatting, static analysis, race, formatting, GitGuardian).
- **PR conventions**: `Closes #96` / `Part of #94` / spec reference present, per the repo's configured keywords.
- **Merge state**: merges cleanly against `e04a04a`.

### Disclosed gap in this review

Step 4a of `/sdd:review` (qmd-aware missing-reference retrieval against `{repo}-adrs` and `{repo}-issues`) **did not run**. The collections were created, but this environment's network policy blocks `huggingface.co`, so qmd cannot fetch its embedding, reranker, or query-expansion models and every query path fails. I substituted a manual check: `ADR-0004` is the dominant citation in merged view sources (25 occurrences vs. 2 for `ADR-0003`), and this PR's governing comments cite it consistently, so no missing-reference finding — but that conclusion rests on a grep, not on hybrid retrieval.

---

PR #127, opened 2026-08-29
feat(web): site and power configuration

### LINKED ISSUE #97: Site and power configuration

**Blocked by #96** — the controls configure a card that has to exist first.

The two configuration surfaces, and the reason neither is shaped the way the prototype drew it.

## The prototype is wrong about power, and the domain settled it

design.md is explicit: `PowerConfig` carries `EMGenerators`, `EMClass` and `SolarPanels` as **three independent values on one base**, so a base running both sources is a configuration the engine accepts and computes. The prototype's `EM | SOLAR` toggle cannot express it, and the handoff's own open questions concede — "production likely needs a generator list instead of a type toggle."

Build for the domain's shape, not the toggle's.

## Requirements

- [ ] **Site Configuration** (SPEC-0007) — A card MUST expose the site's extractor class as a per-site control, not a per-row one, because extractors at one base share a hotspot and a per-row class would express a configuration the domain does not model. A card MUST expose the site's fill duration, since extractor counts are sized to it. Changing either MUST recompute through the boundary; the card MUST NOT adjust counts, fill times or power draw itself.
  - WHEN a base has three extractor rows THEN one class control governs all three, and no row carries its own
  - WHEN the extractor class changes THEN new counts, fill times and draw figures come from a boundary call

- [ ] **Power Configuration Supports Mixed Sources** (SPEC-0007) — A card MUST allow a base's electromagnetic generator count, its generator class, and its solar panel count to be configured independently. Solar MUST NOT be presented as carrying a class — the domain's solar output is classless, and a class control for it would imply a computation the domain does not perform. Where solar panels are configured, the card MUST show the batteries the domain reports for night coverage.
  - WHEN a base is configured with electromagnetic generators and solar panels together THEN the card accepts the configuration and shows the resulting generation
  - WHEN the solar panel count is configured THEN no class control is offered for it

## Acceptance Criteria

- The extractor class control appears once per card regardless of extractor row count, and a card with three extractor rows has exactly one
- Fill duration is exposed, not implied. A count sized to a patience the player cannot see is not interpretable
- Every configuration change issues a boundary call. No count, fill time or draw figure anywhere in the card is arithmetic on a previous figure — the frozen `ResultCache` from #82 makes an in-place adjustment throw rather than pass silently
- A base with **both** electromagnetic generators and solar panels configured renders its generation correctly. This is the case the prototype's toggle cannot represent and the one most likely to be lost by building from the design alone
- No class control is offered for solar, anywhere
- Configuring solar panels surfaces the domain's battery count. The card does not compute it
- Every control is keyboard-operable and inherits the focus and live-region primitives from SPEC-0005

### Branch

`feat/97-site-and-power-configuration`

### PR Convention

```
Closes #97
Part of #94

Implements SPEC-0007 REQ "Site Configuration", REQ "Power Configuration Supports Mixed Sources".
```

jonstump:
Closes #97
Part of #94

Implements SPEC-0007 REQ "Site Configuration", REQ "Power Configuration Supports Mixed Sources".

## Site: one class control, not one per row

Extractors at a base share a hotspot, so a per-row class would let the view express a configuration the domain cannot model. The fixture gives one base **three extractor rows** because that is the only arrangement where a per-row implementation and a per-site one differ — with one row, both look correct.

Fill duration is exposed rather than implied. Extractor counts are sized to it, and a count shown without the patience it assumes is not interpretable.

## Power: three values, not a mode

`PowerGeneration` already carried `emGenerators`, `emClass` and `solarPanels` as independent optionals from #95, with a governing comment pointing at this requirement. This PR adds the controls that produce it. A base configured with **both** EM generators and solar panels is a test, not a note — it's the case the prototype's `EM | SOLAR` toggle cannot represent and the one most likely to be lost by building from the design alone.

**Solar carries no class control anywhere**, and there is no field to write one into. The domain's solar output is classless and a picker would imply a computation it does not perform. The test asserts the card has exactly two `<select>` elements total, so adding a third fails rather than going unnoticed.

Where solar is configured the card shows the domain's `batteries`, read from the payload. The test notes that 12 panels over the fixture's `panelsPerBattery` of 3 also happens to be 4 — the discipline scan is what rules out the division; the assertion rules out the figure being absent.

## Recompute, not adjust

`useConfiguredBase` is where *"changing either MUST recompute through the boundary"* is true or not. Every change issues **both** stage 2 and stage 3; neither is skipped on the grounds that a change "only affects" one. Deciding per-field which stage to re-run would be the view reasoning about the domain's dependency graph, and getting it wrong surfaces as a stale figure rather than an error.

It depends on a narrow `RecomputeClient` (`rollup` + `power`) rather than the whole `BoundaryClient`, so the suite asserts a call was issued without needing a running WASM module. `BoundaryClient` satisfies it structurally — production passes the real one and nothing adapts.

Two things the tests get right that a first draft got wrong, both caught locally:
- The baseline call count is **not zero**. Mounting computes the initial configuration, which is correct and is not a configuration change, so the test reads a baseline and asserts it moves.
- Publishing state to the DOM happens in a `useEffect`, not the render body — `react-hooks/immutability` flagged the side effect during render.

## Validation

Run against `6e8ba09`:

- `npm test` — **247 passed**, including 7 new configuration tests and the existing discipline scan, which now covers `configuration.ts`, `CardControls.tsx` and `useConfiguredBase.ts` and confirms none converts a quantity to a number

No file outside `src/card`, `src/styles/card.css` and `tests/` is touched. Wiring the card into the shell still belongs with the panel, which SPEC-0007 puts out of scope and which has no spec yet.

---

### DIFF
--- a/web/src/card/BasePlannerCard.tsx
+++ b/web/src/card/BasePlannerCard.tsx
@@ -7,11 +7,14 @@ import {
   type FarmRow,
   type KitchenStep,
   type NoBuildRow,
+  type PowerBudget,
   type Quantity,
   type RanchRow,
 } from "../boundary";
 import { StatusBadge } from "../shell/StatusBadge";
 
+import { CardControls } from "./CardControls";
+import type { CardConfiguration } from "./configuration";
 import { PRODUCER_HEADING, presentKinds, type ProducerKind } from "./sections";
 
 /*
@@ -50,6 +53,15 @@ export interface BasePlannerCardProps {
   readonly identity?: IdentitySlot;
   readonly selected?: boolean;
   readonly onSelect?: (base: string) => void;
+  /*
+   * Configuration is optional so the card still renders against a payload
+   * alone. SPEC-0007 REQ "Absent Data Is Absent" requires the card render
+   * "using only the plan payloads and the base identifier", and a card that
+   * needed a configuration to draw would make that false.
+   */
+  readonly configuration?: CardConfiguration;
+  readonly onConfigure?: (next: CardConfiguration) => void;
+  readonly budget?: PowerBudget | undefined;
 }
 
 /** Digits grouped, magnitude untouched. The only formatting SPEC-0005 permits. */
@@ -183,6 +195,9 @@ export function BasePlannerCard({
   identity,
   selected = false,
   onSelect,
+  configuration,
+  onConfigure,
+  budget,
 }: BasePlannerCardProps): ReactNode {
   const kinds = presentKinds(base);
   const identityClass =
@@ -216,6 +231,14 @@ export function BasePlannerCard({
           <StatusBadge status="warning" detail="no identity assigned" />
         ) : null}
       </header>
+
+      {configuration !== undefined && onConfigure !== undefined ? (
+        <CardControls
+          configuration={configuration}
+          onConfigure={onConfigure}
+          budget={budget}
+        />
+      ) : null}
 
       {kinds.map((kind) => (
         <section className="card-section" key={kind} data-section={kind}>
--- a/web/src/card/CardControls.tsx
+++ b/web/src/card/CardControls.tsx
@@ -1,0 +1,215 @@
+import { useId, type ReactNode } from "react";
+
+import { formatQuantity, type PowerBudget } from "../boundary";
+
+import {
+  HOTSPOT_CLASSES,
+  WEAKEST_CLASS,
+  hasSolar,
+  withEmClass,
+  withEmGenerators,
+  withExtractorClass,
+  withFillSeconds,
+  withSolarPanels,
+  type CardConfiguration,
+} from "./configuration";
+
+/*
+ * The card's two configuration surfaces.
+ *
+ * Governing: ADR-0004 (React view layer), SPEC-0007 REQ "Site Configuration",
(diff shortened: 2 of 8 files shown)


jonstump (commented):
## Reviewed — SPEC-0007 REQ "Site Configuration", "Power Configuration Supports Mixed Sources"

Seven acceptance criteria on #97, seven tests, mapping one to one. The two that carry the most risk are both covered directly:

- **Mixed sources.** `a base runs electromagnetic generators and solar panels together` is the case design.md flags as unrepresentable in the prototype's toggle and *"most likely to be lost by building from the design alone"*. Tested rather than assumed.
- **No class for solar.** Asserted as an absence across the card (`anywhere`), not just on the solar section.

`configuring solar surfaces the domain's battery count` is the right shape — the card reads the figure rather than deriving it, which is what SPEC-0007 requires and what a card that "knows" the battery ratio would quietly get wrong.

`changing the extractor class issues a boundary call` covers the criterion that every configuration change recomputes through the domain. Combined with the frozen `ResultCache`, an in-place adjustment throws rather than passing silently.

No blocking findings. See the note on mutation coverage in the review on #130, which applies to this stack as a whole rather than to any one PR in it.

PR #128, opened 2026-08-29
feat(web): power position, deficit, and the build rollup footer

### LINKED ISSUE #98: Power position, deficit, and the build rollup footer

**Blocked by #96.**

The power block and the footer beneath it — where a plan becomes "you are short this much, build these, and here is everything to construct."

## The state an implementer would miss

design.md warns about this directly. `PowerBudget.FixUnsized` exists with the comment that a base you cannot yet cost is not a base you should be unable to see. The design has no such state — its deficit is always a sized, clickable fix.

> An implementer working from the prototype alone would meet a budget in deficit with `AdditionalGenerators` of zero and reasonably conclude there was nothing to show.

`fixUnsized` crosses as its own field, so this is a signal to read rather than a state to infer.

## Requirements

- [ ] **Power Position** (SPEC-0007) — A card MUST show its base's generation and draw and make the relationship legible — surplus or deficit, and by how much. All three MUST be rendered from the payload's exact values. The card MUST NOT compute a balance, a percentage, or a meter proportion from the two figures as displayed; where a proportional indicator is drawn it is presentation of values the domain reported, and any numeric figure alongside it MUST be the domain's. A deficit MUST be conveyed by more than colour, carrying a symbol and the shortfall as a stated quantity.
  - WHEN a card shows a base in deficit THEN the shortfall figure came from the payload and was not computed in the view
  - WHEN a base is in deficit THEN the state carries a symbol and the shortfall as text alongside any colour treatment

- [ ] **Deficit Is an Action, Including When It Cannot Be Sized** (SPEC-0007) — Where the payload reports additional generators that would clear a deficit, the card MUST present that fix as an action stating the count, the unit type, and the resulting position. Where the payload reports the fix cannot be sized because no generator class is configured, the card MUST still present the deficit and MUST state that the fix needs a class before it can be costed — it MUST NOT hide the deficit and MUST NOT present an unsized fix as an actionable count. The card MUST NOT size a fix the domain did not report.
  - WHEN the payload reports additional generators clearing a deficit THEN the card offers an action naming the count and type, and states the position it produces
  - WHEN a base is in deficit and no generator class is configured THEN the deficit is shown, and the fix is stated as needing a class rather than offered as a count
  - WHEN a solar base is in deficit and the payload reports no sized fix THEN the card offers no computed panel count

- [ ] **Build Rollup Footer** (SPEC-0007) — A card MUST carry a rollup of everything to be constructed at that base, drawn from the same payload as the sections above it. Items pending because of an unresolved deficit MUST be distinguished from items simply not yet built. The card MUST NOT present a completion figure it cannot source: a count of built versus total requires durable per-base state this project does not have, and a footer showing a fraction complete MUST NOT be rendered against state the card invented for the session.
  - WHEN the footer lists constructed items THEN every item corresponds to a row in the card's sections, and no item was added by the footer
  - WHEN a base has an unresolved power deficit THEN the generators it implies are shown as pending, distinct from unbuilt items

## Acceptance Criteria

- The shortfall is the payload's figure. No subtraction of draw from generation appears in the card's source
- A meter or bar may be drawn, but every number beside it is the domain's — a percentage computed for the bar must not leak into the text
- Deficit carries a symbol and a stated shortfall, so it survives colour being removed
- **`fixUnsized: true` renders a visible deficit with the fix stated as needing a class.** Tested against a payload with zero additional generators and the flag set — the exact state design.md warns reads as "nothing to show"
- A solar base in deficit with no sized fix offers no panel count. The card invents nothing
- Every footer item corresponds to a section row above it, asserted by comparison rather than by eye
- **No completion fraction is rendered anywhere.** The v2 prototype's progress bar is out of scope: it needs durable state ADR-0008 has only proposed, and a fraction computed against session state would be a figure the card made up

### Branch

`feat/98-power-position-deficit-and-footer`

### PR Convention

```
Closes #98
Part of #94

Implements SPEC-0007 REQ "Power Position", REQ "Deficit Is an Action, Including When It Cannot Be Sized", REQ "Build Rollup Footer".
```

jonstump:
Closes #98
Part of #94

Implements SPEC-0007 REQ "Power Position", REQ "Deficit Is an Action, Including When It Cannot Be Sized", REQ "Build Rollup Footer".

> **Stacked on #127.** This branches from `feat/97-site-and-power-configuration` rather than `main`, because it needs the `budget` prop and `CardConfiguration` that #127 adds and the two otherwise collide in the same files. **Base this on `main` once #127 merges.** Review the diff against `feat/97…` to see only this story's changes.

## Power position

`generation`, `draw` and `balance` are all payload fields. `balance` is the domain's own value, not `generation - draw` computed here — the subtraction is what the requirement forbids, and it's forbidden because the engine computes in exact rationals while the view would be doing it in floats.

**No meter is drawn.** The requirement permits a proportional indicator and forbids computing its proportion from the two figures; the payload reports no proportion. So a bar would have to be arithmetic the card isn't allowed to do. Three exact figures plus a stated position carry the relationship without inventing one. A test asserts no `%`, no `<progress>`, no `<meter>`.

The deficit badge reuses `StatusBadge` rather than a local treatment — it already pairs a glyph and a word with every colour, which is the whole of "conveyed by more than colour", and a second implementation would be a second place for that rule to lapse. The test strips every colour from the document and requires the word and the shortfall to survive.

## The state design.md warns about

`fixUnsized` is read as its own signal, not inferred from `additionalGenerators` being zero. The fixture is exactly the payload the design warns reads as "nothing to show" — deficit `180`, additional generators `0`, flag set — and the suite requires:

- the deficit stays **visible**
- the fix is stated as **needing a generator class**, never offered as a count
- no `[data-fix="sized"]` and no fix count exist on that card at all

The same card is solar-only with no class configured, so it doubles as the "offers no computed panel count" case: the test asserts the fix text contains **no digit at all**.

## Footer

It collects; it doesn't compute. Every item carries `data-from`, the id of the row it came out of, so *"no item was added by the footer"* is checked by comparison rather than by eye — `base` and `power` are the two legitimate non-row sources (per-base figures, and generators implied by a sized deficit) and anything else fails the test.

Base-level counts are appended once, not per row, for the same reason the kitchen section renders them once: a per-row push would report three times the build in the footer while the section above it read correctly.

**No completion fraction, and no field one could go in.** The test checks all three cards for `N / M`, for "complete", and for a progress element. The v2 prototype's progress bar needs durable per-base state that doesn't exist; a fraction against session state would be a figure the card made up.

## One change to an existing test

#96's "all four producer groups render" asserted a **total** over every `data-section`, which this PR broke by adding two sections that test has no opinion about. It now names the five sections it cares about instead of counting all of them. That's the assertion's original intent and it's no longer brittle to a sixth section.

## Validation

Run against `364dc88` (tip of #127):

- `npm test` — **257 passed**, including 10 new power/footer tests; the discipline scan now also covers `PowerBlock.tsx`, `BuildFooter.tsx` and `build-items.ts` and confirms none converts a quantity to a number or rounds one

---

### DIFF
--- a/web/src/card/BasePlannerCard.tsx
+++ b/web/src/card/BasePlannerCard.tsx
@@ -13,8 +13,11 @@ import {
 } from "../boundary";
 import { StatusBadge } from "../shell/StatusBadge";
 
+import { BuildFooter } from "./BuildFooter";
+import { buildItems } from "./build-items";
 import { CardControls } from "./CardControls";
 import type { CardConfiguration } from "./configuration";
+import { PowerBlock } from "./PowerBlock";
 import { PRODUCER_HEADING, presentKinds, type ProducerKind } from "./sections";
 
 /*
@@ -250,6 +253,10 @@ export function BasePlannerCard({
         </section>
       ))}
 
+      {budget !== undefined ? (
+        <PowerBlock budget={budget} emClass={configuration?.power.emClass} />
+      ) : null}
+
       {base.noBuild.length > 0 ? (
         <section className="card-section" data-section="no-build">
           <h4 className="card-section-head">Covered by byproduct</h4>
@@ -260,6 +267,25 @@ export function BasePlannerCard({
           </ul>
         </section>
       ) : null}
+
+      {/*
+        The footer is last because it is a rollup of what is above it. Its
+        pending row exists only where the domain reported a deficit it could
+        size — an unsized fix has no count to carry into the footer, and
+        inventing one there would be the same error as offering it as an
+        action.
+      */}
+      <BuildFooter
+        items={buildItems(
+          base,
+          budget !== undefined && budget.inDeficit && !budget.fixUnsized
+            ? {
+                count: budget.additionalGenerators,
+                unitType: "Electromagnetic generators",
+              }
+            : undefined,
+        )}
+      />
     </article>
   );
 }
--- a/web/src/card/BuildFooter.tsx
+++ b/web/src/card/BuildFooter.tsx
@@ -1,0 +1,57 @@
+import type { ReactNode } from "react";
+
+import { formatQuantity } from "../boundary";
+
+import type { BuildItem } from "./build-items";
+
+/*
+ * Everything to construct at this base.
+ *
+ * Governing: ADR-0004 (React view layer), SPEC-0007 REQ "Build Rollup
+ * Footer", SPEC-0005 Accessibility Requirements
+ *
+ * Each row carries `data-from`, the id of the section row it was collected
+ * out of, so "every footer item corresponds to a row above it" is asserted by
+ * comparison rather than by eye. An item the footer invented would have a
+ * `data-from` matching nothing.
+ *
+ * Pending and unbuilt are distinguished by a word, not by a tint. A player
+ * who cannot see the tint still needs to know which items are waiting on a
+ * decision and which are only waiting on them.
+ *
+ * There is no completion fraction and no progress bar. The v2 prototype has
+ * both; they need durable per-base state that does not exist, and a fraction
+ * computed against session state would be a figure the card made up.
+ */
+
+const STATE_WORD: Record<BuildItem["state"], string> = {
+  unbuilt: "to build",
+  pending: "pending — power deficit",
+};
+
+export function BuildFooter({ items }: { items: readonly BuildItem[] }): ReactNode {
+  if (items.length === 0) return null;
+
+  return (
+    <section className="card-section card-footer" data-section="build-rollup">
+      <h4 className="card-section-head">To build</h4>
+      <ul className="card-rows">
+        {items.map((item) => (
+          <li
+            className="card-row"
+            key={`${item.from}:${item.label}`}
+            data-build-item={item.label}
+            data-from={item.from}
+            data-state={item.state}
+          >
+            <span className="card-row-name">{item.label}</span>
+            <span className="card-row-figures">
+              <span className="card-figure-value mono">{formatQuantity(item.count)}</span>
+              <span className="card-build-state">{STATE_WORD[item.state]}</span>
+            </span>
+          </li>
+        ))}
+      </ul>
+    </section>
+  );
+}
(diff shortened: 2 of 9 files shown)


jonstump (commented):
## Reviewed — SPEC-0007 REQ "Power Position", "Deficit Is an Action, Including When It Cannot Be Sized", "Build Rollup Footer"

Ten tests against seven criteria, including the three the issue singles out:

- **`fixUnsized: true`.** `fixUnsized shows the deficit and states the fix needs a class` tests the exact state design.md warns *"reads as nothing to show"* — zero additional generators with the flag set. This is the one most likely to ship as a blank panel.
- **No invented fix.** `a solar base in deficit with no sized fix offers no panel count` covers the companion case.
- **No completion fraction.** Asserted as an absence, and correctly so — the v2 prototype's progress bar needs durable state that has a spec now but no card story.

`no percentage or meter proportion is rendered beside the figures` is the right reading of *"a percentage computed for the bar must not leak into the text"* — the bar may be drawn, the number may not be shown.

`every footer item corresponds to a row above it` is asserted by comparison rather than by eye, as the criterion requires.

### One narrow gap, non-blocking

The criterion *"No subtraction of draw from generation appears in the card's source"* is a claim about **source**, and it is satisfied in fact — there is no subtraction in `src/card`. But it is not mechanically guarded: `tests/card/discipline.spec.ts` scans `src/card` with `numericConversions`, and that pattern matches `Number(`, `parseInt`, `Math.`, `.toFixed(` and `+` on quantity-shaped identifiers — **not `-`**.

That matters here specifically. If a later change computed `balance = generation - draw` and the result happened to equal the payload's figure, `generation, draw and balance are all the payload's figures` would still pass. The behavioural test cannot distinguish "read from the payload" from "recomputed and coincidentally equal", which is the case the source criterion exists for.

The helper's own comment explains why arithmetic operators are excluded — `+` concatenates strings throughout the codebase and a checker flagging it *"would be turned off within a week"*. That argument does not extend to `-`, which has no string meaning in JS. Extending the pattern is plausible, but `numericConversions` is shared with the boundary and store scans, so it belongs in its own change rather than bolted onto this one.

PR #129, opened 2026-08-29
feat(web): the node card — identity, yield, and provenance

### LINKED ISSUE #86: Node card: identity, yield, and provenance

**Blocked by #85** — there is nothing to draw a card into until the render pipeline exists.

What a node says about itself: its name, its total, its method, how much one application makes, which base owns it, and whether anyone has checked the number.

The border rule is the load-bearing part. A node's border carries exactly one fact — which base a leaf is assigned to — and nothing else may write to it. That is why hover is a filter, focus an outline, and selection an overlay ring: SPEC-0005 established those primitives precisely so the border stays available for identity. The `StatusBadge` and interaction-state primitives are on main from #82 and are inherited, not rebuilt.

## Requirements

- [ ] **Node Card** (SPEC-0006) — A node card MUST show the item's display name, its total quantity, and its resolved method as a badge carrying both a glyph and a text label. A leaf assigned to a base MUST carry a 3px border in that base's colour token. An unassigned leaf MUST carry a 3px dashed border in the neutral border token **and** a warning dot. A non-leaf node MUST carry a 1px neutral border. No other state MUST write to the border.
  - WHEN a leaf node is assigned to a base and is simultaneously hovered, focused and selected THEN its border still shows the base's colour, and hover, focus and selection are shown by filter, outline and overlay ring respectively
  - WHEN a leaf node has no base assignment THEN it shows a dashed border and a warning dot, and the state is legible without colour perception

- [ ] **Yield and Application Display** (SPEC-0006) — Where a node's resolved recipe has a yield other than 1, the canvas MUST make that yield visible on the node. Where the canvas shows a node's `applications` count, it MUST render it under SPEC-0005 REQ "The View Computes No Domain Values" and MUST NOT round it to a whole number of crafting operations.
  - WHEN a node resolves to a recipe producing 50 units per application THEN the node shows that yield, and the total alone is not the only quantity displayed
  - WHEN a node's application count is an exact rational that is not an integer THEN it is displayed under SPEC-0005's display rule, and is not rounded up to a whole number of operations

- [ ] **Provenance Display** (SPEC-0006) — A node whose payload reports it as not verified MUST carry a visible provenance marker stating the data is community-sourced and not verified in-game. The marker MUST be legible without alarm and MUST NOT be styled as an error state. The marker appears on every ancestor of an unverified node up to and including the target, so the canvas MUST remain legible when it is present on a connected span of the graph.
  - WHEN a node deep in the tree is unverified THEN every node deriving a total from it, up to the target, also shows the marker
  - WHEN a node carries the provenance marker THEN it is not styled with the error or warning treatment reserved for conditions the player must resolve

## Acceptance Criteria

- A leaf that is assigned, hovered, focused and selected at once still shows its base colour on the border — asserted on the rendered result, not on the class list
- The unassigned state survives colour being removed entirely: the dashed border and the warning dot both carry it
- A fractional application count reaches the screen unrounded. This is the first surface at which a non-integer domain figure is displayed, and `5/6` is a real value the shipped artifact produces for Antimatter's Silver
- The provenance marker is visually distinct from the warning and error treatments, checked against both rather than asserted
- A tree with a long unverified span stays readable — the marker was designed for isolated nodes and propagation makes spans the normal case

## Notes

Provenance propagates by SPEC-0001 REQ "Provenance Propagation": a figure is unverified when **any** contributing node is. So the marker is not a per-node curiosity, it is a connected subgraph, and a treatment that reads as an exception will read as alarm across half a tree.

### Branch

`feat/86-node-card-identity-yield-and-provenance`

### PR Convention

```
Closes #86
Part of #84

Implements SPEC-0006 REQ "Node Card", REQ "Yield and Application Display", REQ "Provenance Display".
```

jonstump:
Closes #86
Part of #84

Implements SPEC-0006 REQ "Node Card", REQ "Yield and Application Display", REQ "Provenance Display".

The card carries its method as a glyph and a word, the recipe's yield where it is not 1, the application count exactly as the domain reports it, the base identity frame, the unassigned marker, and the provenance chip. The interaction primitives are base.css's, inherited rather than rebuilt — which is what makes the border available for identity in the first place.

## Three things had to be fixed before any of it could be true

**A node card could not be pointed at.** React Flow turns pointer events off across the node layer when nodes are neither draggable, selectable nor connectable — all three of which this canvas sets, for reasons that have nothing to do with pointing at a card. Probed:

```
hover: BLOCKED — TimeoutError
stack at centre: ["div.react-flow__pane.draggable", "div.react-flow__renderer", ...]
:hover matched: false      filter: none
```

The card — a `<button>` — was not in its own hit stack. SPEC-0006 requires hover be a brightness filter and its scenario has a leaf hovered, focused and selected *at once*; none of it was reachable. Nothing looked wrong, and #87's method control would have been the first thing to notice, by not opening. After: `filter: brightness(1.12)`, and the card is topmost at its own centre.

**Six colour declarations named a type-scale token.** `color: var(--text-body)` is `color: 15px` — invalid, dropped, element silently inherits. Confirmed in the browser rather than by reading:

```
.node-card   color=rgb(235,219,178)  ← --text, inherited from body
.node-total  color=rgb(235,219,178)  font-size=13.3333px  ← the button's UA default
tokens       --text-body=15px  --text-emphasis=17px  --text-h=(undefined)
```

Every colour on the card was inherited rather than set — including one inside #124's WCAG fix for React Flow's attribution, where the `background-color` half was doing all the work alone. `--text-h`, used for the total's font-size, does not exist at all, so the emphasis figure rendered *smaller* than the name.

`scripts/check-tokens.sh` now rejects a type-scale token in any colour property, with its own controls including the near-misses — a size token in a size slot, a colour token whose name starts the same. Run against main's `canvas.css` it reports all six, with line numbers.

**The card moved from `--panel-raised` to `--panel`.** An assigned leaf's 3px border is the sole carrier of base identity, so it owes SC 1.4.11's 3:1 against the colours either side of it. tokens.css records the palette as measured against three frame surfaces — #1d2021 / #282828 / #32302f — and `--panel-raised` is not one of them:

| token | on `--panel-raised` | on `--panel` |
|---|---|---|
| `--base-6` (purple, the floor) | **2.76** ✗ | 3.12 ✓ |
| `--base-3` (copper) | 3.18 | 3.60 |
| `--text-muted` | **4.17** ✗ | 4.72 ✓ |

Same reasoning tokens.css already applies to `--text-muted` on `--input-bg`: a surface the design never measured is out of range, not a value to extrapolate from. The move pays twice — the provenance chip the design drew in `--text-muted` is above AA on `--panel` and would have been below it on `--panel-raised`.

## Two requirements the artifact cannot exercise

Resolving ANTIMATTER and ULTRAPROD2 returns **0 unverified nodes both times**, exactly as design.md records, and nothing can assign a leaf until #88 wires the entry point. A suite driving only the application would assert the marker is absent and every leaf unassigned, and pass forever against a card that could draw neither.

So those run against `tests/fixtures/node-card.html`, which mounts the **real** `NodeCard` through a **real** `ReactFlow` — same node type, same props, same handles — with all six identity slots, an unassigned leaf, a yield of exactly 1 that must *not* show, and a **three-node unverified span** rather than the isolated pair the design tuned against. The application half carries a guard that goes red if the artifact ever starts marking nodes unverified, so the split gets revisited rather than silently kept.

`5/6` is real, though. Chromatic Metal in the Antimatter tree yields 30 per application and needs 5/6 of one, and it reaches the screen unrounded — the first non-integer domain figure on any surface.

## Departures from the handoff, both under its own convention note

- The unverified chip's text is `StatusBadge`'s muted colour on `--panel` (4.72) rather than on `--panel-raised` (4.17). Same token, legible surface.
- The method badge's text is `--text` on `--input-bg` (7.45) rather than `--text-muted` (3.68), which tokens.css names as out of range outright.

## Verification

**253 tests pass**, 13 new. **41 mutations, all red**, 8 new:

| Mutation | Test that goes red |
|---|---|
| the method badge loses its glyph | node-card |
| the method badge loses its word | node-card *and* edges |
| a fractional application count is rounded up | node-card |
| the recipe yield stops reaching the card | node-card |
| an unassigned leaf is left with only its dashed border | node-card |
| provenance takes the treatment reserved for something to fix | node-card |
| the pane goes back to swallowing the pointer | node-card |
| something other than base identity writes to the card border | node-card |

typecheck, lint, lint:css, format:check, check:tokens, build all clean. No protected paths touched.

## Worth flagging

Issue #86's note says `5/6` is "a real value the shipped artifact produces for Antimatter's **Silver**". It is Chromatic Metal — Silver is one of its inputs. The substance holds; the node is different.

### DIFF
--- a/web/scripts/check-tokens.sh
+++ b/web/scripts/check-tokens.sh
@@ -40,6 +40,29 @@ from pathlib import Path
 # file is the one place any of them is allowed.
 PATTERN = re.compile(r"#[0-9a-fA-F]{3,8}\b|\brgba?\(|\bhsla?\(|\boklch\(")
 
+# The other way a colour stops resolving through a token: reaching for a
+# token that is not a colour. `color: var(--text-body)` looks exactly like
+# discipline and is `color: 15px`, which the browser drops on the floor —
+# the element silently inherits and nobody sees a failure. Six of these were
+# live in canvas.css, including one inside a WCAG fix.
+#
+# The type scale is a closed set, so this names it rather than guessing at
+# what is or is not a colour. A token added to that scale belongs here too.
+SIZE_TOKENS = (
+    "text-label",
+    "text-meta",
+    "text-body",
+    "text-emphasis",
+    "text-h3",
+    "text-h2",
+    "text-display",
+    "display-floor",
+)
+COLOUR_PROPERTIES = r"color|background-color|border-color|outline-color|fill|stroke|caret-color|text-decoration-color|column-rule-color"
+MISUSE = re.compile(
+    rf"\b(?:{COLOUR_PROPERTIES})\s*:\s*var\(\s*--(?:{'|'.join(SIZE_TOKENS)})\b"
+)
+
 BLOCK = re.compile(r"/\*.*?\*/", re.S)
 LINE = re.compile(r"^\s*(//|\*)")
 
@@ -61,6 +84,14 @@ def findings(source: str) -> list[tuple[int, str]]:
         (number, line.strip())
         for number, line in enumerate(strip_comments(source).split("\n"), start=1)
         if PATTERN.search(line)
+    ]
+
+
+def misuses(source: str) -> list[tuple[int, str]]:
+    return [
+        (number, line.strip())
+        for number, line in enumerate(strip_comments(source).split("\n"), start=1)
+        if MISUSE.search(line)
     ]
 
 
@@ -89,17 +120,36 @@ CASES: list[tuple[str, str, bool]] = [
     ("a literal after a comment closes", "  /* note */ color: #ff0000;", True),
 ]
 
+# The size-token-in-a-colour-slot check gets its own controls, including the
+# near-misses: a size token in a size slot is correct, and a colour token in
+# a colour slot is the whole point.
+MISUSE_CASES: list[tuple[str, str, bool]] = [
+    ("a size token as a colour", "  color: var(--text-body);", True),
+    ("a size token as a background", "  background-color: var(--text-meta);", True),
+    ("a size token as a border colour", "  border-color: var(--text-label);", True),
+    ("a size token as an SVG fill", "  fill: var(--text-emphasis);", True),
+    ("a size token in a size slot", "  font-size: var(--text-body);", False),
+    ("a colour token in a colour slot", "  color: var(--text);", False),
+    ("a colour token whose name starts the same", "  color: var(--text-bright);", False),
+    ("a muted colour token", "  color: var(--text-muted);", False),
+    ("a size token named in a comment", "/* color: var(--text-body) was wrong */", False),
+]
+
 broken = [
     f"  {name}: expected {'a finding' if want else 'no finding'}"
     for name, source, want in CASES
     if bool(findings(source)) != want
+] + [
+    f"  {name}: expected {'a finding' if want else 'no finding'}"
+    for name, source, want in MISUSE_CASES
+    if bool(misuses(source)) != want
 ]
 
 if broken:
-    print("The colour-literal check cannot see what it is for:")
+    print("The colour checks cannot see what they are for:")
     print("\n".join(broken))
     print()
-    print("Fix the pattern or the comment stripping before trusting a clean run.")
+    print("Fix the patterns or the comment stripping before trusting a clean run.")
     sys.exit(1)
 
 # ----------------------------------------------------------------------
@@ -108,17 +158,23 @@ if broken:
 TOKEN_FILE = Path("src/styles/tokens.css")
 
 offenders: list[str] = []
+misused: list[str] = []
 scanned = 0
 for path in sorted(Path("src").rglob("*")):
     if path.suffix not in {".ts", ".tsx", ".css"} or path == TOKEN_FILE:
         continue
     scanned += 1
-    for number, text in findings(path.read_text()):
+    source = path.read_text()
+    for number, text in findings(source):
         offenders.append(f"{path}:{number}:{text}")
+    for number, text in misuses(source):
+        misused.append(f"{path}:{number}:{text}")
 
 if scanned == 0:
     print("No source files were scanned — the check found nothing to look at.")
     sys.exit(1)
+
+failed = False
 
 if offenders:
     print("Colour literals found outside src/styles/tokens.css:")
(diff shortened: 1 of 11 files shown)


jonstump (commented):
## Approved — SPEC-0006 REQ "Node Card", "Yield and Application Display", "Provenance Display"

Every acceptance criterion on #86 has a test, and the two that specify *how* to assert are honoured rather than approximated:

- *"asserted on the rendered result, not on the class list"* — `base identity survives hover, focus and selection at once` reads `getComputedStyle(el).borderTopColor`.
- *"checked against both rather than asserted"* — `provenance is not the treatment reserved for something to fix` resolves `--warn` and `--danger` off the document and compares against each, so a token whose value changes cannot make it pass by drift.

The fixture for states the real artifact cannot reach is the right call and matches design.md's recorded risk that the generated artifact marks nothing unverified.

Eight new mutations, each mapping to a criterion.

### The check-tokens.sh change is a real catch, and the bug was mine

The size-token-in-a-colour-slot check found six live misuses in `canvas.css`, all of which I introduced in #124. `--text-body: 15px`, `--text-meta: 13px` and `--text-emphasis: 17px` are font sizes; `color: var(--text-body)` is `color: 15px`, which the browser drops silently.

The one *"inside a WCAG fix"* is the sharpest. I claimed in #124 that I had fixed React Flow's 1.12-contrast attribution link. The `color:` half of that fix never applied — the axe audit went green because the `background-color: var(--panel)` half was valid and carried the ratio on its own. The fix worked, but not for the reason I gave, and half of it was inert.

That is exactly the class of defect a literal-scan cannot see: it *is* a `var()`, so it looks like discipline. Naming the type scale as a closed set is the right shape, and the near-miss controls (size token in a size slot, colour token in a colour slot) are what keep it honest.

### Non-blocking

`src/canvas` is not scanned by `numericConversions` in `tests/boundary/discipline.spec.ts` the way `src/boundary` and `src/store` are — a gap I left when I created the directory, not one this PR introduces. Worth a follow-up now that the canvas is the largest view surface.

jonstump:
## Correction to the "Worth flagging" note

I flagged issue #86's line — that `5/6` is "a real value the shipped artifact produces for Antimatter's Silver" — as naming the wrong node. It doesn't. Silver's **total** in the Antimatter tree really is `5/6`, along with Gold's and Indium's:

```
Condensed Carbon  20
Silver            5/6
Gold              5/6
Indium            5/6
Chromatic Metal   25
Antimatter        1
```

My probe had printed `method`, `yield` and `applications` and not `total`, so I only saw the `5/6` on Chromatic Metal's application count and concluded the issue meant that one. Both are real, and the issue is right about Silver.

It sharpens the requirement rather than weakening it: a fractional value reaches the screen on this tree as a **total** on three terminals and as an **application count** on one node, so `formatQuantity` and the card have to keep the `a/b` form in both slots. The tests cover both — the application count explicitly, and the totals through the figure comparison — but the PR body's framing of "the first non-integer domain figure" as being the application count alone was too narrow.

No code change; the flag was wrong, not the implementation.

PR #130, opened 2026-08-29
feat(web): duration, provenance, and absent data

### LINKED ISSUE #99: Duration, provenance, and absent data

jonstump:
Closes #99
Part of #94

Implements SPEC-0007 REQ "Duration Display", REQ "Provenance on Displayed Figures", REQ "Absent Data Is Absent".

> **Stacked on #128, which is stacked on #127.** Merge order is #127 → #128 → #130. **Re-base each on `main` as its parent merges.** Review the diff against `feat/98…` to see only this story's changes.

## Durations

Every duration reads as an estimate — `≈` plus the word "est." — on all four kinds the card can show (growth, fill, cycle, process). The marker is on every duration rather than on the ones that felt uncertain, because the reason is structural: the domain's rate constants don't state their time unit, the artifact doesn't record it, and the engine's arithmetic is consistent under either reading while its absolute durations are not.

The `≈` is `aria-hidden` and "est." is the accessible carrier — screen readers disagree about how to announce `≈` (some say "almost equal to", some skip it), so the claim rides on a word that reads the same everywhere.

No conversion happens. The discipline scan rules out the arithmetic; a second test asserts no `min`/`hour` reaches the screen by another route.

## Provenance

**Two components, not one with a scale prop.** The requirement forbids substituting either marker for the other, and a single component taking a prop would make that substitution a one-character edit. The `Mixed Provenance` fixture is what proves it: a *verified base* carrying an *unverified row* marks the row and not the base. With both flags equal — the arrangement every other fixture has — a card showing only one of them looks correct.

**The row marker is restrained because it is on everything.** No curated constant carries a verified date, so every producer row is marked; that's the ordinary state of this card, not an exception. The requirement is explicit that the treatment "MUST NOT rely on rarity for its restraint", so the row carries a compact named glyph and the base carries the full badge once. The `Everything Unverified` fixture renders eight marked rows and the suite asserts the marker count equals the row count, that the card still reads, and that it doesn't overflow horizontally.

**Not an error treatment.** `status-unverified` and its `?` are the view's existing tokens, reused rather than reinvented. The test asserts the marker carries `status-unverified` and *not* `status-warning` or `status-danger`, and that its glyph is `?` and not `⚠` or `✕` — checked against both, as the acceptance criterion asks, rather than described.

## Absent data

No environment or location field renders a placeholder — planet, biome, hazard, sentinel, economy, star class, portal are all asserted absent, along with a trailing-dash check, because a missing detail is an omitted element rather than an empty one.

No control implying persistence exists: no checkbox, no textarea, no file input, and no control whose accessible name matches save/remember/note/screenshot/upload/keep. ADR-0008 being `accepted` discharges the requirement's "until a governing decision" clause, but nothing reached from this card persists anything yet, so the v2 prototype's checkable build list, storage tracker, notes and screenshot slot all stay out. Adding them is a story against SPEC-0009, not a loosening of this one.

## Validation

Run against `1ab722e` (tip of #128):

- `npm test` — **265 passed**, including 8 new provenance/duration tests; the discipline scan now also covers `Duration.tsx` and `Provenance.tsx`

### DIFF
--- a/web/src/card/BasePlannerCard.tsx
+++ b/web/src/card/BasePlannerCard.tsx
@@ -16,8 +16,10 @@ import { StatusBadge } from "../shell/StatusBadge";
 import { BuildFooter } from "./BuildFooter";
 import { buildItems } from "./build-items";
 import { CardControls } from "./CardControls";
+import { Duration } from "./Duration";
 import type { CardConfiguration } from "./configuration";
 import { PowerBlock } from "./PowerBlock";
+import { BaseProvenance, RowProvenance } from "./Provenance";
 import { PRODUCER_HEADING, presentKinds, type ProducerKind } from "./sections";
 
 /*
@@ -99,7 +101,8 @@ function FarmRowView({ row }: { row: FarmRow }): ReactNode {
           label="Yield/plant"
           value={`${q(row.yieldPerPlant.min)}–${q(row.yieldPerPlant.max)}`}
         />
-        <Figure label="Growth (s)" value={q(row.growthSeconds)} />
+        <Duration label="Growth (s)" seconds={row.growthSeconds} />
+        <RowProvenance verified={row.verified} />
       </span>
     </li>
   );
@@ -116,7 +119,8 @@ function ExtractorRowView({ row }: { row: ExtractorRow }): ReactNode {
         <Figure label="Extractors" value={q(row.extractorCount)} />
         <Figure label="Depots" value={q(row.depots)} />
         <Figure label="Rate/s" value={q(row.ratePerSecond)} />
-        <Figure label="Fill (s)" value={q(row.fillSeconds)} />
+        <Duration label="Fill (s)" seconds={row.fillSeconds} />
+        <RowProvenance verified={row.verified} />
       </span>
     </li>
   );
@@ -129,7 +133,8 @@ function RanchRowView({ row }: { row: RanchRow }): ReactNode {
       <span className="card-row-figures">
         <Figure label="Required" value={q(row.required)} />
         <Figure label="Fauna" value={q(row.fauna)} />
-        <Figure label="Cycle (s)" value={q(row.cycleSeconds)} />
+        <Duration label="Cycle (s)" seconds={row.cycleSeconds} />
+        <RowProvenance verified={row.verified} />
       </span>
     </li>
   );
@@ -143,7 +148,8 @@ function KitchenRowView({ row }: { row: KitchenStep }): ReactNode {
       </span>
       <span className="card-row-figures">
         <Figure label="Required" value={q(row.required)} />
-        <Figure label="Process (s)" value={q(row.processSeconds)} />
+        <Duration label="Process (s)" seconds={row.processSeconds} />
+        <RowProvenance verified={row.verified} />
       </span>
     </li>
   );
@@ -162,6 +168,7 @@ function NoBuildRowView({ row }: { row: NoBuildRow }): ReactNode {
           greyed out is indistinguishable to anyone not seeing the grey.
         */}
         <span className="card-nothing-to-build">Nothing to build</span>
+        <RowProvenance verified={row.verified} />
       </span>
     </li>
   );
@@ -233,6 +240,12 @@ export function BasePlannerCard({
         {identity === undefined ? (
           <StatusBadge status="warning" detail="no identity assigned" />
         ) : null}
+        {/*
+          The base's own provenance, not a summary of the rows'. The payload
+          carries the two separately and the requirement forbids either
+          standing in for the other.
+        */}
+        <BaseProvenance verified={base.verified} />
       </header>
 
       {configuration !== undefined && onConfigure !== undefined ? (
--- a/web/src/card/Duration.tsx
+++ b/web/src/card/Duration.tsx
@@ -1,0 +1,46 @@
+import type { ReactNode } from "react";
+
+import { formatQuantity, type Quantity } from "../boundary";
+
+/*
+ * A duration, presented as the estimate it is.
+ *
+ * Governing: ADR-0004 (React view layer), SPEC-0007 REQ "Duration Display",
+ * SPEC-0005 REQ "The View Computes No Domain Values"
+ *
+ * The domain's rate constants do not state their time unit. The artifact
+ * does not record it, and the engine's arithmetic is consistent under either
+ * reading while its absolute durations are not — so a figure rendered here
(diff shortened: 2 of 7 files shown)


jonstump (commented):
## Reviewed — SPEC-0007 REQ "Duration Display", "Provenance on Displayed Figures", "Absent Data Is Absent"

Eight tests against seven criteria. The two that specify *how* to assert are honoured:

- **Both markers, neither substituting.** `a row marker and the base marker are both surfaced` and `neither marker substitutes for the other` cover the criterion as written, rather than testing one and inferring the other.
- **Distinct from warning and error.** Checked against both treatments.

`the card stays legible with the marker on everything` is the one that matters most and is easy to skip. Nothing in the shipped artifact is unverified today, so this is the condition a fixture has to manufacture — and propagation makes a marked span the normal case, not an edge case. A treatment tuned on two chips reads as alarm across a whole card.

`no environment or location field renders a placeholder` correctly tests for an omitted element rather than an empty one or a dash.

The source-level criterion — *"No unit conversion arithmetic appears in the card's source — checkable mechanically alongside the existing no-arithmetic checks"* — **is** met. `tests/card/discipline.spec.ts` scans all of `src/card` with `numericConversions` and carries a negative control, so `Duration.tsx` is covered by the existing check without this PR needing to extend it. I initially read this as a gap because the file predates this PR and so does not appear in its diff; that was my error, not a finding.

### One note on the criterion's rationale, not the work

The criterion justifies excluding persistence controls by saying ADR-0008 *"has no spec and no implementation"*. That is now stale — SPEC-0009 exists and the durable store merged (#111–#114). The criterion's **conclusion** still holds exactly as written: those controls are a story against the durable-store spec, not a loosening of this one. Only the reason given for it has aged.

### Non-blocking, applies to the whole stack

`src/card` has **no mutation coverage**. `scripts/mutation-check.sh` runs 33 mutations across `base.css`, `tokens.css`, the a11y primitives, the shell, the store and `src/canvas` — and none against `src/card`, while #127/#128/#130 add roughly 1,980 lines there. (The `CARD` variable in that script is misleading: it points at `src/canvas/NodeCard.tsx`, not this card.)

None of #97/#98/#99 lists a mutation as an acceptance criterion, so this is not a criterion miss — but in a repo whose central discipline is proving a test can fail, the card is now the largest unguarded surface. The `fixUnsized` and `no completion fraction` tests are the ones I would most want to see turn red on demand, since both assert absences.

PR #131, opened 2026-08-29
Improve mutation test reliability with timeout and syntax fixes

jonstump:
## Summary

This PR improves the reliability and efficiency of the mutation testing script by adding a configurable timeout and fixing a mutation case that was failing for the wrong reason.

## Key Changes

- **Add 8-second timeout to mutation tests**: Replaced the default 30-second Playwright timeout with an 8-second limit. This is generous for the full suite (~54s across two workers) while still catching mutations that cause locators to wait indefinitely. Mutations that break computed values fail immediately, while those breaking element identity cause timeouts—8s is sufficient to distinguish real failures from test infrastructure issues.

- **Fix "landmark goes missing" mutation case**: Changed from renaming only the opening `<footer>` tag to renaming both opening and closing tags. Renaming just the opening tag leaves `</footer>` unmatched, causing a TypeScript syntax error (TS17002) that prevents the page from rendering entirely. This made every locator timeout for the wrong reason (~70s per CI run), defeating the purpose of the mutation test. The fix ensures the mutation fails due to the missing landmark assertion, not compilation failure.

## Implementation Details

- Added comprehensive comment explaining the timeout rationale and the different failure modes of mutations
- Added detailed comment explaining why both tags must be renamed in the landmark mutation case
- Updated the mutation check to use `--timeout=8000` flag with Playwright

https://claude.ai/code/session_014L2HZzNbE55Ft7eJ84n5ee

### DIFF
--- a/web/scripts/mutation-check.sh
+++ b/web/scripts/mutation-check.sh
@@ -73,14 +73,28 @@ check() {
     return
   fi
 
-  # --timeout is far below the 30s default because every run here is expected
-  # to go red, and the two ways a mutation goes red cost very different
-  # amounts. Breaking a computed value — a colour literal, an inset shadow, a
-  # control step — fails its assertion the moment it is evaluated. Breaking an
-  # element's identity leaves a locator waiting for something that will never
-  # appear, at 30s a test. 8s is generous against a full suite that runs in
-  # ~54s across two workers, and a mutation needing longer than that is worth
-  # discovering rather than absorbing.
+  # --timeout is the per-test timeout, below the 30s default because every run
+  # here is expected to go red and waiting is the dominant cost of doing so.
+  #
+  # What it actually bounds: a single waiting assertion is capped by the
+  # *expect* timeout, which is 5s and which this flag does not change —
+  # playwright.config.ts sets neither. The test timeout binds when one test
+  # accumulates several such waits. So a mutation that breaks a computed value
+  # (a colour literal, an inset shadow, a control step) fails the instant the
+  # assertion is evaluated, while one that breaks an element's identity leaves
+  # each dependent assertion waiting out its 5s.
+  #
+  # 8s is chosen against the slowest single *test* in the specs this script
+  # runs, not against the suite's wall time — the suite total is a different
+  # quantity and does not license lowering this further as the suite gets
+  # faster.
+  #
+  # Before lowering it, note what the check cannot see: `check` reads only the
+  # exit code, so a red from a timeout and a red from an assertion are
+  # indistinguishable to it. A timeout tight enough to cut off a slow but
+  # correct test would report that mutation `ok` for the wrong reason — the
+  # same false-pass class the landmark mutation below was an instance of. The
+  # margin here is the safeguard, not slack to reclaim.
   if npx playwright test "$spec" --timeout=8000 --reporter=dot >/dev/null 2>&1; then
     echo "FAIL  $name — the suite still passed, so no test is watching this"
     failures=$((failures + 1))
@@ -213,6 +227,12 @@ check "the live region announces on every render" \
 # reason — it would have gone red with no landmark assertion in the suite at
 # all, which is precisely the thing this script exists to rule out — and it
 # cost ~70s of CI on each of two observed runs while proving nothing.
+#
+# The anchor spans the whole element because `check` performs one substitution,
+# and both tags have to move together. That couples this mutation to the
+# footer's user-facing copy: an editorial change to that sentence breaks the
+# anchor. It breaks loudly — `mutation anchor not found`, then FAIL — so it is
+# a maintenance cost rather than a way to fail open.
 check "a landmark goes missing" \
   tests/shell/shell.spec.ts "$SHELL" \
   '      <footer className="shell-footer">
--- a/web/scripts/mutation-check.sh
+++ b/web/scripts/mutation-check.sh
@@ -73,7 +73,15 @@ check() {
     return
   fi
 
-  if npx playwright test "$spec" --reporter=dot >/dev/null 2>&1; then
+  # --timeout is far below the 30s default because every run here is expected
+  # to go red, and the two ways a mutation goes red cost very different
+  # amounts. Breaking a computed value — a colour literal, an inset shadow, a
+  # control step — fails its assertion the moment it is evaluated. Breaking an
+  # element's identity leaves a locator waiting for something that will never
+  # appear, at 30s a test. 8s is generous against a full suite that runs in
+  # ~54s across two workers, and a mutation needing longer than that is worth
+  # discovering rather than absorbing.
+  if npx playwright test "$spec" --timeout=8000 --reporter=dot >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


jonstump (commented):
Review verdict: **approve** — submitted as a comment because GitHub won't let this account approve a PR it authored.

Both changes are correct, and I verified the landmark fix by reproduction rather than by reading. This is a real bug fix, not a cosmetic one.

## Verified by reproduction

**The old mutation was broken exactly as described.** Applying the opening-tag-only form to a clean tree:

```
src/shell/AppShell.tsx(326,9): error TS17002: Expected corresponding JSX closing tag for 'div'.
```

**The new form fixes it.** With both tags renamed, `npm run typecheck` exits 0 and `tests/shell/shell.spec.ts` fails on the assertion it is meant to fail on:

```
Error: expect(locator).toHaveCount(expected) failed
Locator: getByRole('contentinfo')
  - unexpected value "0"
> 50 |   await expect(page.getByRole("contentinfo")).toHaveCount(1);

1 failed
5 passed (8.3s)
```

**5 of 6 passing is the whole point.** Under the old mutation the page never compiled, so every test in the file went red — meaning the suite would have gone red with no landmark assertion in it at all. Now exactly one test fails, and it is the landmark one.

`npm run test:mutation` passes on this branch: every mutation red, 356s. The multi-line anchor matches `AppShell.tsx` exactly once, and `mutate`'s `str.replace` handles multi-line fine.

## Three notes on the timeout comment

The `--timeout=8000` change is fine. The reasoning written above it isn't quite, and the comment is what the next person will reason from.

**1. The comparison is between two different things.** `--timeout` is *per test*; "a full suite that runs in ~54s across two workers" is wall time for the whole run. What decides whether 8s is safe is the slowest *single* test, not the suite total. As written, the comment invites cutting it further next time the suite gets faster, which wouldn't follow.

**2. "at 30s a test" overstates the old cost for the case described.** A locator wait inside `expect(...).toHaveCount()` is bounded by the **expect timeout** — default 5s, and `playwright.config.ts` overrides neither timeout — not by the 30s test timeout. A single waiting assertion cost ~5s. The 30s only binds when a test stacks several waits or awaits something outside an `expect`. The landmark case was expensive because *every test in the file* failed, not because each waited 30s.

**3. The script cannot tell a timeout-red from an assertion-red.** `check` only inspects the exit code, so "went red" covers both. Lowering the timeout slightly widens the window where a *slow but correct* test times out and the mutation is reported `ok` for the wrong reason — the same false-pass class this PR fixes for the landmark. Not blocking: the specs this script runs are DOM-only and fast, and 356s locally with no spurious reds is decent evidence. But one line noting that the timeout trades cost against exactly the property the script exists to guarantee would help whoever lowers it next.

## One minor coupling

The anchor now embeds user-facing copy — "Figures come from the planner module. Nothing on this page is computed here." An editorial change to that sentence breaks the mutation. It breaks *loudly* (`mutation anchor not found` → `FAIL`), so it's a maintenance cost rather than a correctness risk, and with `check`'s single-substitution signature it's the pragmatic way to keep the tags balanced. Worth knowing it's there.

CI: all 9 checks green, including Frontend tests (completed 18:44:24).

---

jonstump:
All three notes were right, and note 2 was a real correction rather than a wording quibble. Pushed `b63e513` — the flag is unchanged; only the reasoning above it is rewritten, since that comment is what the next person reasons from.

**1 — comparing different quantities.** Fixed. The comment now states 8s against the slowest single *test* in the specs this script runs, and says explicitly that the suite total is a different quantity and does not license lowering the timeout as the suite gets faster.

**2 — "at 30s a test" was wrong.** Verified before changing it: `playwright.config.ts` sets neither `timeout` nor an `expect` block, so the defaults apply — 30s test, **5s expect**. A single waiting assertion is bounded by the expect timeout, which `--timeout` does not touch; the test timeout binds only when one test accumulates several waits. The comment now says that, and attributes the landmark cost to *every test in the file* failing rather than to each waiting 30s. Your reproduction shows it cleanly: 1 failed, 5 passed, 8.3s — a single ~5s wait, not a 30s one.

**3 — the check cannot tell the two reds apart.** Added, and framed as the constraint it is: `check` reads only the exit code, so a timeout-red and an assertion-red are indistinguishable to it. A timeout tight enough to cut off a slow but correct test would report that mutation `ok` for the wrong reason — the same false-pass class the landmark mutation was an instance of. The comment now says the margin is the safeguard, not slack to reclaim.

**On the anchor coupling** — agreed, and recorded rather than changed. The anchor spans the whole element because `check` performs one substitution and both tags have to move together. The comment notes that this embeds the footer's copy and that an editorial change breaks it loudly (`mutation anchor not found` → `FAIL`) rather than failing open. Narrowing it would need `check` to take a list of substitutions, which is a bigger change than this PR should carry.

Verified on the new head: `format:check` clean, all 41 mutations still red.

One correction to my own earlier claim while I'm here. I reported this PR as saving ~148s across the three slow mutations, which holds per-mutation (landmark 70.0s → 10.4s, glyph 66.3s → 22.2s, degraded-layout 63.6s → 19.4s). But the job total only moved 8m30s → 8m11s, because `main` gained 7 mutations and 38 tests between the two runs. The honest figure is per-mutation: 12.0s → 9.2s average. The mutation phase is still 77% of the job, and the largest remaining item is per-invocation server startup at roughly 150s across 41 runs.

---

jonstump:
## Review — approved

CI 9/9, one file, conflict gate clean.

### The landmark fix is the real content, and it is a good catch

Renaming `<footer>` without `</footer>` is a TypeScript syntax error, so `AppShell.tsx` stopped compiling, the page never rendered, and every locator in the spec timed out. The mutation went red — for a reason that had nothing to do with the landmark assertion. As the comment puts it: *"it would have gone red with no landmark assertion in the suite at all, which is precisely the thing this script exists to rule out."*

That is the script's own failure mode turned on itself, and it is exactly the class this project keeps finding. Spanning the whole element is the right fix, and the comment is honest that it couples the anchor to the footer's copy and that the coupling fails loudly rather than open.

### I checked the timeout, because the comment invites it

The rationale names its own hazard — *"a red from a timeout and a red from an assertion are indistinguishable to it"* — which is a claim worth testing rather than admiring. Ran the fifteen spec files the script names, **unmutated**, under the new cap:

```
137 passed (18.8s)  with --timeout=8000
slowest single test: 2.4s        → 3.3x headroom
```

So the margin is real and the cap cannot be silently swallowing a slow-but-correct test. Both figures are on today's suite, which is what the comment says the number is chosen against — and it is right that the suite *total* is a different quantity and not a licence to lower it further.

One thing to note for whoever touches this next: the specs that cap applies to have grown since — `method-control.spec.ts` and `assignment.spec.ts` joined it with #134 and #137. The headroom above is measured with those included, so the number still holds, but it is now the kind of constant that wants re-measuring rather than assuming.

### Verified against the merged state

Rebased onto `main` after #134 and #137 landed — both of which extend the same file — and ran the whole thing:

```
52 mutations, every one turned the suite red
```

Including the repaired landmark mutation, whose anchor spans the footer copy that #134 and #137 both left alone.

Merging.

Issue #132, opened 2026-08-29
Recipe selection: the control the design has not drawn


jonstump:
Part of #84. Governed by [SPEC-0006](docs/openspec/specs/tree-canvas/spec.md), implementing [ADR-0005](docs/adrs/ADR-0005-multiple-recipes-per-output.md).

**Blocked on design, not on code.** Split out of #87, which shipped the Method Selection half.

## Why this is its own story

SPEC-0006 REQ "Recipe Selection" carries a normative constraint on the implementer:

> **The design has not drawn this control.** The prototype's segmented craft|refine control predates ADR-0005 and addresses method only. A surface implementing this requirement MUST carry the design's answer for the recipe control rather than extending the segmented control past the size it was drawn for.

`docs/design/tree-canvas/handoff.md` contains **no mention of "recipe"** — verified, not assumed. It fully specifies the method popover (segmented control, disabled-but-inert with reason, consequence preview, focus trap, live region) and says nothing about recipe alternatives.

design.md's Migration Plan sequences it the same way:

> 4. The method control, which the design has fully specified.
> 5. The recipe control, once the design has answered.
> 6. Leaf assignment, once step 1 has landed.
>
> Steps 4 and 6 are independent of each other; **step 5 is blocked on design rather than on code.**

And its Risks section: *"This should be resolved before the surface is planned into stories, not during implementation."* #87 bundled 4 and 5, so this issue is that resolution.

## Requirements

- [ ] **Recipe Selection** (SPEC-0006) — Where a node's payload carries more than one entry in `legalRecipes`, the canvas MUST offer the alternatives and MUST allow the player to select one. The surface MUST remain usable at the sizes the real data produces. Each alternative MUST be distinguishable by what it consumes and what it yields; a recipe identifier alone MUST NOT be the only distinguishing thing. A node using its engine-default recipe MUST NOT record a selection in plan state.
  - WHEN a node's payload reports 26 legal recipes THEN the canvas offers all 26 and the player may select any of them
  - WHEN two recipes for one output are offered THEN each is shown with what it consumes and how many units it produces
  - WHEN every node in a plan uses its default recipe THEN the plan state carries no recipe selections

## Acceptance Criteria

- The recipe surface is exercised at **61 options**, not at three. Sodium Nitrate carries 26 and the largest cooked output 61; a control whose legibility depends on a small fixed count fails this criterion by construction
- Two recipes for the same output are told apart by inputs and yield, with the identifier not doing that work alone
[...]

PR #134, opened 2026-08-29
feat(web): method selection on the node control

### LINKED ISSUE #87: Method and recipe selection

jonstump:
Closes #87
Part of #84

Implements SPEC-0006 REQ "Method Selection".

> **Scoped to Method Selection.** The Recipe Selection half moved to #132, which is blocked on design rather than on code. `docs/design/tree-canvas/handoff.md` contains **no mention of "recipe"** — verified, not assumed — and SPEC-0006 requires an implementation carry the design's answer rather than stretch the segmented control past the two options it was drawn for. design.md's own Migration Plan already sequences them apart: step 4 (method) is buildable, step 5 (recipe) *"is blocked on design rather than on code"*.
>
> The real data backs that up: **Chromatic Metal carries 25 legal recipes** and Ionised Cobalt 5, measured from the shipped artifact during this work.

## The rule that needed care

*"The options offered MUST be the node's `legalMethods` from the payload; the canvas MUST NOT compute which methods are legal."*

Knowing the vocabulary is raw / craft / refine is the design's fact — it fixes that set and deliberately excludes buy. Which of them apply to a given item is the domain's. `methods.ts` reads only `legalMethods` for that: no terminal-implies-raw shortcut, no inference from the presence of children, no item table.

That distinction is not testable by rendering. **A canvas that computed legality and happened to agree with the payload passes every rendering assertion**, so it is checked two more ways:

- the rendered options are compared against `legalMethods` read out of the captured crossing, not against a fixture
- `methods.ts` is scanned for the shortcuts a computing implementation would need (`terminal`, `children`, `itemId`, `recipe`), and every other canvas source is scanned to confirm none builds a second option list

A method the payload reports and `methods.ts` has never heard of is **still offered**, appended in payload order. Dropping it would be the canvas deciding legality by omission — the same failure from the other direction, and the case a whitelist quietly gets wrong.

## Inert, not hidden

*"A hidden option cannot be distinguished from one that does not exist."* Unavailable methods render `disabled`, and the reason is **text on the page** rather than a `title` attribute — a tooltip is unreachable by keyboard and unread by a screen reader on a disabled control.

The reason states what the payload says (*"The planner reports no refine route for Silver."*) rather than inventing a domain rule the view does not own and would get wrong the first time the data disagreed.

## The consequence line

SPEC-0006 requires the control state the consequence before the change, in the domain's terms. It names the route the payload **actually reports**: current method, recipe yield, and what the node consumes per unit, from its own `children`.

It deliberately does **not** price the alternative. The payload carries no figures for a route the planner did not resolve, so any number there would be the view computing a domain value — the rule this whole surface is built under. The second line names the effect instead, which is true and checkable.

## Where overrides live

Plan state, held beside the crossing rather than in `ViewState` — SPEC-0005 keeps the plan out of view state, and ADR-0002 puts its permanent home in the URL hash. `encodePlanToHash` exists and **nothing calls it**; SPEC-0005 records the decode-on-load path as an open question. Until that lands they sit next to the result cache, and the comment says so.

They are stored **with the target they were chosen against** and derived away when it changes, rather than cleared by an effect. An override keyed by an item id from the previous tree is not an override for this one, and the effect version was a cascading render the linter correctly rejected.

`recomputeWith` rather than `recompute`: `setState` has not re-rendered when the handler runs, so `recompute` would cross with the *previous* overrides and resolve the plan the player just changed away from.

## A test that was wrong first

The change tests originally drove Chromatic Metal from refine to craft. Chromatic Metal is **refine-only** — the terminals are the nodes carrying `raw/refine`. The tests now derive the node and target method from the payload, so they follow the artifact instead of a guess about it.

## Verification

**294 tests pass** (up from 278), 16 new. lint, lint:css, typecheck, format:check, check:tokens, build all clean. **47/47 mutations red**, six new:

| Mutation | Catches |
|---|---|
| the control decides legality instead of reading it | payload-sourced options |
| an unavailable method is hidden rather than rendered inert | present-and-inert |
| an inert option stops saying why | the stated reason |
| the card stops opening its control | click / Enter route |
| a method change stops recomputing through the boundary | recompute, not in-place adjustment |
| the announcement stops naming what changed | the live-region requirement |

Both focus-return routes are tested separately — Escape and the close control — because an Escape-only implementation leaves the other green.

No protected paths touched.

### DIFF
--- a/web/scripts/mutation-check.sh
+++ b/web/scripts/mutation-check.sh
@@ -43,7 +43,10 @@ LAYOUT="src/canvas/layout.ts"
 TREE="src/canvas/TreeCanvas.tsx"
 
 MUTABLE="$BASE $TOKENS $TRAP $LIVE $SHELL $BADGE $STORE $MODEL $EDGE $CARD $CANVAS"
-MUTABLE="$MUTABLE $LAYOUT $TREE"
+METHODS="src/canvas/methods.ts"
+CONTROL="src/canvas/NodeControl.tsx"
+
+MUTABLE="$MUTABLE $LAYOUT $TREE $METHODS $CONTROL"
 
 # shellcheck disable=SC2086
 restore() { git checkout -- $MUTABLE 2>/dev/null || true; }
@@ -449,6 +452,46 @@ check "something other than base identity writes to the card border" \
   ".identity {
   border: var(--border-width) solid var(--border);
 }"
+
+# ----------------------------------------------------------------------
+# The node's method control.
+#
+# SPEC-0006 REQ "Method Selection" is two claims that a behavioural test
+# cannot separate on its own: the options come from the payload, and an
+# unavailable one is present-and-inert rather than absent. A canvas that
+# computed legality and happened to agree with the payload passes every
+# rendering assertion, so the first mutation makes it disagree.
+# ----------------------------------------------------------------------
+
+check "the control decides legality instead of reading it" \
+  tests/canvas/method-control.spec.ts "$METHODS" \
+  "    const available = legal.has(method);" \
+  "    const available = true;"
+
+check "an unavailable method is hidden rather than rendered inert" \
+  tests/canvas/method-control.spec.ts "$CONTROL" \
+  "          {options.map((option) => (" \
+  "          {options.filter((shown) => shown.available).map((option) => ("
+
+check "an inert option stops saying why" \
+  tests/canvas/method-control.spec.ts "$CONTROL" \
+  "          .filter((option) => option.reason !== null)" \
+  "          .filter(() => false)"
+
+check "the card stops opening its control" \
+  tests/canvas/method-control.spec.ts "$CARD" \
+  "          onOpen(node.id);" \
+  "          void node.id;"
+
+check "a method change stops recomputing through the boundary" \
+  tests/canvas/method-control.spec.ts "$SHELL" \
+  "      recomputeWith(next);" \
+  "      void next;"
+
+check "the announcement stops naming what changed" \
+  tests/canvas/method-control.spec.ts "$SHELL" \
+  "    const change = pendingChange.current;" \
+  "    const change: string | null = null;"
 
 echo
 if [ "$failures" -gt 0 ]; then
--- a/web/src/canvas/NodeCard.tsx
+++ b/web/src/canvas/NodeCard.tsx
@@ -58,6 +58,15 @@ export interface NodeCardData extends Record<string, unknown> {
   readonly display: string;
   /** The recipe's yield, when it is not 1. Absent otherwise. */
   readonly yieldDisplay: string | null;
+  /**
+   * Open this node's control.
+   *
+   * SPEC-0006 REQ "Method Selection": "Clicking or pressing Enter on a node
+   * MUST open a control". The card is a `<button>`, so Enter and Space
+   * already arrive here as clicks — there is no key handler, which is what
+   * keeps the pointer and keyboard routes from being able to disagree.
+   */
+  readonly onOpen: (nodeId: string) => void;
   /** The application count, exact and unrounded. Absent for a terminal. */
   readonly applicationsDisplay: string | null;
   /**
@@ -113,7 +122,7 @@ function Figure({
 }
 
 export function NodeCard({ data }: NodeProps): ReactNode {
-  const { node, display, yieldDisplay, applicationsDisplay, identity } =
+  const { node, display, yieldDisplay, applicationsDisplay, identity, onOpen } =
     data as unknown as NodeCardData;
 
   /*
@@ -149,6 +158,10 @@ export function NodeCard({ data }: NodeProps): ReactNode {
         type="button"
         className={`node-card interactive selectable ${frame}`}
         data-method={node.method}
+        aria-haspopup="dialog"
+        onClick={() => {
+          onOpen(node.id);
+        }}
         /* Only a leaf can be assigned, so only a leaf carries the attribute.
          * A non-leaf marked "unassigned" would be describing a state it
          * cannot be in. */
--- a/web/src/canvas/NodeControl.tsx
+++ b/web/src/canvas/NodeControl.tsx
@@ -1,0 +1,133 @@
+import type { ReactNode } from "react";
+
+import type { Quantity } from "../boundary";
+import type { CanvasNode } from "./graph-model";
+import { methodOptions } from "./methods";
+
+/*
+ * The node's control: how this item is made.
+ *
+ * Governing: ADR-0004 (React view layer), ADR-0005 (multiple recipes per
+ * output), SPEC-0006 REQ "Method Selection", Accessibility Requirements →
+ * Keyboard Navigation, Focus Management
+ *
+ * Method only. The recipe half of SPEC-0006's control is issue #132 and is
+ * blocked on design rather than on code: the handoff contains no mention of
+ * "recipe", the spec requires an implementation carry the design's answer,
+ * and the segmented control below is drawn for two options against data
+ * that reaches sixty-one. Stretching it here is the thing the spec names.
+ *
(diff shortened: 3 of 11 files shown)


jonstump (commented):
## Review: SPEC-0006 REQ "Method Selection" — would approve

**Verdict: approve.** One non-blocking defect and one piece of new information. Submitted as a comment rather than an approval because GitHub refuses `APPROVE` on your own pull request and the review token authenticates as you.

CI: all 9 checks green on `f29e1f7`. Conflict-marker gate: clean. Merges cleanly against current `main` (which has moved four commits since this branch was cut, all documentation).

### What I verified rather than read

**Legality comes from the payload.** `methods.ts` consults `legalMethods` and nothing else — no item table, no terminal-implies-raw shortcut, no inference from children. `graph-model.ts` carries `legalMethods` through verbatim. The `METHOD_ORDER` distinction the file argues for holds up: knowing the vocabulary is raw/craft/refine is not deciding which applies, and a method the payload reports that the file has never heard of is still appended rather than dropped. The mutation that flips `available` to `true` is the right test for it.

**Unavailable is present and inert, with the reason discoverable.** `disabled` rather than hidden, and the reason rendered as a visible `<p>` rather than a `title` — correct, since a tooltip on a disabled control is reachable by neither keyboard nor screen reader.

**No view arithmetic.** `inputs` are built from the node's own `children` with `perUnit` carried as `Quantity`, and every figure goes through `format`. SPEC-0005 REQ "The View Computes No Domain Values" holds.

**Keyboard.** The card is a real `<button>`, so Enter and Space arrive as clicks with no key handler to disagree with the pointer route, and `Popover` owns the trap rather than this control reimplementing it — which is what issue #87 asked for.

**Mutation coverage grew.** Six new mutations, none of the existing ones relaxed.

### Defect: a stale announcement survives a no-op selection

`pendingChange.current` is cleared only inside the `useAnnounceOnChange` callback, and that callback does not run when the token is unchanged:

```js
if (first || !changed || token === null) return;
```

Clicking the already-selected method button reaches `onSelectMethod` — only *unavailable* options are `disabled`, so the current one is live — and produces a `next` map with identical contents. `requestKey` is therefore the same string, `run()` takes the cached branch, `setResultToken` sets a value equal to the current one, the effect does not re-run, and `pendingChange.current` keeps its string.

The next genuine recompute then inherits it. Change the quantity, press Recompute, and the live region says `Copper set to craft. Totals updated…` for a change made several interactions ago.

Repro: open a node's control, click its current method twice, close, change the quantity, Recompute, read the live region. Traced through the source rather than reproduced in a browser — I did not stand up the suite for it.

Three shapes of fix, your pick: clear `pendingChange` when `run()` takes the cached path; skip early in `onSelectMethod` when `methods[nodeId] === method`; or key the announcement on a counter that increments per accepted change rather than on `resultToken`.

### New information: the hash question has been answered since you branched

`usePlanResolution.ts` and `AppShell.tsx` both cite ADR-0002 for the overrides' permanent home and call the hash path "an open question" per SPEC-0005. Both were true when this branch was cut. Since then ADR-0010 was accepted and SPEC-0011 merged, and SPEC-0011 REQ "The Hash Owns the Plan, the Store Owns the Player" names the content explicitly: "target, quantity, method and recipe selections, and leaf assignments."

The interim placement is unaffected — holding overrides beside the crossing until the hash path is wired is still right, and the derivation-on-target-change is a better shape than the effect it avoids. What is stale is the citation: the comments point at a question that now has an answer. Worth repointing at SPEC-0011 while the reasoning is fresh, since these comments are the thing a future session will read.

### Not findings, having checked

The `JSON.stringify` cache key is order-sensitive across the override map, but the map is only ever extended by spread within one target and drops entirely when the target changes, so two orderings of the same logical map are not reachable. No extra crossings.

---

jonstump:
## Review — approved

Conflict gate clean, CI 9/9. Verified against SPEC-0006 REQ "Method Selection" and issue #87's criteria.

**The scope split is right, and checkable.** `docs/design/tree-canvas/handoff.md` contains no occurrence of "recipe" — I grepped it rather than taking the claim — and SPEC-0006 REQ "Recipe Selection" says outright that "the design has not drawn this control" and that an implementation "MUST carry the design's answer rather than extending the segmented control past the size it was drawn for". #132 exists and is labelled `blocked`. Splitting there is what the spec asks for, not a shortcut.

**The distinction the PR builds its tests around is the correct one.** *"A canvas that computed legality and happened to agree with the payload passes every rendering assertion"* — that is exactly the failure mode, and comparing rendered options against `legalMethods` read from the captured crossing is the only thing that separates the two. Offering an unrecognised method rather than dropping it is the same argument from the other side, and it is the half a whitelist usually gets wrong.

**Stating the reason as text rather than `title`** is right for a disabled control, where a tooltip is unreachable by keyboard and unread by AT.

Two claims I checked rather than accepted:

- `encodePlanToHash` is exported from `boundary/index.ts` and defined in `plan-hash.ts`; every other occurrence in `web/src` is inside a comment. Nothing calls it. Claim holds.
- The overrides being keyed to the target they were chosen against, and derived away rather than cleared by an effect — this is the same StrictMode hazard #137 hits from the other direction, and both landed on the handler rather than the effect. Consistent, and the right answer.

Verified on an integration branch carrying this PR, #137 and #131 rebased onto current main: **307 tests pass, 52/52 mutations red**, typecheck / lint / lint:css / format:check / check:tokens / build all clean.

**Worth knowing before merge:** this branch was 7 commits behind `main` while GitHub reported CLEAN — the doc PRs (#136–#142) landed underneath it. Merging as-is is fine; the rebase was clean and verified above.

Merging.

PR #136, opened 2026-08-29
ci: skip the frontend suite on documentation-only pull requests

jonstump:
## Summary

Frontend tests is the long pole in CI — 8–11 minutes against 15–40s for every other job, because it runs the 41-mutation check after the suite. #133 spent **11m00s** validating three one-line frontmatter changes under `docs/`.

A `changes` job diffs the PR against its base and sets an output. `web-test` gains `needs: changes` and an `if`.

## Why a job gate and not `paths-ignore`

This is the part worth reviewing. A job skipped by `if` still reports a conclusion, and a required status check accepts that. A workflow that never triggers reports nothing at all — so a required check stays pending and the PR cannot merge.

**If `Frontend tests` is a required check on this repo, a workflow-level `paths-ignore` would have deadlocked every docs-only PR permanently.** I couldn't read the branch-protection settings to confirm one way or the other, so the change is built to be correct under either.

## Two conservative choices

- **Push to `main` always runs everything.** A push has no reliable base to diff against — force-pushes and new branches leave `github.event.before` unusable — and `main`'s green is the load-bearing one.
- **The predicate asks whether anything falls _outside_ the documentation set**, so a path nobody anticipated counts as code and runs the suite. Adding a directory to that set is a deliberate decision to stop testing it.

`.github/` is deliberately **not** in the documentation set: a change to this workflow runs the full suite, so CI cannot switch off its own testing.

Documentation set: `docs/**`, root-level `*.md`, `LICENSE`. Markdown under `web/` is **not** included — it sits beside the code, and being wrong there costs more than the 8 minutes saved.

## Verification

The filter predicate was exercised against 11 cases, including the PRs currently in flight:

| Changed | Result |
|---|---|
| `docs/adrs/ADR-0009-*.md` (#126) | skip |
| `docs/openspec/specs/*/spec.md` (#133) | skip |
| `README.md`, `CLAUDE.md`, `LICENSE` | skip |
| `web/src/card/BasePlannerCard.tsx` | run |
| `web/scripts/mutation-check.sh` | run |
| `internal/rollup/rollup.go` | run |
| docs + code mixed | run |
| `.github/workflows/ci.yml` | run |
| `web/README.md` | run |

All 11 behave as intended. The workflow also parses as YAML.

## What this PR cannot show

It edits `.github/`, so it correctly runs the full suite — the skip path is not exercised here. The skip becomes observable on the first docs-only PR opened after this merges; #133 is the natural candidate if it's rebased.

The `changes` job itself adds one short job (checkout at `fetch-depth: 0`) to every run. On a repo this size that's seconds against 8–11 minutes saved on docs-only PRs.

## Not included

Only `web-test` is gated. `web-build`, `web-lint`, `web-typecheck` and `web-format` also cannot be affected by a docs-only change, but they finish in 15–40s, so gating them trades real risk for little time. Easy to extend later if wanted — they would reuse the same `changes` output.

---
_Generated by [Claude Code](https://claude.ai/code/session_014L2HZzNbE55Ft7eJ84n5ee)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,9 +141,61 @@ jobs:
         # literal and stylelint does not look.
         run: npm run check:tokens
 
+  changes:
+    name: Detect changed paths
+    runs-on: ubuntu-latest
+    outputs:
+      code: ${{ steps.filter.outputs.code }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # The diff below needs the base commit present, not just the tip.
+          fetch-depth: 0
+      - id: filter
+        env:
+          # Empty on a push event; the branch below keys off that.
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+        run: |
+          set -euo pipefail
+
+          # Push to main always runs everything. A push has no single obvious
+          # base to diff against — a force-push or a new branch leaves
+          # github.event.before unusable — and main is the branch whose green
+          # is load-bearing, so it is not the place to economise.
+          if [ -z "${BASE_SHA:-}" ]; then
+            echo "code=true" >> "$GITHUB_OUTPUT"
+            echo "push event — running everything"
+            exit 0
+          fi
+
+          changed=$(git diff --name-only "$BASE_SHA" HEAD)
+          echo "changed files:"
+          echo "$changed" | sed 's/^/  /'
+
+          # `code` is true when ANY changed file falls outside the documentation
+          # set. The test is deliberately this way round: a path nobody thought
+          # about counts as code and runs the suite. Adding a directory here is
+          # a decision to stop testing changes to it.
+          if printf '%s\n' "$changed" | grep -qvE '^(docs/|[^/]+\.md$|LICENSE$)'; then
+            echo "code=true" >> "$GITHUB_OUTPUT"
+            echo "-> code changed, running the frontend suite"
+          else
+            echo "code=false" >> "$GITHUB_OUTPUT"
+            echo "-> documentation only, skipping the frontend suite"
+          fi
+
   web-test:
     name: Frontend tests
     runs-on: ubuntu-latest
+    # The frontend suite plus its mutation check is the long pole in CI —
+    # 8-11 minutes against 15-40s for every other job — and a change confined
+    # to docs/ cannot alter its outcome. Gated on `changes` rather than a
+    # workflow-level paths-ignore: a job skipped by `if` still reports a
+    # conclusion, which a required status check accepts, whereas a workflow
+    # that never triggers leaves that check pending and the pull request
+    # unmergeable.
+    needs: changes
+    if: needs.changes.outputs.code == 'true'
     defaults:
       run:
         working-directory: web
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,7 +195,18 @@ jobs:
     # that never triggers leaves that check pending and the pull request
     # unmergeable.
     needs: changes
-    if: needs.changes.outputs.code == 'true'
+    #
+    # The condition is "not explicitly false" rather than "is true", and it uses
+    # a status-check function so it overrides the default skip-on-needs-failure.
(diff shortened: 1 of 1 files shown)


jonstump (commented):
## Review: docs-only frontend skip — would approve, no defects found

**Verdict: approve.** Submitted as a comment rather than an approval because GitHub refuses `APPROVE` on your own pull request.

CI: all 10 checks green on `55b3b2b`, including the new `Detect changed paths` job reporting on itself. Conflict-marker gate: clean. Merges cleanly against current `main`.

### What I checked

**The test is the right way round.** `grep -qvE` means a path nobody thought about counts as code and runs the suite; only `docs/`, root-level `*.md` and `LICENSE` opt out. Adding a directory to that set is a deliberate decision to stop testing it, which is what the comment says and what the code does.

**The empty-diff case is safe.** An empty `changed` becomes a single blank line, which fails to match the pattern, so `grep -v` matches it and `code=true`. It fails toward running.

**Push events run everything**, and the reasoning for not diffing `github.event.before` is right — a force-push or a new branch leaves it unusable.

**`fetch-depth: 0` is load-bearing** and present: the merge ref's history reaches `base.sha`, so the diff resolves. If the base branch advances mid-flight the diff widens rather than narrows, which again fails toward running.

**Job-level `if` rather than workflow-level `paths-ignore`** — the comment's reasoning is correct and worth keeping. A skipped job reports a conclusion that a required check accepts; a workflow that never triggers leaves that check pending forever and the PR unmergeable. This is the failure mode people hit with `paths-ignore` and it is nice to see it written down rather than rediscovered.

### Two notes, neither blocking

**Branch protection should require `Detect changed paths` too.** A skipped required check counts as passing, so if `Frontend tests` is required and the gate job fails, the skip is indistinguishable from a pass on that check alone — the failing gate job is the only thing standing between a misdetection and an unrun suite. That is fine as long as the gate is itself required. I cannot read protection settings from here, so this is a "check the setting", not a claim that it is wrong.

**The pattern is narrower than "documentation".** A docs PR that also touches `.gitignore` runs the full suite — #140 would have, for exactly that reason. Conservative in the safe direction, just worth knowing the skip fires less often than the name suggests.

Scope is deliberate and I am not asking for more: only `web-test` is gated, so lint, typecheck, build and formatting still run on docs-only PRs at 15–40s each. That is the stated trade and it is the right one — those jobs are cheap and catch a stray file.

---

jonstump:
Note 1 turned out to be fixable here rather than in a setting, so I pushed `22f42f5`.

**The hole was real and worse than "check the setting".** With `if: needs.changes.outputs.code == 'true'`, a *failing* gate job skips `web-test` through `needs`. A skipped required check counts as passing — so a broken detector was indistinguishable from a green suite. Silent, which is the worst shape.

```yaml
if: ${{ !cancelled() && needs.changes.outputs.code != 'false' }}
```

Two changes, both load-bearing. "Not explicitly false" rather than "is true", so an empty output from a failed gate runs the suite. And a status-check function, which overrides the default skip-on-needs-failure so the job is reachable at all when the gate failed — without it the first change would have no effect, because `needs` would skip the job before the condition mattered.

| Case | Outcome |
|---|---|
| gate ok, docs-only (`code=false`) | skip |
| gate ok, code changed (`code=true`) | run |
| **gate fails** (`code=''`) | **run** |
| workflow cancelled | skip |

This is the same direction the predicate already fails in — a path nobody anticipated runs the suite, and now so does a detector that broke. The gate's safety no longer depends on branch protection also requiring the gate job. Requiring it is still worth doing for visibility, but it is no longer the only thing standing between a misdetection and an unrun suite.

**Note 2 needs no change and I have not made one.** A docs PR that also touches `.gitignore` runs the full suite — you are right that #140 would have. That is conservative in the safe direction and I would rather the skip fire less often than the name suggests than widen the documentation set. Worth knowing, as you said; not worth fixing.

On the empty-diff case you checked: agreed, and it is the same mechanism — a single blank line fails to match, `grep -v` matches it, `code=true`. It fails toward running.

CI is re-running on the new head.

---

Issue #135, opened 2026-08-29
Shareable state: what a link carries beyond the plan


jonstump:
Split out of #88. **Blocked on a spec decision, not on code.**

## The gap

SPEC-0006 REQ "Leaf Assignment to Bases" says an assignment "MUST be plan state carried as SPEC-0002 encodes it, not view-local state", and #88's acceptance criteria added that it "survives a URL-hash round trip".

Verified against the code rather than assumed:

| | |
|---|---|
| `Plan` (TS and Go wire) | `{ target, quantity, methods, recipes }` — **no assignments** |
| `RollupRequest` | carries `assignments` as a **sibling of** `plan`, not inside it |
| SPEC-0002 | eleven requirements, **none** covering assignment encoding |
| `encodePlanToHash` | serialises `Plan` only |

So an assignment cannot survive a URL-hash round trip today, and nothing in SPEC-0002 says it should.

## Two readings, and which one #88 took

The requirement admits a narrower reading: *"carried as SPEC-0002 encodes it"* can mean **travels to the domain through the documented boundary contract**, which `RollupRequest.assignments` already satisfies under SPEC-0002 REQ "Boundary Surface". #88 built to that reading — the control assigns, and every change recomputes through `client.rollup()`.

What #88 did **not** do is put assignments in the hash, because `Plan` is what the hash carries and extending it would mean amending SPEC-0002 and the Go wire.

## The wider question this actually raises

Assignments are not the only shareable state that is not in the plan. **SPEC-0007's site and power configuration has the identical shape** — `useConfiguredBase` holds `CardConfiguration` in component state and issues `rollup`/`power` on every change, and none of it reaches a link either. Nobody raised it because SPEC-0007 asks for no hash round trip.

So the decision worth taking once, rather than per-story:

- **Does a shared link carry only the plan** (target, quantity, methods, recipes), leaving assignments and site configuration as per-session state?
- **Or is there a wider "shareable state"** — plan plus assignments plus sites — with its own encoding, of which `Plan` is one part?

The second reading is what a player would expect from "share my base plan", and it is the one that makes SPEC-0006's requirement true as written. It is also a boundary change touching a protected spec, which is why it is not being decided inside a story.

## Also unresolved: the decode-on-load path

`encodePlanToHash` and `decodePlanFromHash` both exist and **nothing in the application calls either**. SPEC-0005 already records this as an open question. Whatever this issue decides about *what* a link carries, something still has to decide *where* reading and writing it lives.

[...]

PR #137, opened 2026-08-29
feat(web): leaf assignment to bases

### LINKED ISSUE #88: Leaf assignment to bases

jonstump:
Closes #88
Part of #84

Implements SPEC-0006 REQ "Leaf Assignment to Bases".

> **Stacked on #134.** Base is `feat/87-method-and-recipe-selection`, not `main` — the assignment control lives in the node popover #87 builds, exactly as the design handoff describes ("base assignment fully operable via the popover select"). **Merge #134 first**, then this retargets cleanly.

## Two gaps found and split rather than papered over

**The URL-hash round trip is #135.** #88's criterion said an assignment "appears in plan state as SPEC-0002 encodes it, and survives a URL-hash round trip". Verified against the code:

| | |
|---|---|
| `Plan` (TS and Go wire) | `{target, quantity, methods, recipes}` — **no assignments** |
| `RollupRequest` | carries `assignments` as a **sibling of** `plan` |
| SPEC-0002 | eleven requirements, **none** covering assignment encoding |
| `encodePlanToHash` | serialises `Plan` only |

This builds to the narrower reading — assignments travel to the domain through the documented boundary contract, which `RollupRequest.assignments` satisfies. Nothing here claims a link survives them. #135 carries the wider question, and notes **SPEC-0007's site configuration has the identical gap**.

**The application has no curated constants.** `RollupRequest` requires them; they exist only in test fixtures, and the base planner card that would own them is not mounted. So `constants: null` in the shell is the honest shape, not a stub: assignments are held and their colours rendered, and the stage-2 dispatch is proven in a fixture that supplies real constants — the same shape every SPEC-0007 card story ships in.

## A bug the tests found

The first version dispatched from an effect on `assignments`, guarded with "skip when the map is empty" to avoid a mount dispatch. That guard **suppressed the clear** — and clearing is a real change: a leaf no longer gathered at a base changes that base's totals exactly as reassigning it does.

The dispatch now happens in the handler. That also removes a StrictMode hazard: an effect cannot tell the first render from a real change without a flag, and StrictMode double-invokes both the effect and the flag. It is mutation #1 below.

## Where each claim is checked

Deliberately not all in one place, because the three claims are observable in different places:

- **keyboard-operable end to end** — the real application, driven with keys from the card through Enter into the select, not by calling a handler
- **recomputes through the boundary** — the fixture, because a shell that cannot dispatch cannot demonstrate a dispatch
- **the canvas adjusts nothing itself** — the source, since a canvas that recomputed *and* adjusted passes both of the above

A base id that no longer maps to a slot reads as unassigned rather than as an uncoloured border — the case a link or an older session produces.

## Verification

**307 tests pass** (up from 294), 13 new. lint, lint:css, typecheck, format:check, check:tokens, build all clean. **52/52 mutations red**, five new:

| Mutation | Catches |
|---|---|
| **clearing an assignment stops recomputing** | the shipped bug, above |
| the rollup goes out without the assignment on it | what actually crosses |
| a non-leaf is offered a base it cannot be gathered at | leaf-only control |
| an assigned leaf stops taking its base colour | the border requirement |
| the assignment announcement stops naming the base | the live-region requirement |

No protected paths touched.

### DIFF
--- a/web/scripts/mutation-check.sh
+++ b/web/scripts/mutation-check.sh
@@ -46,7 +46,10 @@ MUTABLE="$BASE $TOKENS $TRAP $LIVE $SHELL $BADGE $STORE $MODEL $EDGE $CARD $CANV
 METHODS="src/canvas/methods.ts"
 CONTROL="src/canvas/NodeControl.tsx"
 
-MUTABLE="$MUTABLE $LAYOUT $TREE $METHODS $CONTROL"
+ASSIGN="src/canvas/useLeafAssignment.ts"
+BASES="src/canvas/bases.ts"
+
+MUTABLE="$MUTABLE $LAYOUT $TREE $METHODS $CONTROL $ASSIGN $BASES"
 
 # shellcheck disable=SC2086
 restore() { git checkout -- $MUTABLE 2>/dev/null || true; }
@@ -492,6 +495,40 @@ check "the announcement stops naming what changed" \
   tests/canvas/method-control.spec.ts "$SHELL" \
   "    const change = pendingChange.current;" \
   "    const change: string | null = null;"
+
+# ----------------------------------------------------------------------
+# Leaf assignment.
+#
+# The first mutation is the bug this story actually shipped and a test
+# caught: skipping the dispatch when the map goes empty suppresses the
+# *clear*, which is a real change — a leaf no longer gathered at a base
+# changes that base's totals exactly as reassigning it does.
+# ----------------------------------------------------------------------
+
+check "clearing an assignment stops recomputing" \
+  tests/canvas/assignment.spec.ts "$ASSIGN" \
+  "    if (constants === null) return;" \
+  "    if (constants === null || Object.keys(next).length === 0) return;"
+
+check "the rollup goes out without the assignment on it" \
+  tests/canvas/assignment.spec.ts "$ASSIGN" \
+  "    const request: RollupRequest = { plan, assignments: next, constants };" \
+  "    const request: RollupRequest = { plan, assignments: {}, constants };"
+
+check "a non-leaf is offered a base it cannot be gathered at" \
+  tests/canvas/assignment.spec.ts "$CONTROL" \
+  "      {node.terminal && (" \
+  "      {(node.terminal || true) && ("
+
+check "an assigned leaf stops taking its base colour" \
+  tests/canvas/assignment.spec.ts "$BASES" \
+  "  return BASES.find((base) => base.id === id)?.slot;" \
+  "  return undefined;"
+
+check "the assignment announcement stops naming the base" \
+  tests/canvas/assignment.spec.ts "$SHELL" \
+  "      const label = BASES.find((base) => base.id === baseId)?.label;" \
+  "      const label: string | undefined = undefined;"
 
 echo
 if [ "$failures" -gt 0 ]; then
--- a/web/src/canvas/NodeControl.tsx
+++ b/web/src/canvas/NodeControl.tsx
@@ -1,6 +1,7 @@
 import type { ReactNode } from "react";
 
 import type { Quantity } from "../boundary";
+import { BASES } from "./bases";
 import type { CanvasNode } from "./graph-model";
 import { methodOptions } from "./methods";
 
@@ -30,6 +31,14 @@ export interface NodeControlProps {
   readonly node: CanvasNode;
   readonly headingId: string;
   readonly onSelectMethod: (method: string) => void;
+  /**
+   * Assign this leaf to a base, or clear it with null.
+   *
+   * Only a leaf can be assigned, so the control only renders for one.
+   */
+  readonly onAssign: (baseId: string | null) => void;
+  /** The base this leaf is currently assigned to, if any. */
+  readonly assignedTo: string | null;
   /** Grouping only — `formatQuantity` never parses. */
   readonly format: (quantity: Quantity) => string;
 }
@@ -38,6 +47,8 @@ export function NodeControl({
   node,
   headingId,
   onSelectMethod,
+  onAssign,
+  assignedTo,
   format,
 }: NodeControlProps): ReactNode {
   const options = methodOptions(node.legalMethods, node.method, node.name);
@@ -124,9 +135,44 @@ export function NodeControl({
           ))}
       </fieldset>
 
+      {/*
+        Only a leaf is assignable. A non-leaf is produced by the steps below
+        it rather than gathered at a base, and offering the control on one
+        would be describing a state it cannot be in — the same reason the
+        card only carries `data-identity` on a leaf.
+
+        A `<select>` because the design asks for one and because it is
+        operable by keyboard without anything added: SPEC-0006 requires the
+        assignment "be changed and committed without a pointing device", and
+        a bespoke listbox would have to re-earn that.
+      */}
+      {node.terminal && (
+        <p className="node-control-base">
+          <label className="label" htmlFor={`${headingId}-base`}>
+            Gathered at
+          </label>{" "}
+          <select
+            id={`${headingId}-base`}
+            className="control control-sm interactive node-base-select"
+            value={assignedTo ?? ""}
+            onChange={(event) => {
+              onAssign(event.target.value === "" ? null : event.target.value);
+            }}
+          >
+            <option value="">Unassigned</option>
+            {BASES.map((base) => (
+              <option key={base.id} value={base.id}>
+                {base.label}
+              </option>
+            ))}
+          </select>
+        </p>
+      )}
+
       <p className="label node-control-effect">
-        Changing the method re-resolves the plan. Every total below this node is
-        recomputed by the planner.
+        {node.terminal
+          ? "Changing the method or the base recomputes through the planner. No figure is adjusted here."
+          : "Changing the method re-resolves the plan. Every total below this node is recomputed by the planner."}
       </p>
     </div>
   );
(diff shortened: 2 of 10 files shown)


jonstump (commented):
## Review: SPEC-0006 REQ "Leaf Assignment to Bases" — would approve, with one thing that must not merge quietly

**Verdict: approve the code; the identifier question below needs an answer before or with the merge.** Submitted as a comment rather than an approval because GitHub refuses `APPROVE` on your own pull request.

CI: all 9 checks green on `c8edf9f`. Conflict-marker gate: clean. Merges cleanly onto #134's branch, which merges cleanly onto current `main`.

### What I verified rather than read

**Keyboard operability is native, not re-earned.** A `<select>` rather than a bespoke listbox, so SPEC-0006's "changed and committed without a pointing device" comes free. `onChange` also means selecting the already-assigned base fires nothing — the no-op announcement I went looking for is not reachable here.

**No view arithmetic and no boundary method of its own.** `slotFor` is a lookup over `BASES`; `useLeafAssignment` calls `client.rollup()` and adds nothing to the client, which is what #88 asks and what #95 exists to protect.

**The clear is a real change and is dispatched as one.** Rebuilding the map without the key rather than writing `undefined` is right — `{ COBALT: undefined }` would serialise to a field the domain has no contract for. The mutation you added for the suppressed clear is the good kind: it encodes a bug that actually happened, and the comment says so instead of hiding it.

**Superseded replies are dropped by sequence number**, consistent with `usePlanResolution` and `useConfiguredBase` rather than inventing a third convention.

### The production path never dispatches

`AppShell` passes `constants: null`, and `assign` returns before the rollup when constants is null. In the running application, assigning a leaf updates a local map and a border colour and never reaches the boundary. SPEC-0006's "Reassignment recomputes through the boundary" is proven in a fixture, not shipped.

This is disclosed in three places and the announcement deliberately declines to claim totals updated, which is the honest call and better than a stub that lies. I am naming it anyway because `Closes #88` will read as though the acceptance criterion shipped. Either the PR body or the issue should carry the remaining wiring — a curated-constants source — as explicit follow-up, so it is not rediscovered later from the fixture.

Two smaller things in the same vein, both cheap now and awkward once constants land:

- **A failed rollup is silent.** `client.rollup()` resolves to `Outcome<Build>` and never rejects, so `.then(() => …)` counts a failure as a dispatch and surfaces nothing. `usePlanResolution` branches on `outcome.kind`; this will want the same.
- **`plan` falls back to `EMPTY_PLAN`** when the inputs do not validate, so an assignment made against an unusable target would dispatch a rollup for an empty plan. Worth skipping the dispatch when validation failed.

### `bases.ts` mints identifiers a decision now forbids

`BASES` mints `base-1` … `base-6`, and the file says a base's real name "comes from the bases map, which has no spec and no surface yet."

That was accurate when you branched. It is not now. ADR-0010 §1 has been accepted and SPEC-0011 merged, and SPEC-0011 REQ "A Place Is Authored, and a Plan References It" is direct about it:

> `BaseID` MUST be the place record's `id` … The application MUST NOT mint a second identifier for a place, and MUST NOT derive a place's identity from a plan's assignments.

Six synthetic ids are precisely the second identifier that rule forbids, and SPEC-0009 is implemented, so the record they should be is already there. SPEC-0011 REQ "An Assignment Naming an Absent Place Is Unassigned" also now governs what this hook does when an assigned place is deleted — which is nothing, today.

I am not asking you to rewrite this PR for a decision that post-dates it, and the placeholder is labelled honestly. But it should not land carrying a comment that says the answer does not exist, because that comment is what the next session will believe. Minimum before merge: repoint `bases.ts`'s comment at ADR-0010 §1 and SPEC-0011, say the six slots are a pre-places placeholder, and link the follow-up that replaces them with place ids.

---

jonstump:
## Review — approved, rebased off the stack

Conflict gate clean, CI 10/10 after the rebase. Verified against SPEC-0006 REQ "Leaf Assignment to Bases" and issue #88's criteria.

### The rebase

The PR body predicted this exactly — *"Merge #134 first, then this retargets cleanly."* Half happened: GitHub auto-retargeted the base to `main` when #134's branch went away, and retargeting moves the base without rebasing the commits, so it came back **DIRTY** at +1806/-54 across 16 files. `git rebase --onto origin/main f29e1f7` drops the duplicated #134 commit and leaves one: **+826/-6 across 10 files**, which is the real change. Now CLEAN.

Worth recording as a pattern, since this is the second time this session: on a stacked PR, `mergeable: CLEAN` is a statement about the *old* base. Check `baseRefName` and the diff size, not just the flag.

Verified the rebased tree rather than trusting a clean rebase — 307 tests, all gates green — and separately on an integration branch carrying #134, #137 and #131 together: **52/52 mutations red**.

### The bug the tests found is the best thing here

> The first version dispatched from an effect on `assignments`, guarded with "skip when the map is empty" to avoid a mount dispatch. That guard **suppressed the clear** — and clearing is a real change.

That is the whole class in one sentence: a guard written for one case that silently swallows another. Mutation #1 (*"clearing an assignment stops recomputing"*) keeps it, and moving the dispatch into the handler also removes the StrictMode hazard, which an effect-plus-flag cannot survive because StrictMode double-invokes both. #134 landed on the handler for the same reason independently — two stories, same conclusion.

### The three-way split of where each claim is checked

Keyboard operability in the real application, the boundary dispatch in a fixture, and "the canvas adjusts nothing itself" in the source. The reasoning given is right: *"a canvas that recomputed **and** adjusted passes both of the above."* Absence is not observable by driving.

### Verified in the running application

Not just in tests. Resolved ANTIMATTER, opened Indium's control, assigned it:

```
dialog: "Indium · Now raw · Method raw craft refine ·
         The planner reports no craft route for Indium.
         The planner reports no refine route for Indium.
         Gathered at: Unassigned, Base 1…6"
after:  data-identity="3"   border: 3px solid rgb(232, 84, 58)   ← --base-3
```

The inert options state their reason as text, and the border takes the base colour. This is the first build where the tool does something a player asked for.

### Two things this does not do yet — both correctly out of scope, both worth naming

**`constants: null` means no rollup ever dispatches in the application.** The PR says so plainly and I confirmed it: `Curated` appears in `web/src` only in `boundary/requests.ts` (the type) and `card/useConfiguredBase.ts` (a consumer), and the only values anywhere are in `tests/boundary/stages-integration.spec.ts` and `tests/fixtures/card-config-harness.tsx`. ADR-0001 defines Tier 2 as a hand-maintained curated set, but **no Tier 2 artifact is built, shipped, or loaded** — nothing in `cmd/` or `internal/domain/tier1.go` emits one. So assignments are held and coloured, and the recompute they are supposed to cause cannot happen for want of constants rather than for want of code.

That is a real hole with no issue covering it, and it blocks SPEC-0007's card just as hard. I'll open one.

**The bases are "Base 1…6", not the player's bases.** Placeholder identity slots, which is the honest shape while ADR-0010 and SPEC-0011 are the things that make places first-class and authorable. Nobody should read this PR as delivering "assign work to my actual bases" — it delivers the control and the binding.

Merging.

Issue #90, opened 2026-08-28
Tests: node card, controls, and provenance


jonstump:
Part of #84. Covers #86 and #87. Governed by [SPEC-0006](docs/openspec/specs/tree-canvas/spec.md).

**Blocked by #87** — it covers that story and #86 beneath it

> **Partially covered by #158.** Six of the eight requirements below were already met by the tests #86 and #87 shipped, and the two remaining gaps — the backdrop close route, and the colour-never-alone enumeration — are closed there (the enumeration turned out to already exist in `a11y-baseline.spec.ts`, which iterates `STATUSES` and guards the enumeration itself).
>
> **What is left is recipe-only.** "Options come from the payload" for the *recipe* control, and "recipe control at real size (61 options)", both need a control that does not exist: #87 shipped method-only and the recipe control is #132, blocked on design. This issue stays open for those two and should close with #132..

## Test Requirements

- [ ] **Border ownership**: a leaf that is assigned, hovered, focused and selected at once still shows its base colour on the border. A rendered assertion, not a class-list check — the states are a filter, an outline and an overlay, and only the paint result shows whether one of them reached the border
- [ ] **Colour is never alone**: for the unassigned state and the provenance marker, the distinction survives colour being removed. Enumerated from the component's own status table rather than a list written in the test, so a state added without a second carrier fails here
- [ ] **Unrounded application counts**: a node whose `applications` is an exact rational renders it exactly. Include a value with no finite binary expansion, so a pipeline that parsed a double and printed a fraction back cannot pass by luck
- [ ] **Provenance propagation**: a tree with one deep unverified node shows the marker on every ancestor up to the target, and the marker is visually distinct from both the warning and the error treatments
- [ ] **Options come from the payload**: the method and recipe controls offer exactly what the payload carries, with an unavailable method rendered, inert, and stating why
- [ ] **Recipe control at real size**: exercised at 61 options, not at three
- [ ] **Defaults record nothing**: a plan where every node is on its default produces plan state with no recipe selections
- [ ] **Keyboard operation**: both controls open, navigate, select and dismiss by keyboard, with focus returning to the invoking node on close by **every** route — Escape, backdrop, and the close control, each tested separately

## Acceptance Criteria

- Every focus-return route is a separate test. A single "close the popover" test passes while two of the three routes are broken — this is why #82's focus trap puts the restore in the effect cleanup rather than in any handler, and the tests must be able to tell the difference
[...]

PR #158, opened 2026-08-30
test(web): close the gaps the five test stories actually name

### LINKED ISSUE #89: Tests: rendering, ordering, and layout independence
Part of #84. Covers #85. Governed by [SPEC-0006](docs/openspec/specs/tree-canvas/spec.md).

**Blocked by #85** — there is nothing to render until it exists.

Three of that story's guarantees are the kind that pass by accident and fail silently. This story makes each of them mechanical.

## Test Requirements

- [ ] **Crossing count**: rendering a 36-node tree issues exactly one `resolve`. Counted at the network or client boundary, not asserted from reading the code — the failure mode is a component that re-resolves per node and looks identical on screen
- [ ] **Order preservation**: the rendered node sequence equals the payload's sequence, over a real payload rather than a fixture of three nodes. Plus a source-level check that the canvas contains no comparator applied to nodes, since a sort that happens to agree with the payload today passes any output comparison
- [ ] **Layout input**: the layout engine's call site receives nodes and edges only. Asserted by capturing what is actually passed, so a total reaching it fails here rather than at review
- [ ] **Layout independence**: changing target quantity leaves every node position unchanged. Positions compared exactly, both before and after a recompute that changes every total
- [ ] **Tab order**: focus moves through nodes in the payload's order and not the layout's — the deliberate divergence design.md records, which a later "fix" would otherwise quietly reverse
- [ ] **Edge facts**: every edge shows its per-unit quantity, and with edge styling stripped entirely, every fact the styling conveyed is still present as text on the connected nodes

## Acceptance Criteria

- Each test fails against a deliberately broken canvas, checked by breaking it rather than assumed — the `scripts/mutation-check.sh` pattern from #78, extended to cover this story's rules
- The layout-input assertion captures the real argument list; a test that only checks positions did not move would pass a layout engine handed a total it ignored
- The order test uses a payload of at least 36 nodes. Ordering bugs do not reproduce at three
- The edge-facts test removes styling rather than reasoning about it, so "the fact is also in the text" is observed and not argued

## Notes

The crossing-count and layout-input tests are the two worth writing first. Both guard rules that are invisible on screen when broken: a canvas that resolves per node renders correctly and is unusably slow on a large tree, and a layout engine handed a total produces a correct picture right up until someone uses the value.

### Branch

`feat/89-tests-rendering-ordering-and-layout`

### PR Convention

```
Closes #89
Part of #84

Implements SPEC-0006 REQ "Graph Rendering From the Boundary Payload", REQ "Layout Geometry Is Not a Domain Value", REQ "Edge Rendering".
```

### LINKED ISSUE #100: Tests: composition, identity, and producer sections
Part of #94. Covers #96. Governed by [SPEC-0007](docs/openspec/specs/base-planner-card/spec.md).

**Blocked by #96.**

## Test Requirements

- [ ] **Counts are the domain's**: every count on a rendered card traces to a payload field. A source-level check that no quantity-divided-by-rate appears in the card, alongside a rendered assertion that a farm row's plants and biodomes match the payload exactly — the source check catches the construct, the render check catches the result
- [ ] **Base-level counts stay base-level**: a base with three kitchen steps shows its nutrient processor count once. Asserted against a three-step payload, because the overcount only appears above one step
- [ ] **Selection semantics**: the selection control reports the right role and is keyboard-operable. A `div` with `tabIndex` must fail this even when it behaves correctly, so the assertion is on the accessible role rather than on the click working
- [ ] **Identity survives selection**: a selected card's base-colour frame is unchanged, and the ring is an overlay — asserted on the rendered result, since the identity frame and the ring are different mechanisms and only the paint shows whether one wrote to the other
- [ ] **Colour is not identity**: with colour removed, every card is still identifiable by name
- [ ] **All four producer groups**: a payload carrying farm, extractor, ranch and kitchen renders all four; a payload missing one renders three, not four with an empty
- [ ] **Instruction over demand**: a farm row shows plants and biodomes, not the required quantity alone
- [ ] **Pessimistic bound**: where a crop yield is a range, the planning figure is the pessimistic bound. Asserted against a payload with a range, since a single-value payload cannot distinguish the two
- [ ] **Byproducts appear**: a demand covered by another producer's byproduct renders as a row naming the covering producer and marked as needing nothing built

## Acceptance Criteria

- Each test fails against a deliberately broken card, checked by breaking it — extend `scripts/mutation-check.sh`, which #93 grew to 21 mutations covering the stylesheet and the accessibility primitives
- The three-kitchen-step case is a real payload, not a single step generalised. The bug it catches does not reproduce at one step
- The selection-semantics test asserts the role, so a correct-behaving wrong element fails
- The range test uses a payload where the bounds differ. Equal bounds cannot tell the two readings apart

### Branch

`feat/100-tests-composition-identity-and-producers`

### PR Convention

```
Closes #100
Part of #94

Implements SPEC-0007 REQ "Card Composition From the Build Payload", REQ "Base Identity and Selection", REQ "Producer Sections", REQ "Byproducts Are Shown, Not Omitted".
```

### LINKED ISSUE #101: Tests: power, deficit, provenance, and absence
Part of #94. Covers #97, #98 and #99. Governed by [SPEC-0007](docs/openspec/specs/base-planner-card/spec.md).

**Blocked by #98 and #99.**

## Test Requirements

- [ ] **Mixed power sources**: a base configured with electromagnetic generators *and* solar panels renders its generation correctly. The case the prototype's `EM | SOLAR` toggle cannot express, and the one most likely to be lost by building from the design
- [ ] **Solar carries no class**: no class control is offered for solar, anywhere on the card
- [ ] **One class control per site**: a card with three extractor rows has exactly one extractor class control
- [ ] **Reconfiguration crosses the boundary**: changing class, fill duration, generator count or panel count issues a boundary call, and no figure is arithmetic on a previous figure
- [ ] **The shortfall is the domain's**: no subtraction of draw from generation in the card's source, and the displayed figure matches the payload
- [ ] **`fixUnsized` is a visible deficit**: a payload with zero additional generators and `fixUnsized: true` renders the deficit and states the fix needs a class. Distinct from `fixUnsized: false` with the same zero — the two payloads differ in one boolean and must render differently
- [ ] **No invented fix**: a solar base in deficit with no sized fix offers no panel count
- [ ] **Footer agrees with the sections**: every footer item corresponds to a row above it, compared programmatically
- [ ] **No completion fraction**: nothing on the card renders a built-versus-total figure
- [ ] **Durations read as estimates**, and no unit-conversion arithmetic appears in the card's source
- [ ] **Row and base provenance are independent**: a verified base with an unverified row marks the row and not the base
- [ ] **Legible with everything marked**: the card rendered with every producer row unverified — the real condition today — stays readable, and the marker is distinct from the warning and error treatments
- [ ] **No placeholder environment data**, and **no control implying persistence** anywhere on the card

## Acceptance Criteria

- Each test fails against a deliberately broken card, checked by breaking it. The `fixUnsized` and mixed-source mutations are the two worth adding first: both are states an implementer building from the prototype would not produce, so a test that does not discriminate them will look fine
- The `fixUnsized` pair differs by one boolean and must produce two different renders. A test that only exercises the sized case passes against an implementation that ignores the flag entirely
- The everything-unverified case uses a payload where no constant is verified, which is what the current constant set actually produces — not a synthetic worst case
- The persistence check is an absence, so it is asserted over the rendered control set rather than by exercising a flow

### Branch

`feat/101-tests-power-deficit-and-presentation`

### PR Convention

```
Closes #101
Part of #94

Implements SPEC-0007 REQ "Site Configuration", REQ "Power Configuration Supports Mixed Sources", REQ "Power Position", REQ "Deficit Is an Action, Including When It Cannot Be Sized", REQ "Build Rollup Footer", REQ "Duration Display", REQ "Provenance on Displayed Figures", REQ "Absent Data Is Absent".
```

### LINKED ISSUE #115: Tests: preferences, empty state, deletion, and custody

jonstump:
Closes #89
Closes #100
Closes #101
Closes #115
Part of #84, #94, #110

## What this is, and what it deliberately is not

The five test stories were **audited against the existing suite before anything was written**. Most of what they ask for already exists — each implementation story shipped its own tests — so this closes the gaps rather than restating coverage.

Written as specified, these five would have produced roughly fifty duplicate tests and buried the handful of things genuinely missing. The audit is the deliverable as much as the code is.

| Story | Requirements | Already covered | Gap closed here |
|---|---|---|---|
| #89 rendering, ordering, layout | 6 | 5 | edge-facts must *remove* styling, not argue |
| #100 composition, identity, producers | 9 | 8 | source check for quantity ÷ rate |
| #101 power, deficit, provenance | 13 | 10 | subtraction source check · `fixUnsized` contrast · 3 more reconfig fields |
| #115 preferences, empty state, deletion | 10 | 9 | backdrop close route |
| #90 node card, controls, provenance | 8 | 5 | backdrop route · **2 blocked on #132** |

Things I expected to be gaps and verified were not: #90's "colour is never alone, enumerated from the component's own status table" is already done — `a11y-baseline.spec.ts` iterates `STATUSES` *and* guards the enumeration itself. I checked rather than assumed.

## The two that matter

Both are absences no rendered assertion can catch.

`numericConversions` matches `Number(`, `parseInt`, `Math.` and deliberately **not** `-` or `/` — the shared checker's own comment explains why: `+` concatenates strings throughout the codebase and flagging it "would be turned off within a week". That reasoning does not extend to the other two operators, and both are precisely what the requirements name.

**Division.** *"No quantity-divided-by-rate appears in the card's source"* is written at the top of `tests/card/discipline.spec.ts` — and was enforced by nothing. The file's stated purpose and its actual check disagreed. A farm row carries `required` and `yieldPerPlant`, so `required / yieldPerPlant.min` looks like the plant count and is not; SPEC-0007 requires the domain's `plants` *even where both are in the payload*.

**Subtraction.** *"No subtraction of draw from generation"* had only a rendered assertion — which **cannot distinguish a computed balance from the payload's when the two agree, and they always will**. I raised this reviewing #128 and it is now enforced.

`domainArithmetic` is the operator half: scoped to the domain's vocabulary and to *spaced* operators, so hyphenated class names and JSX's `/>` cannot match. Both real source trees are clean; the mutations prove the checker sees it when they are not.

## Also closed

- **Edge facts, observed not argued.** #89's criterion asks the test to *remove* styling. The existing test reads a `data-` attribute and finds the same word on a card — a claim about the model. The new one flattens every edge to one stroke and dash, **asserts the flattening actually worked**, then requires every fact to still be readable as text.
- **`fixUnsized` contrast.** A new fixture base with the *same* zero additional generators and the flag off. One boolean apart, and they must render differently — without it, the existing test passes against a card that states the unsizeable fix whenever the count is zero.
- **Reconfiguration beyond class.** #101 lists four fields; only the extractor class was covered. Fill duration, generator count and panel count now each dispatch, driven separately so a partial regression cannot hide.
- **Backdrop close route**, on the method control and the deletion confirmation, each asserted separately.

## A test bug worth recording

My first backdrop tests failed and the product was fine. The backdrop spans the viewport, so its **centre point sits underneath the dialog** — `click({force: true})` skips the actionability check but still clicks the centre, which lands on the dialog and dismisses nothing. A direct DOM `.click()` closed it immediately. Both tests now click a corner, and the comment says why.

## Not here

Two of #90's requirements — the recipe control's options and its behaviour at 61 options — need a control that does not exist. That is **#132**, blocked on design. #90 stays open for them; everything else it names is covered.

## Verification

**315 tests pass** (up from 307). lint, lint:css, typecheck, format:check, check:tokens, build all clean. **55/55 mutations red**, three new:

| Mutation | Catches |
|---|---|
| the card computes a plant count it was given | division |
| the card subtracts draw from generation | subtraction |
| an unsizeable fix is claimed for a fix the domain sized | the one-boolean contrast |

No protected paths touched.

### DIFF
--- a/web/scripts/mutation-check.sh
+++ b/web/scripts/mutation-check.sh
@@ -49,7 +49,10 @@ CONTROL="src/canvas/NodeControl.tsx"
 ASSIGN="src/canvas/useLeafAssignment.ts"
 BASES="src/canvas/bases.ts"
 
-MUTABLE="$MUTABLE $LAYOUT $TREE $METHODS $CONTROL $ASSIGN $BASES"
+PLANNERCARD="src/card/BasePlannerCard.tsx"
+POWERBLOCK="src/card/PowerBlock.tsx"
+
+MUTABLE="$MUTABLE $LAYOUT $TREE $METHODS $CONTROL $ASSIGN $BASES $PLANNERCARD $POWERBLOCK"
 
 # shellcheck disable=SC2086
 restore() { git checkout -- $MUTABLE 2>/dev/null || true; }
@@ -572,6 +575,34 @@ check "the assignment announcement stops naming the base" \
   tests/canvas/assignment.spec.ts "$SHELL" \
   "      const label = BASES.find((base) => base.id === baseId)?.label;" \
   "      const label: string | undefined = undefined;"
+
+# ----------------------------------------------------------------------
+# The absences the test stories asked for and nothing was enforcing.
+#
+# `numericConversions` catches Number()/parseInt/Math., and deliberately not
+# `-` or `/`. So "no quantity-divided-by-rate appears in the card's source"
+# was written at the top of tests/card/discipline.spec.ts and enforced by
+# nothing, and "no subtraction of draw from generation" had only a rendered
+# assertion — which cannot tell a computed balance from the payload's when
+# the two agree, and they always will.
+# ----------------------------------------------------------------------
+
+check "the card computes a plant count it was given" \
+  tests/card/discipline.spec.ts "$PLANNERCARD" \
+  '        <Figure label="Plants" value={q(row.plants)} />' \
+  '        <Figure label="Plants" value={q(row.required / row.yieldPerPlant.min)} />'
+
+check "the card subtracts draw from generation" \
+  tests/card/discipline.spec.ts "$POWERBLOCK" \
+  '        <Figure label="Balance" value={formatQuantity(budget.balance)} />' \
+  '        <Figure label="Balance" value={formatQuantity(budget.generation - budget.draw)} />'
+
+# The two payloads differ in one boolean. A card that stated the unsizeable
+# fix whenever the count was zero would pass every other power assertion.
+check "an unsizeable fix is claimed for a fix the domain sized" \
+  tests/card/power.spec.ts "$POWERBLOCK" \
+  "  const sized = budget.inDeficit && !budget.fixUnsized;" \
+  "  const sized = false;"
 
 echo
 if [ "$failures" -gt 0 ]; then
--- a/web/tests/canvas/edges.spec.ts
+++ b/web/tests/canvas/edges.spec.ts
@@ -211,3 +211,86 @@ test("the per-unit figure is text, not a line thickness", async ({ page }) => {
 
   expect(new Set(widths).size, "edge width varies, and only quantity varies").toBe(1);
 });
+
+test("with edge styling actually removed, every fact it carried is still there", async ({
+  page,
+}) => {
+  /*
+   * #89's acceptance criterion words this precisely: the test "removes
+   * styling rather than reasoning about it, so 'the fact is also in the
+   * text' is observed and not argued".
+   *
+   * The companion test above argues it — it reads each edge's method off a
+   * data attribute and finds the same word on the card. That is a claim
+   * about the model, and it would still pass if the styling were the only
+   * thing a *person* could see.
+   *
+   * So this one flattens the styling for real: every edge is forced to one
+   * stroke, one dash pattern and one width. The first assertion is that the
+   * flattening worked — without it, "the facts survived" would be true
+   * because nothing was removed.
+   */
+  await resolve(page, "ANTIMATTER");
+  const canvas = page.getByRole("region", CANVAS);
+
+  const before = await canvas.locator(".tree-edge").evaluateAll((nodes) =>
+    nodes.map((node) => {
+      const style = getComputedStyle(node);
+      return `${style.stroke}|${style.strokeDasharray}`;
+    }),
+  );
+  expect(
+    new Set(before).size,
+    "the edges were already identical, so removing styling proves nothing",
+  ).toBeGreaterThan(1);
+
+  await page.addStyleTag({
+    content: `.tree-edge, .tree-edge[data-method] {
+      stroke: rgb(128 128 128) !important;
+      stroke-dasharray: none !important;
+      stroke-width: 1.5px !important;
+    }
+    .edge-label, .edge-label[data-method] {
+      color: rgb(128 128 128) !important;
+      background-color: transparent !important;
+      border-color: transparent !important;
+    }`,
+  });
+
+  const after = await canvas.locator(".tree-edge").evaluateAll((nodes) =>
+    nodes.map((node) => {
+      const style = getComputedStyle(node);
+      return `${style.stroke}|${style.strokeDasharray}`;
+    }),
+  );
+  expect(new Set(after).size, "the styling did not actually flatten").toBe(1);
+
+  /*
+   * Now every fact the appearance carried has to be readable as text. The
+   * per-unit quantity is on the edge's own label; the method it fed is on
+   * the card it points at.
+   */
+  const expected = await payloadEdges(page);
+  const labels = await canvas
+    .locator(".edge-label")
+    .evaluateAll((nodes) => nodes.map((node) => node.textContent ?? ""));
+  expect([...labels].sort()).toEqual([...expected.map((edge) => edge.perUnit)].sort());
+
+  const methodsOnCards = await canvas
+    .locator(".react-flow__node")
+    .evaluateAll((nodes) =>
+      Object.fromEntries(
+        nodes.map((node) => [
+          node.getAttribute("data-id") ?? "",
+          node.querySelector(".node-method")?.textContent ?? "",
+        ]),
+      ),
+    );
+  for (const edge of expected) {
+    const target = edge.id.split("->")[1] ?? "";
+    expect(
+      methodsOnCards[target],
+      `with styling gone, nothing says the edge into ${target} feeds a ${edge.targetMethod} step`,
+    ).toContain(edge.targetMethod);
+  }
+});
--- a/web/tests/canvas/method-control.spec.ts
+++ b/web/tests/canvas/method-control.spec.ts
@@ -347,6 +347,35 @@ test.describe("in the shell", () => {
     await expect(card).toBeFocused();
   });
 
+  test("the backdrop returns focus too — the third route", async ({ page }) => {
+    /*
+     * #90 asks for every close route "each tested separately": Escape, the
+     * backdrop, and the close control. The trap's backdrop route is proven
+     * once on the shell's own popover in a11y-primitives.spec.ts; this
+     * asserts it on *this* dialog, because a surface that added its own
+     * backdrop handler would break here and nowhere else.
+     */
+    const card = page
+      .getByRole("region", CANVAS)
+      .locator(".node-card")
+      .filter({ hasText: "Antimatter" })
+      .first();
+
+    await card.click();
+    await expect(page.getByRole("dialog")).toBeVisible();
+
+    /*
+     * A corner, not the centre. The backdrop spans the viewport, so its
+     * midpoint is underneath the dialog — `force: true` skips the
+     * actionability check but still clicks the centre point, which lands on
+     * the dialog and dismisses nothing. This cost a debugging detour.
+     */
+    await page.locator(".popover-backdrop").click({ position: { x: 8, y: 8 } });
+
+    await expect(page.getByRole("dialog")).toHaveCount(0);
+    await expect(card).toBeFocused();
+  });
+
   test("focus stays inside the control while it is open", async ({ page }) => {
     await openControlFor(page, "Chromatic Metal");
     const dialog = page.getByRole("dialog");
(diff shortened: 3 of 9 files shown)


jonstump:
## Review — approved

Conflict gate clean, CI 10/10, zero commits behind `main`. Verified locally: **315 tests pass, 55/55 mutations red**, typecheck / lint / lint:css / format:check / check:tokens / build all clean.

### The audit is the deliverable, and it holds up

Closing four stories by showing most of what they ask for already exists is the kind of claim that deserves checking rather than crediting, so I picked the one with the largest asserted gap — #101's reconfiguration fields — and read `tests/card/configuration.spec.ts` **as it stands on main**:

```
changing the extractor class issues a boundary call     ← the only crossing assertion
a base runs electromagnetic generators and solar panels together   ← fills, then renders
configuring solar surfaces the domain's battery count              ← fills, then renders
fill duration is exposed, not implied                              ← visibility only
```

All four fields were *touched*; exactly one asserted that changing it crossed the boundary. So "only the extractor class was covered" is accurate rather than generous, and driving the other three separately "so a partial regression cannot hide" is the right shape.

The refusal to write fifty duplicate tests is correct. A suite that restates coverage is worse than one that doesn't, because it makes the gaps harder to see.

### The subtraction check

> *"cannot distinguish a computed balance from the payload's when the two agree, and they always will"*

This is the finding I raised reviewing #128 and it is now enforced rather than noted. That is the right end state for it.

### I probed `domainArithmetic` rather than reading it

Ran the checker over ten snippets to find out what it actually does:

```
CAUGHT  natural division            required / yieldPerPlant.min
CAUGHT  natural subtraction         generation - draw
missed  renamed operands            need / per
missed  unspaced division           required/yieldPerPlant.min
missed  a domain term in a comment
missed  a hyphenated class name     card-row-name
missed  JSX self-close              <Figure label={total} />
missed  a loop counter              rows.length - 1
missed  negative literal            -1
```

The bottom five are the point: **no false positives** on any of the shapes the spacing rule and the vocabulary scoping were designed to exclude. `scan` already strips comments, so the comment case is safe by construction.

The unspaced miss is closed by another gate, which I verified rather than assumed — prettier rewrites `row.required/row.yieldPerPlant.min` to the spaced form, and `format:check` runs in CI, so the unspaced shape cannot land. Worth one line in the comment that the spacing rule leans on the formatter, since that is a dependency between two files.

The renamed-operand miss is inherent to scanning by identifier, and the checker catches the form the requirement actually describes. Not a defect — but it is the reason this check complements the rendered assertions rather than replacing them.

### One thing to add

**The two new backdrop tests are real, and nothing keeps them that way.**

I removed `onClick={onBackdrop}` from `Popover.tsx` and re-ran both files:

```
2 failed, 30 passed
  method-control.spec.ts:350  the backdrop returns focus too — the third route
  data-custody.spec.ts:397    the backdrop dismisses the confirmation and returns focus
```

Exactly those two of thirty-two. The coverage is precise and it works today.

But no mutation names either of them. The existing focus-route pair names `a11y-primitives.spec.ts`, and its comment explains why routes are tested separately — *"the Escape-only implementation leaves the Escape test green while the backdrop and close-control tests go red"* — which is the same argument these two tests are an instance of, on two concrete surfaces.

Given this PR's own thesis, and given it records that its first backdrop attempt failed while the product was fine, these are the tests most owed a mutation. The corner-click detail makes it sharper: a future edit that re-centres the click would leave both green forever, and nothing would say so.

Not blocking — the probe above is today's proof. Worth two entries in `mutation-check.sh`, and I'm happy to add them.

### Bookkeeping

`Closes #89, #100, #101, #115` with **#90 left open** for the two requirements that need #132's control is the right call, and the body says which two.

Merging.

Issue #143, opened 2026-08-29
Implement Shell Surface


jonstump:
Implements [SPEC-0011: Shell Surface](../blob/main/docs/openspec/specs/shell-surface/spec.md) ([design.md](../blob/main/docs/openspec/specs/shell-surface/design.md)).

Governed by [ADR-0010](../blob/main/docs/adrs/ADR-0010-places-first-and-the-shell.md) — a player authors places first, and a plan assigns leaves to places that already exist.

## Why this epic exists

SPEC-0007 deferred this as "application-shell furniture … belongs with the shell surface, which has no spec yet". The deferral is why `BasePlannerCard` is imported by exactly one file under `web/src/` — a type-only import of `IdentitySlot` in `canvas/NodeCard.tsx` — and renders only in Playwright fixtures. It is why `<nav aria-label="Surfaces">` holds one button with no handler. And it is why the target control is a bare `<input>` demanding an item id the player cannot know.

The load-bearing content is not navigation. It is the direction of the base/plan relationship: `BaseID` stops being a key a plan invents and becomes the identifier of a place record that outlives every plan.

## Stories

| # | Story | Label |
|---|---|---|
| #145 | Places are first-class: identity, hand creation, and unassignment | `foundation` |
| #146 | The shell: bases-first entry and surface navigation | |
| #147 | Target search and the catalogue boundary call | |
| #148 | State ownership: the hash carries the plan, the store carries the player | |
| #149 | Tests: place identity, unassignment, and state ownership | `test` |
| #150 | Tests: shell entry, navigation, and target search | `test` |

**Order:** #145 (foundation) → #146, #147, #148 in parallel → #149, #150.

#147 changes the module's contract version — it is the only story here that grows the boundary surface.

## Out of scope

The visual form of any control (the design bundle owns it), the sharing model, the identity flow (ADR-0009), and save import (SPEC-0008 — this epic makes hand-authoring the *other* route, not a replacement).

### Branch

`epic/143-implement-shell-surface`

Issue #148, opened 2026-08-29
State ownership: the hash carries the plan, the store carries the player


jonstump:
Part of #143. Governed by [SPEC-0011](../blob/main/docs/openspec/specs/shell-surface/spec.md) and [ADR-0010](../blob/main/docs/adrs/ADR-0010-places-first-and-the-shell.md).

Both state mechanisms now exist and each is individually a reasonable home for the same value, so the drift would be silent. This story fixes the boundary and asserts it.

`docs/design/README.md` line 37 lists "todo checks" among hash contents. That line predates [ADR-0008](../blob/main/docs/adrs/ADR-0008-durable-user-data-store.md) and does not govern; ticks are per-place player data and belong in the store.

## Requirements

- [ ] **The Hash Owns the Plan, the Store Owns the Player** (SPEC-0011) — The URL hash MUST carry plan state only: target, quantity, method and recipe selections, and leaf assignments. The durable store MUST own player-authored data: places and their annotations, ticked build items, and view preferences. No player-authored durable value may be encoded into the hash, and no hash-derived value may be written to the store as though the player authored it. Hash-derived state remains untrusted input under SPEC-0005 § Redirect Validation.
  - WHEN a hash is encoded from a workspace with named places, notes and ticked items THEN the encoded value contains plan state only, with no place annotation, note or tick present
  - WHEN a hash carrying assignments is decoded THEN no place record is created or modified as a result

- [ ] **Signing In Is Not a Sync Trigger** (SPEC-0011) — Completing a sign-in MUST attach an owner to the workspace and MUST NOT transmit any place record by itself. This is the boundary ADR-0008 fixed and ADR-0009 §4 carried forward; the shell is where sign-in is invoked and "authenticate, then reconcile" is the natural shape to build here.
  - WHEN a player signs in with places stored and nothing marked shared THEN no place record is transmitted

## Acceptance Criteria

- Encoding a hash from a fully-populated workspace yields plan state only. The test needs notes and ticks present, or it proves nothing
- Decoding a hash authors no store record. This is the direction that would be easy to get wrong while making shares "work better"
- ADR-0008's opt-in-sync test extends to cover the moment of authentication, not only the steady state
- The `docs/design/README.md` line 37 divergence is resolved in favour of ADR-0008, and the reasoning is recorded rather than the line silently ignored

### Branch

`feat/148-state-ownership-hash-carries-plan-store-carries`

### PR Convention

```
Closes #148
Part of #143

[...]

PR #159, opened 2026-08-30
feat: places are first-class — identity, hand creation, and unassignment

### LINKED ISSUE #145: Places are first-class: identity, hand creation, and unassignment

**Foundation story.** It changes what `BaseID` means, and every other story in this epic assumes the new meaning. Merge this before the others begin.

Extends `internal/domain/rollup.go` (where `BaseID` is currently documented as identifying "a base within a plan"), `internal/domain/producer.go` (`BaseBuild`), and `web/src/store/durable-store.ts` (`putPlace`, which exists at line 264 with no caller under `web/src/`). The capability to create a place is already there; what is missing is the route to it.

## Requirements

- [ ] **A Place Is Authored, and a Plan References It** (SPEC-0011) — A place MUST exist independently of any plan. `BaseID` MUST be the place record's `id` as SPEC-0009 REQ "A Place Is One Record Type, Whatever Its Kind" defines it — the same value, not a mapping onto it. The application MUST NOT mint a second identifier, and MUST NOT derive a place's identity from a plan's assignments. A plan MUST remain resolvable when it references no places at all.
  - WHEN a place is created and later assigned a leaf THEN the value the plan carries as `BaseID` is the place record's own `id`, and no other identifier for that place exists in the store
  - WHEN a plan is resolved against a workspace holding no places THEN it resolves, and every leaf is reported unassigned rather than failing

- [ ] **An Assignment Naming an Absent Place Is Unassigned** (SPEC-0011) — Where a plan carries an assignment whose `BaseID` matches no place in the workspace, the leaf MUST be treated as unassigned and presented as such. Deleting a place MUST NOT delete a plan, MUST NOT remove its leaves, and MUST NOT leave an identifier rendered as a base with no name. MUST hold for every source of an assignment, including one decoded from a hash authored on another device.
  - WHEN three leaves are assigned to a place and that place is deleted THEN the plan survives, the three leaves appear in the unassigned group, and no dangling identifier is rendered
  - WHEN a hash carrying assignments is decoded on a device whose workspace holds none of the named places THEN the plan renders with those leaves unassigned, and no error state is presented

- [ ] **A Place Is Creatable by Hand** (SPEC-0011) — The bases surface MUST provide a route to create a place without a save file, independent of SPEC-0008. The minimum before a plan may assign to it is its generated `id` and a name; every other field MUST be optional at creation. A place with no site configuration MUST be assignable, its rollup treating the site as unconfigured and the card rendering that absence per SPEC-0007 REQ "Absent Data Is Absent".
  - WHEN a place is created with a name and nothing else THEN it persists across a reload and a plan can assign a leaf to it
  - WHEN a leaf is assigned to a place that has no site configuration THEN the card presents the missing configuration as absent, not as a configured value of zero

## Acceptance Criteria

- No second identifier for a place exists anywhere in the store or the domain — checkable by reading the schema, not by review
- The `BaseID` doc comment in `internal/domain/rollup.go` no longer says "within a plan", because it is no longer true
- Deleting a place with leaves assigned is the case that catches a cascade: the plan must survive intact
- A place created with only a name is assignable. If it is not, the minimum has grown and the first-run path is longer than the thing it unblocks
- `Unassigned` remains the empty value and is not a place

### Branch

`feat/145-places-are-first-class-identity-hand-creation`

### PR Convention

```
Closes #145
Part of #143

Implements SPEC-0011 REQ "A Place Is Authored, and a Plan References It", REQ "An Assignment Naming an Absent Place Is Unassigned", REQ "A Place Is Creatable by Hand".
```

jonstump:
Closes #145
Part of #143

Implements SPEC-0011 REQ "A Place Is Authored, and a Plan References It", REQ "An Assignment Naming an Absent Place Is Unassigned", REQ "A Place Is Creatable by Hand".

`BaseID` stops meaning "a key this plan invented" and starts meaning "a place that exists". The doc comment in `internal/domain/rollup.go` says so, and the view stops minting the six identifiers that made it untrue.

## Identity

`web/src/canvas/bases.ts` derived `base-1` through `base-6` with placeholder labels, on the stated grounds that a base's real name came from a bases map with no spec and no surface. SPEC-0011 forbids exactly that — "The application MUST NOT mint a second identifier for a place" — so the assignable set is now the workspace's places and `Base.id` is the SPEC-0009 record's own `id`.

The identity slot survives as what it always was: which of six colour tokens a place draws with. It repeats past six places, because a colour is paint and an id is not.

## Creation

A place is creatable from the Saved places panel with a name and nothing else, which is the whole minimum the requirement sets. The form is a real `<form>`, so Enter submits and the route does not sit behind a pointing device.

The id comes from `crypto.randomUUID()` rather than the name. A name-derived id changes when the player renames a place, silently reassigning every leaf that pointed at it — the fragmentation ADR-0010 rejected the side-table for.

## Unassignment

`web/src/state/assignments.ts` resolves held assignments against the places that exist, **on read rather than on delete**. That is what makes the rule hold for "every source of an assignment": a hash authored on another device names places this one never had, and there is no deletion event to hang a prune on.

Deleting a place removes the record and nothing else. The assignment is kept, the leaf is presented unassigned, and restoring the place restores it — so deleting a place cannot destroy a plan. The resolved map is what crosses the boundary, so a deleted place's id never reaches the domain and no base that exists nowhere is ever rendered.

## An unconfigured place is assignable, and that changed the domain

`GroupLeaves` refused any base named in an assignment that had no site configuration. SPEC-0011 makes that base assignable, so the binary becomes three states.

Both halves of the old rule survive. **Nothing is defaulted** — an extractor class picked for the caller is still a number they never chose, so an unconfigured base sizes *no* extractors rather than sizing them at class `""`; it reports them as `Unsited` demands instead. **A site that is supplied is still validated**, because "not configured yet" and "configured wrongly" are different mistakes.

Only extraction needs the site. A crop's yield and a fauna's rate are the item's own facts, so farms and ranches size normally: an unconfigured place still says what to plant.

The card renders that gap as a gap — an "Awaiting site configuration" section carrying the demand and the words, never a count of zero. The extractor-class select gains a "Not configured" option while the class is empty, because a select whose value matches nothing falls back to its first option and would have read as class C.

## Boundary contract 1.2.0 → 1.3.0

`configured` and `unsited` are additive, but with a sharper edge than the previous two bumps: a consumer that does not know about `configured` reads a base with a zero site as one configured at class `""` for zero seconds — the configured-value-of-zero the requirement exists to prevent. The decoder requires the field rather than guessing, and a test asserts the refusal.

## Also

`DurableStore.deletePlace`, which SPEC-0009 did not have — its deletion story is workspace-wide and this rule needs one place at a time.

Six mutations added to the mutation check, covering the three rules a behavioural test cannot pin down alone: a name-derived id renders identically until a rename, a resolver that keeps everything looks right until a place is deleted, and a card that shows the site's zeros looks like a configured base.

## Verification

| Check | Result |
|---|---|
| `go test ./internal/...` | pass |
| Frontend suite | 326 passed |
| Typecheck, lint, `check:tokens` | clean |
| Mutation check | the first run found one unwatched rule; see below |

The mutation check earned its place here. Replacing `crypto.randomUUID()` with a slug of the name left the suite green — the comment beside it argued why that is wrong, and nothing was watching. Two tests now cover it, and I confirmed by applying the mutation by hand that both go red while the other seven in the file stay green: two places named the same are two places (a slug makes them one, and the second write overwrites the first), and the id carries no trace of the name.

The full mutation re-run was interrupted locally; CI runs the same gate and exits non-zero on any unwatched rule.

## Follow-up not in scope

Five shell assignment tests that assigned to `base-3` now create a place through the shipped route first. That is the honest shape — it is what a player does — but it means those tests exercise creation as a precondition, and a failure there will surface as an assignment failure.

---

### DIFF
--- a/internal/bridge/bridge.go
+++ b/internal/bridge/bridge.go
@@ -42,7 +42,14 @@ import (
 // 1.2.0 added provenance to every producer row and to each base, and the
 // verified-dates map to the request. Same reasoning: additive, but a
 // consumer that branches on provenance needs to know it is there.
-const ContractVersion = "1.2.0"
+//
+// 1.3.0 added `configured` to each base and the `unsited` rows beside it,
+// when SPEC-0011 made a place with no site configuration assignable. Same
+// reasoning again, with a sharper edge than the previous two: a consumer
+// that does not know about `configured` reads a base with a zero site as
+// one configured at class "" for zero seconds, which is exactly the zero
+// SPEC-0011 REQ "A Place Is Creatable by Hand" forbids presenting.
+const ContractVersion = "1.3.0"
 
 // Quantity is an exact quantity on the wire.
 //
--- a/internal/bridge/stages.go
+++ b/internal/bridge/stages.go
@@ -222,10 +222,37 @@ type NoBuildRow struct {
 	Verified bool     `json:"verified"`
 }
 
+// UnsitedRow is a demand that needs a site configuration the base does not
+// have.
+//
+// Governing: SPEC-0011 REQ "A Place Is Creatable by Hand"
+//
+// Carried across the boundary rather than dropped, for the same reason
+// NoBuildRow is: an absent row is indistinguishable from an overlooked
+// requirement, and here the whole point is that the view render the gap.
+type UnsitedRow struct {
+	ItemID   string   `json:"itemId"`
+	Name     string   `json:"name"`
+	Required Quantity `json:"required"`
+	Verified bool     `json:"verified"`
+}
+
 // BaseBuild is one base's construction instructions.
 type BaseBuild struct {
 	Base string `json:"base"`
 	Site Site   `json:"site"`
+
+	// Configured reports whether the base has a site configuration. When
+	// false, Site carries zero values that mean nothing and the view MUST
+	// render the absence rather than the zeros.
+	//
+	// Governing: SPEC-0011 REQ "A Place Is Creatable by Hand", SPEC-0007
+	// REQ "Absent Data Is Absent"
+	//
+	// Always emitted, never omitempty: false is the value that carries the
+	// meaning here, and a field that disappears when it matters most is the
+	// opposite of what the requirement asks for.
+	Configured bool `json:"configured"`
 
 	Farms      []FarmRow      `json:"farms,omitempty"`
 	Extractors []ExtractorRow `json:"extractors,omitempty"`
@@ -237,6 +264,10 @@ type BaseBuild struct {
 	PelletFeeders      Quantity `json:"pelletFeeders"`
 
 	NoBuild []NoBuildRow `json:"noBuild,omitempty"`
+
+	// Unsited are demands that need extraction at a base with no site
+	// configuration. Empty whenever Configured is true.
+	Unsited []UnsitedRow `json:"unsited,omitempty"`
 
 	// Verified is the base-level answer: every row here, and the constant
 	// that sized its processors.
@@ -502,6 +533,7 @@ func encodeBaseBuild(b domain.BaseBuild) BaseBuild {
 			ExtractorClass: string(b.Site.ExtractorClass),
 			FillSeconds:    QuantityOfInt(b.Site.FillSeconds),
 		},
+		Configured:         b.Configured,
 		NutrientProcessors: QuantityOfInt(b.NutrientProcessors),
 		PelletFeeders:      QuantityOfInt(b.PelletFeeders),
 		Verified:           b.Verified,
@@ -569,6 +601,14 @@ func encodeBaseBuild(b domain.BaseBuild) BaseBuild {
 			From:     n.From,
 			Required: QuantityOf(n.Required()),
 			Verified: n.Verified,
+		})
+	}
+	for _, u := range b.Unsited {
+		out.Unsited = append(out.Unsited, UnsitedRow{
+			ItemID:   u.ItemID,
+			Name:     u.Name,
+			Required: QuantityOf(u.Required()),
+			Verified: u.Verified,
 		})
 	}
 	return out
--- a/internal/bridge/stages_test.go
+++ b/internal/bridge/stages_test.go
@@ -550,3 +550,95 @@ func TestProducerProvenanceCrossesTheBoundary(t *testing.T) {
 		t.Error("base crossed unverified with every date supplied")
 	}
 }
+
+// SPEC-0011 REQ "A Place Is Creatable by Hand":
+// WHEN a leaf is assigned to a place that has no site configuration
+// THEN the boundary reports the base as unconfigured with the demand
+// unsized, rather than failing the call or reporting zero extractors.
+//
+// Asserted at the boundary rather than only in the domain because the view
+// renders the absence, and a field the encoder drops is a gap the view
+// cannot tell from a base with nothing to extract.
+func TestRollupCarriesAnUnconfiguredBase(t *testing.T) {
+	m := stagesModule(t)
+
+	blob, _ := callOK(t, m, "rollup", `{
+	  "plan":{"target":"widget","quantity":"10"},
+	  "assignments":{"crop_a":"base1","gas_a":"base1"},
+	  "constants":`+curatedJSON+`}`)
+
+	var out struct {
+		Data struct {
+			Build struct {
+				Bases []struct {
+					Base       string `json:"base"`
+					Configured bool   `json:"configured"`
+					Extractors []struct {
+						ItemID string `json:"itemId"`
+					} `json:"extractors"`
+					Farms []struct {
+						ItemID string `json:"itemId"`
+					} `json:"farms"`
+					Unsited []struct {
+						ItemID   string `json:"itemId"`
+						Required string `json:"required"`
+					} `json:"unsited"`
+				} `json:"bases"`
+			} `json:"build"`
+		} `json:"data"`
+	}
+	if err := json.Unmarshal(blob, &out); err != nil {
+		t.Fatalf("unparseable rollup output: %v", err)
+	}
+	if len(out.Data.Build.Bases) != 1 {
+		t.Fatalf("bases = %d, want 1", len(out.Data.Build.Bases))
+	}
+
+	base := out.Data.Build.Bases[0]
+	if base.Configured {
+		t.Error("a base with no sites entry crossed as configured")
+	}
+	if len(base.Extractors) != 0 {
+		t.Errorf("extractor rows = %d, want none", len(base.Extractors))
+	}
+	if len(base.Unsited) != 1 || base.Unsited[0].ItemID != "gas_a" {
+		t.Fatalf("unsited = %+v, want gas_a alone", base.Unsited)
+	}
+	if base.Unsited[0].Required == "" {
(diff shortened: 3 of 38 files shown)


jonstump:
## The failing test was not the test — fixed in `4d6d609`

`tests/shell/places.spec.ts` "two places with the same name are two places" was timing out on a **Create place button that was disabled and stayed disabled**. The CI call log has the shape of it:

```
- locator resolved to <button type="submit" …>Create place</button>
- element was detached from the DOM, retrying
- locator resolved to <button disabled type="submit" …>Create place</button>
- 57 × waiting … element is not enabled
```

Detached, replaced, and permanently disabled — so whatever the test typed had gone.

### The cause

`StoredPlaces` rendered `<CreatePlace>` inside **both** branches, at a different child position in each — second in the empty state (after the "Nothing saved" paragraph), first in the populated one. React reconciles a fragment's children by position, so creating the first place flips `data.empty`, the form moves from index 1 to index 0, and React unmounts and remounts it. Its `useState` name goes with it.

Confirmed rather than reasoned — a marker set on the input element, checked across each transition:

```
first create  (empty → populated):  input element survived = false
second create (populated → …):      input element survived = true
```

**So this is a product bug on the first-run path, every time.** A player types their first place, submits, starts typing the second, and loses it the moment the write settles — then finds a disabled button on an empty field with nothing saying why. Only ever the first one, which is why it reads as a flake.

The irony is that the submit handler clears the field early *precisely* to protect the next keystrokes — its comment says so: *"a field that empties late eats the first characters of it."* The remount ate them anyway, and ate all of them.

### The fix

One render site. The two states differ in what *follows* the form, not in whether there is one, so the form sits above the branch and the paragraph-or-list below it.

A shared `key` across the two branches would also hold the instance, and is worse: it can be half-applied, and it leaves two sites that have to keep each other's rule.

### What keeps it fixed

A regression test asserting the **input element's identity** across the transition, rather than the typed text — text can be lost for other reasons, and the remount is the specific fault. Plus mutation 59, which reintroduces it as a key varying with `data.empty`: the shape someone reaches for when they want to reset a form on a state change.

I also wrote the consequence test — type a second name while the first write settles, assert it survives — and **removed it after proving it toothless**. It passes against the unfixed component whenever the fill lands after the write settles, which locally it does every time. A test that passes against the defect is worse than no test, and this PR's own suite is built on that principle.

### Also rebased

The branch was one commit behind and `DIRTY`. Rebased onto `main` (now carrying #158); `mutation-check.sh` conflicted in two places and both were additive — #158's three mutations and this PR's block both belong, and `PLANNERCARD` was defined on both sides.

### Verified on the rebased branch

```
337 tests pass        59/59 mutations red
typecheck · lint · lint:css · format:check · check:tokens · build   all clean
```

Now `MERGEABLE`. Not reviewed on the merits yet — this was a fix to the red suite so it *can* be reviewed. Ready for `/sdd:review`.

jonstump:
## Review — approved

Conflict gate clean, CI 10/10, `CLEAN` against `main`.

**Disclosure:** the top commit on this branch (`4d6d609`, the create-form remount fix) is mine — I wrote it last turn to get the suite green so this PR could be reviewed at all. It has its own write-up above. What follows is a review of the two commits I did not write, plus the whole against #145's criteria.

### The five acceptance criteria

**"No second identifier for a place exists anywhere in the store or the domain — checkable by reading the schema, not by review."** Checked that way. `web/src/store/schema.ts` carries exactly one `id` on a place record. The only other id-shaped field is `ownerId`, which is the workspace's account owner and a different thing entirely.

**"The `BaseID` doc comment no longer says 'within a plan'."** Done, and `grep -rn "within a plan" internal/ web/src/` returns nothing repo-wide. The replacement earns its length — *"a plan-scoped key is invented by whichever plan first named the base, so the same real base acquires a different key in every plan and the annotations against it fragment — a failure that surfaces as data loss rather than as a type error."* That is the argument ADR-0010 rests on, stated where someone changing the type will read it.

**"Deleting a place with leaves assigned is the case that catches a cascade."** Verified in the running application rather than inferred — see below.

**"A place created with only a name is assignable."** Verified — see below.

**"`Unassigned` remains the empty value and is not a place."** `const Unassigned BaseID = ""`, `IsUnassigned()` unchanged, and the select renders it as the empty-string option.

### The domain change is the right shape

`GroupLeaves` used to refuse any base named in an assignment that had no site configuration. Turning that into three states is the risky part of this PR and it is done carefully:

- `Configured bool` as a **separate flag rather than a sentinel inside `SiteConfig`**, with the reason stated: the zero `SiteConfig` is a legitimate shape to hold — it is what a caller sends before choosing a class — and *"a struct that answers 'am I real' about itself makes every reader check a field they did not know was there."*
- Both halves of the old rule survive. Nothing is defaulted: an unconfigured base sizes **no** extractors rather than sizing them at class `""`. And an explicit entry with an unusable class is **still an error**, because *"the distinction between 'not configured yet' and 'configured wrongly' is the whole point."*

Removing an error path is where a permissive change usually loses something. This one gave up exactly the case the spec now permits and kept the rest.

### `resolveAssignments` — resolve-on-read is the correct answer

The requirement says the rule must hold *"for every source of an assignment, including one decoded from a hash authored on another player's device."* Resolving on read rather than pruning on delete is what makes that true without a new check per source — as the body puts it, a hash from another device *"names places this one never had, and there is no deletion event to hang a prune on."* One function, and a future source inherits the rule instead of remembering it.

**One accuracy point.** The module comment says *"The two sources today are a place the player deleted and a hash from another device."* The hash source is not live: `decodePlanFromHash` is exported but has no caller anywhere under `web/src`. The architecture is right and forward-looking; the word "today" is not. `assignment.spec.ts` is already precise about this (it points at #135), so it is only the one comment that overstates. #148 is the story that makes it true.

### Verified in the running application

Because "assign leaves to my actual bases" is the thing that has been missing, I drove it rather than reading it. Fresh workspace, place created with a name and nothing else:

```
created:   { id: "8c061ea7-0161-4bbd-af00-d019c2ec544a", name: "Verdant Shelf" }
select:    [ { label: "Unassigned",   value: "" },
             { label: "Verdant Shelf", value: "8c061ea7-0161-4bbd-af00-d019c2ec544a" } ]
```

The option's value **is the place record's own id**. That is criterion 1 observed rather than argued, and the generic "Base 1…6" placeholders are gone.

Then the cascade case:

```
assigned:  data-identity="1"          border rgb(250,189,47)   ← --base-1
delete the place ↓
after:     data-identity="unassigned" border rgb(80,73,69)     ← --border
           nodes rendered: 6   places: 0   card: "Indium 5/6 ▽ raw Unassigned"
```

The plan survived intact, the leaf returned to the unassigned state with its dashed neutral frame, and no dangling identifier was rendered anywhere. That is the criterion the issue singles out, and it passes on the live path.

### Verification

On the rebased branch: **337 tests pass, 59/59 mutations red**, typecheck / lint / lint:css / format:check / check:tokens / build all clean. Mutations worth naming — *"a place takes an id derived from its name instead of a generated one"*, *"an assignment naming a deleted place is kept rather than unassigned"*, and *"the unsited rows stop being rendered at all"* — each break one of the three requirements, and each goes red.

Merging.

PR #160, opened 2026-08-30
feat(web): the hash carries the plan, the store carries the player

### LINKED ISSUE #148: State ownership: the hash carries the plan, the store carries the player

jonstump:
Closes #148
Part of #143

Implements SPEC-0011 REQ "The Hash Owns the Plan, the Store Owns the Player", REQ "Signing In Is Not a Sync Trigger".

## This also settles #135

#135 asked where leaf assignments live, because `Plan` carried none, `RollupRequest` carried them *beside* the plan, and SPEC-0002 encoded neither. **SPEC-0011 answered it**: the hash carries "target, quantity, method and recipe selections, and leaf assignments". So `Plan` is the shareable unit and assignments are part of it. #135 can close as resolved by spec.

## The boundary

**The hash gains assignments.** A link that dropped them would share a plan whose leaves land nowhere.

**`planToWire` deliberately does not send them.** That is the shape `resolve` parses, and stage 1 has no concept of a base — `internal/bridge.Plan` carries target, quantity, methods and recipes and nothing else. Assignments reach the domain on the stage-2 rollup request. Sending them anyway would be *harmless* today (the Go decoder ignores unknown fields on a plan) and wrong in the way that matters: it would say stage 1 takes an input it does not take.

**The decoder reads five fields and has nowhere to put a sixth.** A hash carrying `ticks` or `notes` is not rejected — SPEC-0005 requires an *undecodable* hash yield the empty plan, and a valid plan with extra keys is not undecodable. The keys simply never reach a `Plan`. That is the cheapest way to honour *"no hash-derived value may be written to the store as though the player authored it"*: the decoder has no vocabulary for one.

## Which direction is guarded hardest

Decoding. A leaked note is visible the first time someone reads a link. **A decoded hash quietly authoring a place record is not visible at all**, and it is exactly the change someone makes while trying to make sharing "work better".

The encode test populates the store with a place carrying a name, notes, tags, ticks *and* stocked quantities before encoding — the acceptance criterion is explicit that it "needs notes and ticks present, or it proves nothing", and a hash encoded from an empty session contains no note for the trivial reason that there was none.

## Sign-in is the transition, not the steady state

The store's discipline suite already proves no store path reaches the network. What the criterion asks for is *the moment of authentication*. There is no sign-in yet and `ownerId` is the whole of what one would change, so the test writes it onto a workspace with places present and asserts the recorder saw nothing — and that the places are still there afterwards.

## The design README

Line 37 listed *todo checks* among hash contents and predates ADR-0008. Corrected **in favour of ADR-0008 with the reasoning recorded**, per the criterion, rather than the line quietly dropped: ticks describe what someone has built at a base, not what they asked the planner for, so sharing a link would hand the recipient someone else's progress and receiving one would overwrite theirs.

"No localStorage" is kept and narrowed — ADR-0008 lifted it for durable player data only, and that store is IndexedDB. Plan state keeps the original rule. The adjacent **State Management** paragraph has the same staleness and is pointed at rather than left to be rediscovered.

`docs/design/` is not a protected path; `docs/adrs` and `docs/openspec/specs` are, and neither is touched.

## Verification

**343 tests pass** (up from 315), 6 new. lint, lint:css, typecheck, format:check, check:tokens, build all clean. **61/61 mutations red**, two new:

| Mutation | Catches |
|---|---|
| the hash stops carrying assignments | a link that shares a plan landing nowhere |
| **the plan decoder starts accepting a player's ticks** | the invisible direction |

No protected paths touched.

### DIFF
--- a/docs/design/README.md
+++ b/docs/design/README.md
@@ -34,7 +34,15 @@ Specified per surface in the three handoff.md files. Cross-cutting rules:
 - **Borders carry identity only**: 3px base-color frames; hover = brightness(1.12), focus = 2px #8ec07c outline outboard, selection = 2px #8ec07c ring **inboard via overlay element** (never inset box-shadow — it paints under positioned children). Nothing else may write to a border.
 - Every recompute (method/class/type change, deficit fix, todo check) announces via `aria-live="polite"`.
 - Color is never the sole carrier: methods have glyph+text, bases have names, warnings have ⚠+text, deficits have stated kPs + red meter segment.
-- Plan state (target, methods, assignments, todo checks) should serialize into a shareable URL hash; no localStorage.
+- Plan state (target, methods, recipes, assignments) serializes into a shareable URL hash; no localStorage.
+
+  > **Superseded in part by [ADR-0008](../adrs/ADR-0008-durable-user-data-store.md).** This line originally listed *todo checks* among the hash contents and predates the durable-store decision. Ticked build items are per-place player data, not plan state: they describe what someone has built at a base, not what they asked the planner for. Sharing a link would otherwise hand the recipient someone else's progress, and receiving one would overwrite theirs.
+  >
+  > SPEC-0011 REQ "The Hash Owns the Plan, the Store Owns the Player" fixes the boundary: the hash carries target, quantity, method and recipe selections and leaf assignments; the store carries places, annotations, ticks and view preferences.
+  >
+  > "No localStorage" still holds and is narrower than it reads. ADR-0008 lifted it for durable player data only, and that store is IndexedDB — asynchronous, structured and quota-bounded. Plan state keeps the original rule.
+
+  The same correction applies to the per-base list in **State Management** below, which predates ADR-0008 for the same reason: its `todo done-map` and `note tags + text` are store-owned, not UI state.
 
 ## State Management
 Per-base UI state (v2 prototype's `ui` map is the reference shape): power type + EM class, extractor class, added generators, section open/closed, todo done-map, note tags + text, selection. Plan-level: target, device/batch quantity. All quantities derive from the dependency graph — the prototypes' math (plants = qty÷yield, domes = plants÷16, extractors sized to ~1.5h fill, power gen = units × class-multiplied output) shows the intended rollup shape, not real constants.
--- a/web/scripts/mutation-check.sh
+++ b/web/scripts/mutation-check.sh
@@ -56,7 +56,11 @@ STORED="src/state/useStoredData.ts"
 PLANNERCARD="src/card/BasePlannerCard.tsx"
 POWERBLOCK="src/card/PowerBlock.tsx"
 
+HASH="src/boundary/plan-hash.ts"
+PLANSTATE="src/boundary/plan.ts"
+
 MUTABLE="$MUTABLE $LAYOUT $TREE $METHODS $CONTROL $ASSIGN $BASES $PLANNERCARD $POWERBLOCK"
+MUTABLE="$MUTABLE $HASH $PLANSTATE"
 MUTABLE="$MUTABLE $RESOLVE $STORED"
 
 # shellcheck disable=SC2086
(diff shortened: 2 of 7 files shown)


jonstump:
## Review — approved

Conflict gate clean, CI 10/10, zero commits behind `main`. Verified against SPEC-0011 REQ "The Hash Owns the Plan, the Store Owns the Player", REQ "Signing In Is Not a Sync Trigger", and issue #148's criteria.

### The guarantee comes from the shape, not from a blocklist

The claim that *"the decoder reads five fields and has nowhere to put a sixth"* is the load-bearing one, so I read it rather than took it. `Plan` has exactly five members — target, quantity, methods, recipes, assignments — and `validatePlan` builds one by reading each by name. A hash carrying `ticks` or `notes` is not rejected; those keys are simply never read, because nothing reads them.

That is the right construction. A blocklist of forbidden keys is a list someone has to remember to extend; a type with no field for them cannot be extended by accident. And it gives the requirement its answer for free: *"no hash-derived value may be written to the store as though the player authored it"* holds because there is no path, not because a check stands in the way.

### Guarding the invisible direction hardest is the correct instinct

> A leaked note is visible the first time someone reads a link. **A decoded hash quietly authoring a place record is not visible at all**, and it is exactly the change someone makes while trying to make sharing "work better."

That asymmetry is real and the mutation follows it — *"the plan decoder starts accepting a player's ticks"* is the one that matters, and it goes red.

The encode test populating the store with a name, notes, tags, ticks **and** stocked quantities before encoding is the other half. The criterion says it "needs notes and ticks present, or it proves nothing", and a hash encoded from an empty session omits a note for the trivial reason that there was none. That is the toothless-test failure mode named and avoided rather than discovered later.

### `planToWire` omitting assignments

Verified — absent, with the reason recorded at the omission:

> Sending them anyway would be harmless today (the Go decoder ignores unknown fields on a plan) and wrong in the way that matters: it would say stage 1 takes an input it does not take.

A wire shape is a claim about what the other side accepts. Keeping it honest when the dishonest version costs nothing is the harder discipline and the right one.

### #135 and the design README

Closing #135 as resolved-by-spec is correct: it asked where assignments live, and SPEC-0011 answers it in the requirement this PR implements. Nothing left to decide.

The README correction is the right shape too — the stale line is corrected **with the reasoning recorded** rather than dropped, and the adjacent State Management paragraph is pointed at rather than left to be rediscovered. `docs/design/` is not a protected path; `docs/adrs` and `docs/openspec/specs` are, and neither is touched.

### Verification

On an integration branch carrying this PR and #161 on current `main`: **353 tests pass, 64/64 mutations red**, typecheck / lint / lint:css / format:check / check:tokens / build all clean. On this PR alone CI is 10/10.

Merging first, so #161 can rebase off the stack.

Issue #157, opened 2026-08-29
Tier 2 curated constants have no artifact and no loading path


jonstump:
ADR-0001 defines Tier 2 as "a hand-maintained YAML file of the base-economy constants that are genuinely *not* in the game files, plus the ones that are planner policy rather than game data at all." The type exists. The values do not ship.

## What is actually there

| | |
|---|---|
| `Curated` type | `web/src/boundary/requests.ts` |
| Consumers | `web/src/card/useConfiguredBase.ts`, `useLeafAssignment.ts` |
| Values anywhere | `web/tests/boundary/stages-integration.spec.ts`, `web/tests/fixtures/card-config-harness.tsx` — **fixtures only** |
| Emitted by the build | nothing in `cmd/` or `internal/domain/tier1.go` produces a Tier 2 artifact |
| Loaded by the app | nothing |

`scripts/build-wasm.sh` emits `planner.wasm`, `wasm_exec.js` and `tier1.json`. There is no `tier2` anything.

## Why it matters now

`RollupRequest` requires `Constants Curated`. Stage 2 cannot run without them, so:

- **#137 (merged)** holds leaf assignments and renders their base colours, but passes `constants: null` and therefore never dispatches a rollup. SPEC-0006 REQ "Leaf Assignment to Bases" says reassignment "MUST cause the affected figures to be recomputed through the boundary" — the code path is built and proven in a fixture, and in the application it cannot fire.
- **SPEC-0007's base planner card** has the same dependency from the other end. `useConfiguredBase` takes `constants: Curated` as a required prop, and nothing in the application can supply one.

So two surfaces are complete in code and inert in the product, for the same missing input. This is not a bug in either of them.

## What is needed

1. Where the curated set lives — ADR-0001 says a hand-maintained YAML, so: the file, its schema, and where it sits in the repo.
2. How it reaches the browser. `tier1.json` is emitted by `build-wasm.sh` and fetched at runtime; the obvious answer is the same shape, but that is a decision rather than an obvious consequence — Tier 2 carries `source` and `verified` dates per entry, which feed the `unverified` badge convention, and that has a bearing on whether it travels beside Tier 1 or inside it.
3. Who owns loading it and handing it to `rollup`. Today both consumers take it as a prop from a caller that does not exist.

ADR-0001 records the set as having shrunk to "seven scalars plus two classification maps and a verified-date map", so the content is small and known. The gap is delivery, not discovery.

## Notes

[...]

PR #161, opened 2026-08-30
feat(web): the shell opens on bases, with surfaces as view state

### LINKED ISSUE #146: The shell: bases-first entry and surface navigation

Blocked by #145 — the bases surface lists places, so it needs places to be first-class first.

Extends `web/src/shell/AppShell.tsx`, whose `<nav aria-label="Surfaces">` currently holds one button, "Plan", carrying `aria-current="page"` and no handler. This story is what finally gives `BasePlannerCard` a route.

## Requirements

- [ ] **The Shell Opens on Bases and Renders Without the Domain** (SPEC-0011) — The entry surface MUST be the bases surface. The shell's entry render MUST NOT depend on the WASM module: per SPEC-0005 REQ "Module Loading" the module loads lazily after the shell is interactive, so the entry surface MUST be complete and operable with the module unavailable, unreachable, or still loading. It MUST NOT present a loading state in place of its content, and MUST NOT present module unavailability as an error on entry.
  - WHEN the application is opened with the WASM module unavailable THEN the bases surface renders complete and interactive, with no error and no loading state standing in for its content
  - WHEN the application is opened with no places stored THEN the bases surface presents an intentional empty state inviting the player to add a place, not a screen of zeroes

- [ ] **Surfaces Are Shell View State** (SPEC-0011) — Surface selection MUST be view state the shell holds; the application MUST NOT introduce a router library. The shell MUST expose exactly one `role="navigation"` landmark, named, containing the surface controls. Cross-navigation that is not part of the switcher MUST be a content link inside `main`. A surface MUST NOT be selectable before it can render; a surface whose data is unavailable MUST present its own empty or loading state rather than being absent from the switcher.
  - WHEN each surface in turn is selected THEN exactly one `role="navigation"` landmark exists in the document in every case, and it is named
  - WHEN the module is unavailable and a surface that needs it is offered THEN the surface remains listed and presents its own loading state when selected, rather than disappearing from the switcher

## Acceptance Criteria

- The application opened with no module reaches a usable bases surface. A spinner on entry fails this
- Exactly one named `role="navigation"` landmark exists with every surface reachable — the mutation check's landmark case is the model for asserting it
- No router package appears in `web/package.json`
- The design's cross-navigation links ("view planner →", "view atlas →") are content links inside `main`, not a second nav
- The set of surfaces does not change under the player as module readiness changes

### Branch

`feat/146-shell-bases-first-entry-and-surface-navigation`

### PR Convention

```
Closes #146
Part of #143

Implements SPEC-0011 REQ "The Shell Opens on Bases and Renders Without the Domain", REQ "Surfaces Are Shell View State".
```

jonstump:
Closes #146
Part of #143

Implements SPEC-0011 REQ "The Shell Opens on Bases and Renders Without the Domain", REQ "Surfaces Are Shell View State".

> **Stacked on #160.** Base is `feat/148-state-ownership…`, not `main` — both touch `scripts/mutation-check.sh` at the same append point, and an append-vs-append conflict is worse than a deliberate rebase. **Merge #160 first.**

## Entry renders without the domain

Nothing on the bases surface crosses the boundary — places come from the durable store — so the criterion *"a spinner on entry fails this"* is met by the surface **having nothing to wait for**, not by waiting quickly. The test blocks `planner.wasm` outright and asserts the content is there, the controls are enabled, and no pending or error treatment is standing in for it.

## Surface selection is a sixth key on `ViewState`

It is the same kind of thing as `selection` — what the player is looking at, not what the domain computed — and it is a closed union of two strings. The shape test asserts the key set, so this arrived as a **visible diff**, which is what that test is for rather than something it forbids. No router: asserted against `package.json` rather than by grepping imports, because a router added and not yet imported is still a router in the tree.

## The switcher does not change under the player

Every surface is listed unconditionally, including one whose data has not arrived. A list that shrank while the binary downloaded would look correct on a fast connection and **move controls under someone's cursor mid-click** on a slow one. That is the second mutation below, and it is the one the requirement is really about.

## Two mistakes worth recording

**`.surface` was `display: contents` first.** That generates no box, so the region was invisible to the model *both* Playwright's `toBeVisible` and a screen reader's region navigation use. Every test waiting for it hung for the full thirty seconds. It is a flex column now, and the comment says why.

**Splitting the surfaces separates controls that were on screen together.** The target control is on Planner; the saved-place list and display preferences are on Bases. Ten specs needed a surface switch, and three are genuinely two-surface flows now — set a target, then go look at your places; change a display preference, then come back to the figures. That is what a player does, so the tests do it rather than being given a back door.

One consequence worth a second opinion: **"Group digits" now lives a surface away from the figures it groups.** It is player-owned state and sits with the rest of it, which is defensible — but if you would rather display preferences sat in the shell chrome and applied everywhere, that is a small change and better made now than later.

## Scope

Two surfaces, Bases and Planner, both of which exist in code. SPEC-0010's atlas is its own story; listing an empty surface for it would be inventing one. The cross-navigation links the criterion mentions ("view planner →") are **conditional** in the requirement — *if* such navigation exists it must be a content link in `main` — so the invariant is asserted (exactly one navigation landmark on every surface) rather than links invented without a design.

## Verification

**353 tests pass** (up from 343), 10 new. lint, lint:css, typecheck, format:check, check:tokens, build all clean. **64/64 mutations red**, three new:

| Mutation | Catches |
|---|---|
| the shell opens on the planner instead of bases | the entry requirement |
| **a surface leaves the switcher while the module is unavailable** | the stable-set requirement |
| switching surface leaves focus on the switcher | focus management |

No protected paths touched.

### DIFF
--- a/web/scripts/mutation-check.sh
+++ b/web/scripts/mutation-check.sh
@@ -60,7 +60,9 @@ HASH="src/boundary/plan-hash.ts"
 PLANSTATE="src/boundary/plan.ts"
 
 MUTABLE="$MUTABLE $LAYOUT $TREE $METHODS $CONTROL $ASSIGN $BASES $PLANNERCARD $POWERBLOCK"
-MUTABLE="$MUTABLE $HASH $PLANSTATE"
+SURFACES="src/shell/surfaces.ts"
+
+MUTABLE="$MUTABLE $HASH $PLANSTATE $SURFACES"
 MUTABLE="$MUTABLE $RESOLVE $STORED"
 
 # shellcheck disable=SC2086
@@ -679,6 +681,29 @@ check "the plan decoder starts accepting a player's ticks" \
   "      assignments: assignments.overrides,
       ...(raw[\"ticks\"] === undefined ? {} : { ticks: raw[\"ticks\"] }),
     },"
+
+# ----------------------------------------------------------------------
+# The shell's surfaces.
+#
+# The second mutation is the one the requirement is really about: a
+# switcher that hid a surface until its data arrived would look correct on
+# a fast connection and move controls under someone's cursor on a slow one.
+# ----------------------------------------------------------------------
+
+check "the shell opens on the planner instead of bases" \
+  tests/shell/surfaces.spec.ts "$SURFACES" \
+  'export const ENTRY_SURFACE: SurfaceId = "bases";' \
+  'export const ENTRY_SURFACE: SurfaceId = "planner";'
+
+check "a surface leaves the switcher while the module is unavailable" \
+  tests/shell/surfaces.spec.ts "$SHELL" \
+  "          {SURFACES.map((surface) => (" \
+  "          {SURFACES.filter((shown) => !shown.needsModule).map((surface) => ("
+
+check "switching surface leaves focus on the switcher" \
+  tests/shell/surfaces.spec.ts "$SHELL" \
+  "    surfaceRegion.current?.focus();" \
+  "    void surfaceRegion.current;"
 
 echo
 if [ "$failures" -gt 0 ]; then
--- a/web/src/shell/AppShell.tsx
+++ b/web/src/shell/AppShell.tsx
@@ -2,6 +2,7 @@ import {
   lazy,
   Suspense,
   useCallback,
+  useEffect,
   useId,
   useMemo,
   useRef,
@@ -49,6 +50,7 @@ const TreeCanvas = lazy(async () => ({
 }));
 import { DataCustody } from "./DataCustody";
 import { StoredPlaces } from "./StoredPlaces";
+import { SURFACES } from "./surfaces";
 import { ViewPreferences } from "./ViewPreferences";
 
 /*
@@ -253,6 +255,29 @@ function Chrome({
   readonly store: DurableStore;
 }): ReactNode {
   const state = useViewState();
+  const dispatch = useViewDispatch();
+
+  /*
+   * Focus follows the surface change.
+   *
+   * SPEC-0011 Focus Management: "Switching surface MUST move focus to the
+   * newly selected surface's region rather than leaving it on the switcher,
+   * and MUST NOT drop focus to the document body."
+   *
+   * Skipped on the first render. Moving focus into the page on load steals
+   * it from wherever the browser put it and is the behaviour screen-reader
+   * users report as a page that "jumps" — the requirement is about a
+   * change, and the first render is not one.
+   */
+  const surfaceRegion = useRef<HTMLElement | null>(null);
+  const shownSurface = useRef<string | null>(null);
+  useEffect(() => {
+    const first = shownSurface.current === null;
+    const changed = shownSurface.current !== state.surface;
+    shownSurface.current = state.surface;
+    if (first || !changed) return;
+    surfaceRegion.current?.focus();
+  }, [state.surface]);
 
   /*
    * Per-node method overrides: plan state, held here rather than in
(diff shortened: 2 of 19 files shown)


jonstump:
## Review — approved, rebased off the stack

Conflict gate clean, CI 10/10 after the rebase. Verified against SPEC-0011 REQ "The Shell Opens on Bases and Renders Without the Domain", REQ "Surfaces Are Shell View State", and issue #146's criteria.

### The rebase

Predicted in the body and it happened as predicted: GitHub auto-retargeted to `main` when #160's branch went away, and retargeting moves the base without rebasing, so this came back **DIRTY**. `git rebase --onto origin/main 18c1c82` drops the duplicated #160 commit and leaves **+581/-69 across 19 files**, which is the real change. Verified the rebased tree rather than trusting a clean rebase: 353 tests, all gates green.

### "Renders without the domain" is literally true, not merely fast

The criterion says *"a spinner on entry fails this"*, and the body's answer — the surface has nothing to wait for rather than waiting quickly — is the right distinction. Confirmed in the running application on a cold load:

```
opens on:                Bases   (aria-current="page")
navigation landmarks:    1, named "Surfaces"
regions in main:         Bases · Saved places · Your data · Preferences
domain assets fetched:   []      ← planner.wasm, wasm_exec.js, tier1.json: none requested
pending / error badges:  []
```

Zero domain requests on entry. Not "the spinner is brief" — there is no spinner because there is nothing to wait for.

### The stable switcher, and focus

The requirement this is really about — *"the set of surfaces does not change under the player"* — checked live across the transition:

```
before:            [Bases, Planner]
immediately after: [Bases, Planner]
after settling:    [Bases, Planner]
focus after switch: SECTION "Planner"   ← the surface region, not the switcher
```

The reasoning for it is right and worth keeping: a list that shrank while the binary downloaded *"would look correct on a fast connection and move controls under someone's cursor mid-click on a slow one."* Mutation 2 keeps it, mutation 3 keeps the focus move.

### Two mistakes recorded, both worth having recorded

`display: contents` generating no box — so the region was invisible to the same model `toBeVisible` and screen-reader region navigation both use — is a genuinely non-obvious failure and the comment now carries it.

And the honest note that splitting surfaces separated controls that used to be on screen together, with ten specs needing a switch and three becoming real two-surface flows. Making the tests do what a player does rather than giving them a back door is the right call.

### The second opinion you asked for — and I think you should move it

> "Group digits" now lives a surface away from the figures it groups.

I checked rather than reasoned, and the case is stronger than "a bit awkward". **Every consumer of `preferences.groupSeparator` is on Planner:**

```
AppShell.tsx:109   the live region describing the resolution
AppShell.tsx:226   the figure rows
TreeCanvas.tsx     184, 196, 209, 311
```

`StoredPlaces` never calls `formatQuantity` — stocked quantities go through `storedQuantity`/`ABSENT_DISPLAY` unformatted. So the Preferences panel now sits on a surface where **it has no observable effect at all**. A player toggles "Group digits", nothing on screen changes, and the consequence is one surface away. That is a control operated blind.

There is a second reason from this repo's own history. In #124 the canvas ignored this preference while the flat list honoured it — the same figure grouped in one place and not the other. What made that *catchable in review* was that both were on screen at once. After this PR they cannot be.

My recommendation is shell chrome. It is player-owned state, which is the argument for Bases, but it is player-owned state **about how figures are rendered**, and the figures are all on Planner. Not blocking — the requirement does not specify placement, and you flagged it honestly rather than letting it pass.

**A related note for whenever the base planner card mounts:** `BasePlannerCard`, `BuildFooter`, `Duration`, `PowerBlock` and `CardControls` all call `formatQuantity(value)` with no options, so they group with the default comma and ignore the preference entirely. Pre-existing and not this PR's doing — the card is not mounted (#157) — but it is a third surface that will need wiring.

### Scope

Listing only Bases and Planner is right: SPEC-0010's atlas is its own story and an empty surface for it would be inventing one. Asserting the invariant (exactly one navigation landmark per surface) rather than inventing "view planner →" links the design has not drawn is the same judgement, and the requirement's own conditional wording supports it.

### Verification

**353 tests pass**, 10 new. **64/64 mutations red**, three new — the entry requirement, the stable-set requirement, and focus management. All gates clean, on the rebased branch.

Merging.

PR #162, opened 2026-08-30
feat(web): the catalogue crosses the boundary, and the target is a search

### LINKED ISSUE #147: Target search and the catalogue boundary call

Extends `internal/bridge/` (which today exposes `resolve`, `rollup` and `power` and no catalogue), `web/src/boundary/client.ts`, and the target control in `web/src/shell/AppShell.tsx` — currently a bare `<input>` whose value goes to the domain as an item id, so the only way to load anything is to already know a string like `ULTRAPROD2`.

**This story changes the module's contract version.** It is the one place in this epic where the boundary surface grows.

## Requirements

- [ ] **The Catalogue Crosses the Boundary** (SPEC-0011) — The module MUST expose an entry point returning the searchable item catalogue: for each selectable item, its id and its display name. The call MUST follow SPEC-0002's envelope, error-code and contract-version rules exactly as `resolve`, `rollup` and `power` do, and MUST be subject to the same not-ready handling. Adding this entry point is a change to the module's contract version.
  - WHEN the catalogue is requested before the module is ready THEN the view presents a loading state and retries once readiness resolves, rather than reporting an error

- [ ] **Target Selection Is a Search Over Known Items** (SPEC-0011) — The target control MUST be a search over known items, not a field accepting only an internal item id. It MUST match against both display name and id, MUST tolerate partial and inexact input, and MUST present the display name as primary with the id as secondary. The searchable list MUST arrive through the module boundary: per SPEC-0005 REQ "Boundary Client" the view MUST NOT read the Tier 1 artifact directly, and MUST NOT ship a compiled-in copy of the item list.
  - WHEN a player types a partial display name THEN the matching item is offered, and selecting it sets the target without the id being typed
  - WHEN a player types an exact item id THEN that item is offered
  - WHEN the view's sources are scanned THEN no item list literal is present, and the catalogue is reached only through the boundary client

## Acceptance Criteria

- A player who has never seen an item id can reach any selectable item. This is the criterion the current control fails
- No item list literal exists in the view sources — checkable mechanically, the way `tests/card/discipline.spec.ts` checks for arithmetic
- The contract-version bump is handled by SPEC-0002's existing mismatch rule: a view built against the older contract reports the mismatch and refuses the payload
- Per SPEC-0011 § Rate Limiting, the search MUST NOT issue a catalogue call per keystroke
- The control is operable as a combobox by keyboard alone, and its result count is announced as it changes

### Branch

`feat/147-target-search-and-the-catalogue-boundary-call`

### PR Convention

```
Closes #147
Part of #143

Implements SPEC-0011 REQ "The Catalogue Crosses the Boundary", REQ "Target Selection Is a Search Over Known Items".
```

jonstump:
Closes #147. Part of #143.

Implements SPEC-0011 REQ "The Catalogue Crosses the Boundary", REQ "Target Selection Is a Search Over Known Items".

> Stacked on #161 → #160. Merge order: **#160 → #161 → this**.

The target control was a bare input whose value went to the domain as an item id, so the only way to load anything was to already know a string like `ULTRAPROD2`. It is a search now, over a catalogue that arrives through the module boundary, matching display names and ids, with the name leading.

## The Go half

`catalogue` is an entry point returning each selectable item's id and display name, listed in `EntryPoints`, reachable through `Call` by name, and subject to the same readiness sentinel as every other stage. Three Go tests cover the payload shape, the not-ready path, and dispatch by name.

The payload carries **only** id and name — asserted by marshalling the envelope and checking `recipes`, `default_method` and `verified` are absent, because anything more would be domain data the view is not allowed to reason about.

**The contract moves to 1.4.0.** Additive, but still a contract change: a view built against 1.4.0 calls an entry point a 1.3.0 module does not carry. SPEC-0002's mismatch rule handles the skew, and `useCatalogue` never retries a mismatch — asking a module that lacks the entry point a second time only loops.

## The view half

A decoder, a client method, a hook that fetches once (not per keystroke, per SPEC-0011 § Rate Limiting), and a combobox that matches names *and* ids, tolerates partial input, shows the name first and the id second, filters locally, moves through results with the arrow keys, dismisses with Escape, and announces its result count as it changes.

`tests/shell/target-search.spec.ts` is new — nine tests covering reach-by-name, the name leading, one crossing regardless of keystrokes, the count announcement, and Escape dismissing without clearing. One of them is mechanical rather than behavioural: it scans `src/{shell,canvas,card,state}` for real artifact ids, because "the view carries no item list of its own" is an absence, and an absence has to be checked at the source or not at all.

## Five defects found on the way, all mine

**A StrictMode deadlock.** A boolean "have I asked" guard is incompatible with the double-mount: the first effect starts the fetch, its cleanup marks the result unwanted, and the second sees the flag set and never fetches — so nothing settles and the control sits on its loading state forever. The in-flight promise is cached instead, so one crossing happens and both mounts subscribe to it.

**Two ARIA faults in my own markup.** Options were nested `<button>`s, which puts a tab stop per result into the shell's tab order. And `aria-controls` pointed at a listbox that only existed while open — an idref to nothing. The listbox is always present and `hidden`; options are list items and focus stays on the input via `aria-activedescendant`.

**A third tab stop I did not predict.** The listbox scrolls, and Chrome makes scrollable containers focusable whether or not they have a role that wants focus — so it took a tab stop anyway, below both the controls it sits between. The shell's reading-order assertion caught it. `tabIndex={-1}`.

**Two racing live regions.** The result count was a second `aria-live` region, so every test locating the shell's announcer matched two elements. The count keeps its own polite region and is deliberately not `role="status"`; tests locate the announcer by role.

**Choosing a target by text is ambiguous against the real artifact.** `"ANTIMATTER"` matches five of the 2,237 items, four of them by name, and the first in artifact order is `AM_HOUSING` — so tests silently resolved the wrong tree and every assertion after it looked for nodes that were never going to be there. Options carry `data-item-id` and the helper selects on it.

## Two existing tests changed rather than worked around

The composite-widget guard asked in its own failure message to be replaced by a test that drives the arrow keys, once a composite widget existed. This is that widget, so it is replaced by a real arrow-key test and the guard is narrowed to the roles still uncovered (`tablist`, `menu`, `menubar`, `tree`, `grid`, `radiogroup`, `toolbar`).

The figures-pending test observed a window that no longer exists: it selected a target before the module was ready, and target selection now needs the catalogue, which needs the module. It asserts the requirement's actual content instead — that nothing is fabricated as a zero while the module loads — with a comment recording why the old observation point is unreachable.

## Mutations

Four added, all verified to turn the suite red:

| Mutation | Watched by |
|---|---|
| the search stops matching display names | `target-search.spec.ts` |
| a compiled-in item list appears in the view | `target-search.spec.ts` |
| the results list goes back into the tab order | `a11y-baseline.spec.ts` |
| the result count becomes a second status region | `target-search.spec.ts` |

The third initially named the wrong spec and did not go red — the tab order is watched by the shell's reading-order assertion, not by the search's own file. Retargeted, and the whole 68-mutation suite passes.

**Green:** 363 web tests, all Go tests, and the six gates (`lint`, `lint:css`, `typecheck`, `format:check`, `check:tokens`, `build`).

### DIFF
--- a/internal/bridge/bridge.go
+++ b/internal/bridge/bridge.go
@@ -49,7 +49,7 @@ import (
 // that does not know about `configured` reads a base with a zero site as
 // one configured at class "" for zero seconds, which is exactly the zero
 // SPEC-0011 REQ "A Place Is Creatable by Hand" forbids presenting.
-const ContractVersion = "1.3.0"
+const ContractVersion = "1.4.0"
 
 // Quantity is an exact quantity on the wire.
 //
@@ -113,9 +113,30 @@ type Envelope struct {
 // absent rather than present-and-empty. Exactly one is set by any entry
 // point this package exposes.
 type ResultPayload struct {
-	Graph *Graph `json:"graph,omitempty"`
-	Build *Build `json:"build,omitempty"`
-	Power *Power `json:"power,omitempty"`
+	Graph     *Graph     `json:"graph,omitempty"`
+	Build     *Build     `json:"build,omitempty"`
+	Power     *Power     `json:"power,omitempty"`
+	Catalogue *Catalogue `json:"catalogue,omitempty"`
+}
+
+// Catalogue is the searchable item list on the wire.
+//
+// Governing: SPEC-0011 REQ "The Catalogue Crosses the Boundary" — "for each
+// selectable item, its id and its display name".
+//
+// Nothing else. The view searches over this and sends back an id, so a
+// recipe, a method or a yield would be data the search does not use and the
+// view is not allowed to reason about anyway. Every item in the artifact is
+// listed: the domain decides what resolves, and a view that pre-filtered
+// would be deciding which items exist.
+type Catalogue struct {
+	Items []CatalogueItem `json:"items"`
+}
+
+// CatalogueItem is one selectable item: what to send, and what to show.
+type CatalogueItem struct {
+	ID   string `json:"id"`
+	Name string `json:"name"`
 }
 
 // ErrorPayload is the failure half.
--- a/internal/bridge/module.go
+++ b/internal/bridge/module.go
@@ -220,6 +220,32 @@ func (m *Module) loaded() (*domain.Tier1, bool) {
 	return m.artifact, m.artifact != nil
 }
 
+// Catalogue returns the selectable items, each as an id and a display name.
+//
+// Governing: SPEC-0011 REQ "The Catalogue Crosses the Boundary", SPEC-0002
+// REQ "Boundary Surface", REQ "Module Lifecycle and Readiness"
+//
+// Subject to the same not-ready handling as every other entry point: the
+// artifact is what makes the list, so before it loads there is no list, and
+// saying so with the readiness sentinel is what lets the view show a
+// loading state and retry rather than report an error.
+//
+// The argument is ignored. The catalogue takes no input — filtering is the
+// view's job, over a list it already holds, and a per-keystroke crossing is
+// what SPEC-0011 § Rate Limiting forbids.
+func (m *Module) Catalogue(_ string) Envelope {
+	artifact, ok := m.loaded()
+	if !ok {
+		return FailureFrom(fmt.Errorf("catalogue: %w", ErrNotReady))
+	}
+
+	items := make([]CatalogueItem, 0, len(artifact.Items))
+	for _, item := range artifact.Items {
+		items = append(items, CatalogueItem{ID: item.ID, Name: item.Name})
+	}
+	return Success(ResultPayload{Catalogue: &Catalogue{Items: items}})
+}
+
 // EntryPoints are the names the namespace object carries.
 //
 // Governing: SPEC-0002 REQ "Boundary Surface" — one named entry point per
@@ -229,7 +255,7 @@ func (m *Module) loaded() (*domain.Tier1, bool) {
 // Enumerated here rather than inline in the js shim so the surface is
 // listable from a plain test, and so adding one is a change to this list
 // rather than a change scattered through registration code.
-var EntryPoints = []string{"load", "ready", "resolve", "rollup", "power"}
+var EntryPoints = []string{"load", "ready", "resolve", "rollup", "power", "catalogue"}
 
 // Namespace is the single global name the module registers under.
 const Namespace = "nmsPlanner"
@@ -253,6 +279,8 @@ func (m *Module) Call(name, arg string) Envelope {
 		return m.Rollup(arg)
 	case "power":
 		return m.Power(arg)
+	case "catalogue":
+		return m.Catalogue(arg)
 	default:
 		return Failure(CodeMalformedInput,
 			fmt.Sprintf("%q is not an entry point on %s; expected one of %v", name, Namespace, EntryPoints))
--- a/internal/bridge/module_test.go
+++ b/internal/bridge/module_test.go
@@ -241,3 +241,84 @@ func TestMalformedPlanAtTheEntryPoint(t *testing.T) {
 		t.Errorf("the module broke after a malformed call: %+v", env.Error)
 	}
 }
+
+// SPEC-0011 REQ "The Catalogue Crosses the Boundary":
+// WHEN the catalogue is requested THEN the module returns each selectable
+// item's id and display name, through the same envelope every other entry
+// point uses.
+func TestCatalogueListsSelectableItems(t *testing.T) {
+	m := loadedModule(t)
+
+	env := m.Catalogue("")
+	if !env.OK {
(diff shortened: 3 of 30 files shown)


jonstump:
## Review — approved

Conflict gate clean, CI 10/10, zero commits behind `main` (the stack ahead of it merged, and this arrived rebased). Verified against SPEC-0011 REQ "The Catalogue Crosses the Boundary", REQ "Target Selection Is a Search Over Known Items", and issue #147's criteria.

### This is the complaint that started the shell work, and it is fixed

The original report was that the only way to load anything was to already know a string like `ULTRAPROD2`, and that nothing visible in the chart could be found by typing. Driven in the running application, typing `chrom`:

```
Chromatic Starship Trail   SHIP_RAINBOW
Dual Chrome Firework       SPEC_FIREWORK05
Chromatic Metal            STELLAR2
Polychromatic Zirconium    TRA_MINERALS4
```

Name first, id second, and it matches **mid-word** — `chrom` reaching `Polychromatic` is a contains-match rather than a prefix one, which is what makes it usable by someone who half-remembers a name. Arrow keys moved `aria-activedescendant` to `Chromatic Metal STELLAR2`, Enter put **"Chromatic Metal"** in the field rather than the id, and Recompute resolved it: heading "Chromatic Metal", four node cards.

### The rate-limit claim, measured rather than asserted

SPEC-0011 § Rate Limiting requires the catalogue be fetched once, not per keystroke. Wrapped the module's `catalogue` entry point and typed five characters:

```
catalogue crossings during "c" → "ch" → "chr" → "chro" → "chrom":  0
```

Zero. Fetched once on mount, filtered locally, and the in-flight-promise cache means the StrictMode double-mount subscribes to one crossing rather than racing two.

That StrictMode finding is worth its paragraph in the body. A boolean "have I asked" guard is the obvious implementation and it **deadlocks**: first effect starts the fetch, its cleanup marks the result unwanted, the second sees the flag set and never fetches, and the control sits on its loading state forever. This is the third story this epic where an effect-plus-flag lost to StrictMode — #137 and #134 both landed on the handler for the same reason. The pattern is worth remembering.

### The three ARIA faults, verified fixed

Each one checked on the live control rather than in the diff:

```
role="combobox"  label="Target"  aria-expanded toggles false → true
aria-controls  → "_r_6_"   element exists: true      ← was an idref to nothing
buttons nested inside the listbox: 0                 ← was a tab stop per result
listbox tabindex: "-1"                               ← Chrome focuses scrollable containers regardless of role
```

The third is the one worth flagging to whoever writes the next scrolling widget: a container becomes focusable because it scrolls, not because it asked to be, and it landed *below* both controls it sits between. The shell's reading-order assertion caught it, which is exactly what that assertion is for.

And the live regions:

```
"4 items match"  → aria-live, role=null     (the count, its own region)
""               → role="status"            (the shell's announcer, separate)
```

One announcer locatable by role, the count beside it. That is the fix, and it is the right shape — two `role="status"` regions is a locator ambiguity in tests and a double announcement for a screen reader.

### The payload carries only what the view may reason about

`Catalogue` on the wire is `{items: [{id, name}]}`, and the Go test asserts `recipes`, `default_method` and `verified` are **absent** by marshalling the envelope and checking for their keys. Asserting an absence at the serialised form rather than at the struct is the right level: a field added later with a JSON tag would fail there and nowhere else.

The contract bump to 1.4.0 is consistent on both sides — `EXPECTED_CONTRACT_VERSION` in `contract.ts` and `ContractVersion` in `bridge.go`, with the version history comment extended. And not retrying on mismatch is right: asking a module that lacks the entry point a second time only loops.

### The two existing tests changed rather than worked around

The composite-widget guard **asked in its own failure message** to be replaced once a real composite widget existed. This is that widget, so it is replaced by a real arrow-key test and the guard narrowed to the roles still uncovered. That is the intended lifecycle of a placeholder assertion, not an erosion of it.

The figures-pending test observed a window that no longer exists — selecting a target now needs the catalogue, which needs the module — so it asserts the requirement's actual content instead, with a comment recording why the old observation point is unreachable. Right call, and recorded rather than quietly rewritten.

### The defect I would have missed

> `"ANTIMATTER"` matches five of the 2,237 items, four of them by name, and the first in artifact order is `AM_HOUSING` — so tests silently resolved the wrong tree and every assertion after it looked for nodes that were never going to be there.

Selecting by visible text against a 2,237-item catalogue is ambiguous in a way that produces a *passing-looking* wrong answer, and `data-item-id` is the fix. Worth recording because every future test that picks a target has the same trap.

### Verification

**363 web tests pass**, all Go tests pass, **68/68 mutations red**, and all six gates clean — run locally on this branch, not just taken from CI. The retargeted mutation is noted honestly in the body: it named the wrong spec, did not go red, and was pointed at the file that actually watches the tab order.

Merging.

Issue #144, opened 2026-08-29
Implement Base Atlas


jonstump:
Implements [SPEC-0010: Base Atlas](../blob/main/docs/openspec/specs/base-atlas/spec.md) ([design.md](../blob/main/docs/openspec/specs/base-atlas/design.md)).

Governed by [ADR-0015](../blob/main/docs/adrs/ADR-0015-base-atlas-coordinate-space-and-route-graph.md) — Atlas positions and route graphs are authored data rendered by the view, with no new boundary crossing.

## Why this epic exists

`docs/design/bases-map/` has been a complete, high-fidelity design for a long time — a pixel map of every base, dashed district territories, harvest runs with numbered waypoints and per-leg travel methods, a handoff document and an interactive prototype — with no ADR, no spec, no issues and no implementation, while two accepted decisions already reasoned about it as though it existed.

Two things arrive with this surface that the project does not otherwise have: a **coordinate space** (places acquire positions the player authors and the game never supplies) and a **route graph** (a harvest run is an ordered sequence of stops with a travel method per leg). ADR-0015 decided both are authored data rendered by the view, because there is no distance to minimise — Atlas positions are arrangement, not geography.

## Stories

| # | Story | Label |
|---|---|---|
| #151 | Positions, districts, and the schema change | `foundation` |
| #152 | Harvest runs: authoring, seeding, activation, and unresolved stops | |
| #153 | The Atlas surface: canonical list, run identity, and shell integration | |
| #154 | Atlas boundaries: no boundary call, no hash, and share disclosure | |
| #155 | Tests: positions, districts, the freighter, and the schema bump | `test` |
| #156 | Tests: runs, the canonical list, colour independence, and boundaries | `test` |

**Order:** #151 (foundation) → #152, #154 in parallel → #153 → #155, #156.

## Cross-epic dependency

#153 is blocked by #146 in the Shell Surface epic (#143). SPEC-0010 REQ "The Atlas Is a Surface in the Shell" depends on the shell's surface registry and its single navigation landmark, which #146 builds.

### Branch

`epic/144-implement-base-atlas`

PR #163, opened 2026-09-14
Positions, districts, and the schema change

### LINKED ISSUE #151: Positions, districts, and the schema change
Part of #144. Governed by [SPEC-0010](../blob/main/docs/openspec/specs/base-atlas/spec.md) and [ADR-0015](../blob/main/docs/adrs/ADR-0015-base-atlas-coordinate-space-and-route-graph.md).

**Foundation story.** It changes the stored schema, and every other story in this epic writes into it. Merge this before the others begin.

Extends `web/src/store/durable-store.ts` and the SPEC-0009 `PlaceRecord`.

## Requirements

- [ ] **Position Is Optional and Authored** (SPEC-0010) — A place's Atlas position MUST be a nullable field on the SPEC-0009 `PlaceRecord`, not a separate record type and not a required property. A position MUST be two integers in the Atlas's own grid space, meaningful only relative to other positions in the same workspace. It MUST NOT be derived from, seeded from, or reconciled against `GalacticAddress` or the in-game `Position`. A place with no position is a first-class state: it MUST appear in the place list, MUST be available as a run stop, MUST NOT appear on the map, and no placeholder coordinate may be substituted.
  - WHEN a workspace holds a place that has never been positioned THEN the place appears in the place list, is selectable as a run stop, is absent from the map, and no error or empty-state warning is raised about it
  - WHEN a place is created by save import from a save carrying `GalacticAddress` and `Position` THEN its Atlas position is null

- [ ] **A Freighter Is a Route Node Without a Position** (SPEC-0010) — A freighter MUST be representable as a run stop while carrying no position, per ADR-0006. The surface MUST NOT special-case the freighter: it follows from position being optional, and any conditional on place kind in the positioning or map-rendering path is a defect. The freighter MUST NOT be drawn on the map, enclosed by a district rectangle, or contribute a point to any bounding box.
  - WHEN an active run stops at a freighter between two positioned bases THEN the freighter appears in the run's ordered legs at its authored position in the sequence, and does not appear on the map

- [ ] **A District Is a Tag and Its Rectangle Is Derived** (SPEC-0010) — A district MUST be a name carried on a place. There MUST NOT be a district record, and a district's geometry MUST NOT be stored. The dashed territory MUST be the bounding box of its positioned members, computed in the view at render time. A place with no district is a first-class state, drawn outside any territory, with no default or "ungrouped" pseudo-district record.
  - WHEN a member of a district is repositioned outside the current territory THEN the rendered rectangle encloses the new arrangement, with nothing written to the store beyond the moved place's position
  - WHEN a positioned place carries no district THEN it appears on the map outside every territory, and no district record was created for it

- [ ] **The Schema Change Fails Legibly** (SPEC-0010) — Adding position and district to `PlaceRecord`, and adding the run record, is a schema change under SPEC-0009. It MUST take a `schemaVersion` increment and MUST inherit REQ "Versioned, and Fails Legibly": a workspace at the previous version loads nothing and reports both versions, rather than loading places with positions omitted. Positions and runs MUST inherit REQ "Storage Is Evictable and the Application Must Not Imply Otherwise".
  - WHEN a workspace written at the previous `schemaVersion` is opened THEN nothing loads, and both the stored and the expected version are reported

## Acceptance Criteria

- Position is a nullable field on `PlaceRecord`. A separate positions table fails this, however convenient
- No code path assigns an Atlas position from save-file contents — checkable by searching for `GalacticAddress` and `Position` in any positioning path
- No conditional on place kind exists in the positioning or map-rendering path. The freighter working is a consequence, not a special case
- No district record type and no stored rectangle exist. Moving a member changes the territory with one store write, not two
- The version bump is exercised by a test that opens an old workspace, not assumed
- The surface does not describe an arrangement as saved in terms stronger than an evictable store can honour

### Branch

`feat/151-positions-districts-and-the-schema-change`

### PR Convention

```
Closes #151
Part of #144

Implements SPEC-0010 REQ "Position Is Optional and Authored", REQ "A Freighter Is a Route Node Without a Position", REQ "A District Is a Tag and Its Rectangle Is Derived", REQ "The Schema Change Fails Legibly".
```

jonstump:
Closes #151
Part of #144

Implements SPEC-0010 REQ "Position Is Optional and Authored", REQ "A Freighter Is a Route Node Without a Position", REQ "A District Is a Tag and Its Rectangle Is Derived", REQ "The Schema Change Fails Legibly".

Foundation story for the Atlas epic. Adds the durable shape the other stories write into, and nothing else — no surface, no rendering.

## What changed

| File | |
|---|---|
| `store/schema.ts` | `SCHEMA_VERSION` 1 → 2; `AtlasPosition`, `position`, `district` on `PlaceRecord`; `RunRecord` / `RunStop` / `TravelMethod` |
| `store/absence.ts` | `positionOf`, `districtOf` — the single readers |
| `store/durable-store.ts` | Runs object store, strict version checks, `readRun`, `putRun`, `deleteRun` |
| `tests/store/durable-store.spec.ts` | Six tests, including the version bump the criteria name |

## The load-bearing fix

The version check read `> SCHEMA_VERSION`. While version 1 was the only version there had ever been, nothing older could exist and the asymmetry was invisible.

At version 2 it stops being invisible: an accepted version-1 place loads as a place with **no position** — and that is a real state a player can author. The guess would be indistinguishable from the truth. It is now strict in both directions, for the workspace, places, and runs, and there is a test that opens a version-1 workspace rather than assuming the bump works.

## Decisions worth a second look

**Position is a field on the place, not a positions table.** Unpositioned has two spellings on the wire — absent, and present-but-null — and `positionOf()` is the single reader that turns both into `ABSENT`, so the two never reach logic. This follows the `absence.ts` idiom already established for stocked quantities and ticks, where the whole point is that `?? 0` does not typecheck.

**The freighter needs no branch.** ADR-0006 gives it no fixed location, and it takes the same field through the same path. `grep 'kind === "freighter"'` over `web/src` returns nothing. The test includes a freighter specifically to show it is not special-cased.

**No position comes from save contents.** `grep GalacticAddress web/src` returns exactly one hit: the comment in `schema.ts` explaining why it is never used. The Atlas is an arrangement the player authors; seeding it would make the map a lossy render of coordinates it cannot project while overwriting what the player arranged.

**Runs take stops as given** — not sorted, not deduplicated, not filtered for stops whose place is gone. `deletePlace` deliberately does not cascade to runs, which is what lets a deleted place's stop stay in sequence and be rendered unresolved rather than renumbering the route around the gap.

## One thing I changed outside the obvious diff

The "places without a workspace record" test planted `schemaVersion: 1` as a literal. With the bump that reads as a stale assumption, so I passed `SCHEMA_VERSION` in — and it failed, because `page.evaluate` runs its callback in the browser where a module-scope import does not exist. It typechecks and throws at runtime. Now passed as an argument, with a comment, since the next person to touch a store test will hit the same edge.

## Verification

355 Playwright tests pass. Typecheck, lint, Prettier, and the token check are clean. No design documents were modified.


jonstump (commented):
## Review: #151 against SPEC-0010

Conflict-marker gate: clean. CI: all ten checks green. Mergeable: clean.

> Posted as a **comment**, not an approval. I authored two of these three commits, and GitHub refuses self-approval. The findings below are not blocking — on a PR I hadn't written I would approve with them.

### Acceptance criteria, checked against the head (`ec89919`)

| Criterion | Evidence |
|---|---|
| Position is a nullable field, not a positions table | `position?: AtlasPosition \| null` on `PlaceRecord`; no second store |
| No position assigned from save contents | `git grep GalacticAddress web/src` → one hit, the comment saying why it is never used |
| No conditional on place kind in the positioning path | `git grep 'kind === "freighter"' web/src` → none |
| No district record, no stored rectangle | The only `rectangle` hit is the comment saying it is derived at render time |
| The version bump is exercised, not assumed | Two tests: a previous-version workspace, and a previous-version place |

### `ec89919` fixed a real bug in the original commit

The first commit validated a position on **read** but not on **write**. `AtlasPosition` types its axes as `number` while the read check requires integers, so a fractional position — the natural output of a drag — typechecked, wrote `ok`, and then made `load()` fail. Since a load is all-or-nothing by design, that is not one lost marker; it is every place and run in the workspace, with no server copy under ADR-0008.

Sharing `checkPlacement` between both directions is the right fix, and the two added tests assert the part that matters: the neighbours survive and the workspace still loads. I missed this, and it is exactly the shape of bug the foundation story should catch before a drag interaction trips it.

### Finding 1 — the governing ADR is not cited

Issue #151 is governed by **SPEC-0010 and ADR-0015** (*"The Base Atlas — an authored coordinate space, and a route graph that is presentation"*). The governing comments cite SPEC-0010 throughout and ADR-0015 nowhere.

The code already honours it:

- ADR-0015: *"Whatever shape positions take, 'has a position' cannot be a property every place carries."* → position is optional.
- ADR-0015 extends SPEC-0006's line that *"positions are the view's business, quantities are the domain's."* → positions live in `web/src/store`, not the Go domain.

So this is a traceability gap, not a design one. Adding `ADR-0015` to the `Governing:` lines on `AtlasPosition`, the `position` field, and `RunRecord` lets `/sdd:check` trace the store back to the decision that shaped it.

### Finding 2 — `readPlace` lost its doc comment

`ec89919` inserted `checkPlacement` between `readPlace`'s JSDoc block and `readPlace` itself (`durable-store.ts`, lines ~81–124 at the head). The result is two stacked comment blocks above `checkPlacement` — tooling attaches only the second — and no comment on `readPlace` at all.

The orphaned block is the one worth keeping: it explains why `readPlace` returns `StoreResult` rather than `PlaceRecord | StoreFailure`, because both carry a `kind` field and narrowing on it "happened to work and read as though it were designed." Moving that block back above `readPlace` fixes it.

### Process note — the qmd ADR index is stale

The `nms-base-planner-adrs` collection does not contain ADR-0015; a query for *"base atlas coordinate space route graph"* returns ADR-0001 through ADR-0007 and nothing later. That is why the review skill's automated missing-reference retrieval could not surface Finding 1 — it is blind to every ADR written since the last re-index. `/sdd:index update` would close it.

### Standing of this review

The first two commits are mine, so this is not an independent check of them. `ec89919` was independent, and it caught a defect in my work — which is the better evidence of what an outside read contributes.

